PR #155, opened 2014-03-24
Add a type for purging files from Fastly

sihil:
Replacement pull request for reverted #147 

Whilst it doesn't really feel like a top-level deployment type, it's something we're currently doing manually as part of a deploy and I didn't want to bloat the S3 deployment type with it. I don't know if there's some other mechanism that would make more sense, but it doesn't feel like a case for a hook.


sihil:
After almost two and a half years I don't feel like this is relevant anymore. Shout if you think this should be revived @TBonnin @philwills

TBonnin:
I am fine with that, especially since it is a >2y PR 
👌

PR #208, opened 2014-12-05
NO MORE REGULAR EXPRESSIONS.

philwills:
I've gone for readability over brevity and even with the imports it's barely bigger than the regex.


sihil:
:+1:

And another :+1: for the CAPITAL LETTERS.

adamnfish:
If it's only the filename / extension you're after why not just use a library designed to work that out (like `FilenameUtils`)?

PR #220, opened 2014-12-24
Catch the exception that we get when there are no updates to apply

sihil:
This changes the error you get when there are no changes:
screen shot 2014-12-24 at 21 01 06

Into a successful deploy:
screen shot 2014-12-24 at 21 01 25

There is currently a bug in which the `CheckUpdateEventsTask` task doesn't really work and if fixed might break this again - but this will work for now.

### DIFF
--- a/magenta-lib/src/main/scala/magenta/tasks/UpdateCloudFormationTask.scala
+++ b/magenta-lib/src/main/scala/magenta/tasks/UpdateCloudFormationTask.scala
@@ -1,5 +1,6 @@
 package magenta.tasks
 
+import com.amazonaws.AmazonServiceException
 import magenta.{MessageBroker, Stage, Stack, KeyRing}
 import com.amazonaws.regions.Regions
 import com.amazonaws.services.cloudformation.AmazonCloudFormationAsyncClient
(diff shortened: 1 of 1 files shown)


magenta-lib/src/main/scala/magenta/tasks/UpdateCloudFormationTask.scala
```diff
@@ -27,7 +28,12 @@ case class UpdateCloudFormationTask(cloudFormationStackName: String, template: P
     MessageBroker.info(s"Parameters: $actualParameters")
 
     if (CloudFormation.describeStack(cloudFormationStackName).isDefined)
-      CloudFormation.updateStack(cloudFormationStackName, template.string, actualParameters)
+      try {
+        CloudFormation.updateStack(cloudFormationStackName, template.string, actualParameters)
+      } catch {
+        case ase:AmazonServiceException if ase.getMessage contains "No updates are to be performed." =>
```
philwills: I'm guessing you looked at the errorCode on this and it wasn't helpful.

sihil: The HTTP status code or an amazon error code? If the latter, I didn't notice it existed.

philwills: I was thinking Amazon error code, which there seems to be a method for on `AmazonServiceException`

sihil: Ah - good spot. I'll try and revisit that - that'd certainly be a more robust approach.

sihil: I think that the error code returns a pretty generic value from the failure examples in Riff-Raff in CODE. e.g. http://riffraff.code.dev-gutools.co.uk/deployment/view/edf42137-200b-4021-8925-650881a79255
Doesn't seem to be a specific identifiable code that is likely to be unique.

philwills: :disappointed:

PR #221, opened 2015-01-07
Added a "warmup" grace period.

gklopper:
This leaves the loadbalancer at "double" size for a configurable number of seconds to give the instances time to warm up.

We never realised it, but the 2nd `HealthcheckGrace` (recently removed) was fulfilling this function for us, hence crash and burn when deploying under load today.

@sihil @philwills (please)

### DIFF
--- a/magenta-lib/src/main/scala/magenta/deployment_type/AutoScaling.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/AutoScaling.scala
@@ -72,6 +72,7 @@ object AutoScaling  extends DeploymentType with S3AclParams {
   )
   val secondsToWait = Param("secondsToWait", "Number of seconds to wait for instances to enter service").default(15 * 60)
   val healthcheckGrace = Param("healthcheckGrace", "Number of seconds to wait for the AWS api to stabilise").default(20)
+  val warmupGrace = Param("warmupGrace", "Number of seconds to wait for the instances in the load balancer to warm up").default(1)
 
   def perAppActions = {
     case "deploy" => (pkg) => (lookup, parameters, stack) => {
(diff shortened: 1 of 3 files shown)


philwills:
I'm going to merge this, but I can't see how it can make a difference compared to just having a slightly larger healthcheckGrace, which concerns me.

I'd suggest we're still careful.

philwills:
I've now deployed this, but can you point me to the deploy record for the bad deploy.  I've had a quick look through the history and can only find Nathaniel's deploy to PROD today and that looks OK, at least for facia and article.

gklopper:
We need a pause at the start of the deploy and at the end. The one at the start gives the Amazon API a chance to catch up (we have observed it initially report incorrectly in the past) and does not materially affect the length of the deploy. 

The second one just gives the newly started up instances a chance to warm up under lesser load before culling off the old instances.

You could achieve the same effect by lengthening the initial healthcheckGrace, but you would have to make it long enough to guarantee all the instances are up before moving on (I'm not sure that is feasible).

If you mean the healthcheck grace built into the autoscaling group then I have considered that, but I don't think it will help as this problem occurs after the healthcheck passes.

I am doing some thinking about whether this is an appropriate way of warming up boxes, but this has just come to my attention and will take a bit of time for me to figure that out.

gklopper:
and thanks :smile:

philwills:
Ah, OK. I understand now. That also makes sense as to why we haven't seen it as an issue for other services and no problem.

gklopper:
To answer the rest of your question, it was Nathaniel's deploy. The deploy itself went fine (all green), the new boxes just jumped straight to 100% cpu as soon as the old boxes were terminated.

PR #224, opened 2015-01-22
Improve cloudformation check logic

sihil:
- Now the CFN task does creation as well as updates, check for those events
too
- When checking for an update, ensure that the stackname is what we expect


sihil:
This also should fix the premature completion seen with nested stacks @tackley

philwills:
:+1:

PR #225, opened 2015-02-17
[CONTROVERSIAL] Remove build pinning logic

philwills:
I just can't take them any more.

The pins must die!


sihil:
The websys in me is genuinely concerned about the less loved projects that are deployed infrequently as TeamCity only keeps artifacts for a fortnight and then they are gone. Possibly this is the scarring from the time we couldn't roll back the Content API because of exactly this.

philwills:
Oh god, I'm going to end up writing a teamcity project, aren't I?

philwills:
Oh god, I'm going to end up writing a teamcity plugin, aren't I?

philwills:
Having shown by masochistic tendencies by writing a teamcity plugin (https://github.com/guardian/teamcity-s3-plugin), I'm going to close this.

PR #226, opened 2015-03-06
Support S3 prefix configuration

nicl:
The code already supports various options to configure S3 uploads, but they aren't accessible to the end developer! And I'd really like to use them.


sihil:
Yes - let's do that. Can you make the parameter naming match those in the S3 type where it is exposed? (See https://github.com/guardian/deploy/blob/master/magenta-lib/src/main/scala/magenta/deployment_type/S3.scala#L19-L24)

nicl:
@sihil done!

philwills:
Looks good. Do you want some help going through how to test and roll this out?

nicl:
@philwills yeah that would be useful actually. I don't want to break Riff Raff and ruin everyone's day!

nicl:
@philwills @sihil Tested this on CODE, though not thoroughly, and appears to work.

sihil:
:+1: Want to push it out yourself?

PR #228, opened 2015-04-13
Add S3 latency analysis page

sihil:
Given that I wrote it, I thought I may as well check it in to make future analysis easier.


philwills:
:+1:

PR #229, opened 2015-04-29
Template for RiffRaff S3 buckets

philwills:
Includes expiration policy and IAM groups and policies.

Note that I haven't yet tried actually running this.

### DIFF
--- a/cloudformation/riffraff.template
+++ b/cloudformation/riffraff.template
@@ -13,7 +13,7 @@
         "BucketName" : "riffraff-builds",
         "LifecycleConfiguration" : {
           "Rules" : {
-            "ExpirationInDays" : 14,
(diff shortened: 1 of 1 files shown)


sihil:
Looks fine - only question is the motivation for expiring objects? Are simply objects discarded after 14 days? Does that mean it's impossible to deploy something older than 14 days?

philwills:
The artifact bucket doesn't have a lifecycle policy, only those for metadata.  Those dropping away will mean they aren't in auto-completes after a restart, but you should still be able to deploy as the artifact will still be there.

philwills:
I think this should address your concerns @sihil

sihil:
I am now very happy :bowtie: :+1:

PR #231, opened 2015-05-20
For nested stacks, top-level identified by LogicalResourceId

philwills:
StackName which is same for all levels


sihil:
Shiny :+1:

PR #233, opened 2015-05-26
We're already wrapping `buildAt` in a Try, do the same for `buildJsons`

philwills:


philwills:
Incidentally I think returning a `Try[Seq[S3Location]]` and letting the caller deal with recovery here is probably cleaner, but this solves the immediate issue.

PR #235, opened 2015-06-05
Add stack name to deployment report heading 

davidfurey:
Also:
- replaces host with stack in recent deploys table
- adds stack to ASG doubling log message


philwills:
:+1: 

Let me know if you need any help deploying.

PR #237, opened 2015-06-08
Support stage-specific params for cfn deployment types

nicl:
Introduces a new property `templateStageParameters` which supports stage-specific parameters. Currently parameters must be the same across stages, which is somewhat limiting.

Riff Raff will look for a key in the property's map matching the stage of the deployment. If found, it will merge the contained parameters into `templateParameters` (the global parameters if you like), with
stage-specific keys overriding global ones in case of collision.

@philwills @sihil what are your thoughts on this? Happy to discuss.

### DIFF
--- a/magenta-lib/src/main/scala/magenta/deployment_type/CloudFormation.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/CloudFormation.scala
@@ -62,7 +62,7 @@ object CloudFormation extends DeploymentType {
       val fullCloudFormationStackName = cloudFormationStackNameParts.mkString("-")
 
       val globalParams = templateParameters(pkg)
-      val stageParams = stageName.flatMap(stage => templateStageParameters(pkg).lift(stage)).getOrElse(Map())
+      val stageParams = templateStageParameters(pkg).lift(parameters.stage.name).getOrElse(Map())
       val combinedParams = globalParams ++ stageParams
 
       List(
--- a/magenta-lib/src/main/scala/magenta/deployment_type/CloudFormation.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/CloudFormation.scala
@@ -1,7 +1,7 @@
 package magenta.deployment_type
 
-import magenta.{UnnamedStack, NamedStack}
 import magenta.tasks.{CheckUpdateEventsTask, UpdateCloudFormationTask}
+
 import scalax.file.Path
 
 object CloudFormation extends DeploymentType {
@@ -62,7 +62,7 @@ object CloudFormation extends DeploymentType {
       val fullCloudFormationStackName = cloudFormationStackNameParts.mkString("-")
 
       val globalParams = templateParameters(pkg)
-      val stageParams = templateStageParameters(pkg).lift(parameters.stage.name).getOrElse(Map())
+      val stageParams = templateStageParameters(pkg).lift.apply(parameters.stage.name).getOrElse(Map())
       val combinedParams = globalParams ++ stageParams
 
       List(
--- a/magenta-lib/src/main/scala/magenta/deployment_type/CloudFormation.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/CloudFormation.scala
@@ -34,6 +34,9 @@ object CloudFormation extends DeploymentType {
   val templateParameters = Param[Map[String, String]]("templateParameters",
     documentation = "Map of parameter names and values to be passed into template. `Stage` and `Stack` (if `defaultStacks` are specified) will be appropriately set automatically."
   ).default(Map.empty)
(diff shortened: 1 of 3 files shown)


magenta-lib/src/main/scala/magenta/deployment_type/CloudFormation.scala
```diff
@@ -34,6 +34,9 @@ object CloudFormation extends DeploymentType {
   val templateParameters = Param[Map[String, String]]("templateParameters",
     documentation = "Map of parameter names and values to be passed into template. `Stage` and `Stack` (if `defaultStacks` are specified) will be appropriately set automatically."
   ).default(Map.empty)
+  val templateStageParameters = Param[Map[String, Map[String, String]]]("templateStageParameters",
+    documentation = "Like templateParameters, but supports stage-specific configuration."
```
philwills: I think we'd need a more detailed description, though I need to get round to revamping the documentation in RiffRaff to make this info more prominent.

philwills:
I'm still not sure this is a better idea than using mappings in the cloud formation template, but I don't feel massively strongly either way. 

Have you looked how Ophan does it's Cloud Formation these days with nested templates and YAML, transpiled to JSON? It might alleviate some of the pain of that approach.

nicl:
Okay thanks for the feedback. Conceptually I'm trying to treat CFNs like a function and pass in the arguments. I'm less convinced by stage maps. They're good for some things, but can prove restrictive when you want to spin up temporary or new stages - for example TEST stages. And they hamper template reuse between teams too.

And the Ophan approach is definitely nicer than writing JSON, but I'm not sure that YAML really addresses the key problems I see with cloudformation templates - namely the difficulty of reuse (of templates in the entirety, but also of logic and resource groups within). I still wonder whether working at a higher level (i.e. with a programming language) might prove useful here in generating cfn templates. Though, this PR is aimed at solving a different problem.

Having said that, I haven't looked at Ophan in detail, so I'll go away and do that now!

philwills:
I'm just not sure why it's more difficult to add extra parameters into a cfn template than a deploy.json.

The reason I was mentioning the Ophan nested templates is that it helps with segregating different parts of templates, which otherwise turn into hulking masses.

nicl:
Ah sorry, could have been more clear. Once you have formalised a stage, as you say there's no difference in effort. But if you are experimenting, it's nice not to have to hack the template (you just override stuff in the console instead). The other justification though is sharing templates/reuse, which is easier when arguments are passed in instead of hardcoded.

nicl:
@philwills I've expanded the documentation somewhat, so hopefully a bit clearer now.

philwills:
:+1: 

That looks great from a documentation point of view.  Let me know if you need help testing/deploying.

sihil:
There are a handful of issues with parameters - principally secrets and per stage mappings (although that should arguably be stored in the application config in the application or S3 instead). A lot of these should not be checked in. So should we create a mechanism for storing them in S3 in a way that Riff-Raff can look them up?

nicl:
@sihil I agree, neither params nor stage mappings are okay for secrets - we don't currently have an official/standard way of handling secrets, would be great if RiffRaff could do this!

nicl:
Also, re: stage mappings, they are useful when you have a fixed set of options and know it, but otherwise they can prove frustrating when this turns out to not be the case (for example, a stage map when you are experimenting/spinning up new stacks). One approach to mitigate this somewhat is to abstract. For example, rather than a stage map which specifies instance types and asg sizes, abstract into a 'capacity' map ('high' or 'low') and then pass in capacity as a parameter, with a sensible default if there is one. Of course, abstraction is not free though; it's a judgement call.

philwills:
@nicl I'd strongly advise against putting ASG sizes in CloudFormation or a parameter thereof. It makes it much more likely that applying a template will do something you don't want.

nicl:
@philwills yes, sorry that was a bad example - I always forget that!

magenta-lib/src/main/scala/magenta/json/JValueExtractable.scala
```diff
@@ -29,14 +28,16 @@ object JValueExtractable {
     def extract(json: JValue) = extractOption(JArray.unapply)(json) map (_.flatMap(jve.extract(_)))
   }
   implicit def MapExtractable[V](implicit jve: JValueExtractable[V]) = new JValueExtractable[Map[String,V]] {
-    def extract(json: JValue) = {
-      val kvs: List[(String, Option[V])] = for {
-        JObject(fields) <- json
+    def extract(json: JValue): Option[Map[String, V]] = {
+      val JObject(fields) = json
```
nicl: Note, this has been lifted out of the for expression to fix a nasty bug when extracting nested maps. Previously, it would fail, I think because JObject(fields) <- json would act recursively, which caused problems later on. Very subtle and tricky to work out what was going on.

philwills:
:+1:

PR #240, opened 2015-08-01
Correct parameter name in docs

OliverJAsh:


philwills:
Hmm, never really looked at this, but looking at 

https://github.com/guardian/deploy/blob/959bf2e4cb76b73a2fca359554db6d10938cadf6/riff-raff/app/controllers/api.scala

it seems like the request takes `project`, but the response is `projectName`, which seems a bit rubbish.

@sihil am I scanning this incorrectly?

OliverJAsh:
@philwills I tried using the parameter `project` and it didn't work. `projectName` did. I haven't looked at the code :smile:

philwills:
OK, on that basis, I'm willing to :+1:

PR #241, opened 2015-08-26
Fix doc links breaking on github.com

gidsg:
Needed .md extension


philwills:
Sorry @gidsg, just noticed this ages after you put it in. Did you try running it locally to check it still worked within the app?

gidsg:
Oh, I didn't realise the app rendered these pages. I will close this for now as I don't have the credentials to run riff raff locally.

PR #242, opened 2015-09-23
Allow fronts tool account access to buckets

philwills:
The idea is that once we've done this, any builds which a team wants to make just need to create their own credentials to upload to the buckets.

I don't see a way to apply the same policy to multiple buckets, which is annoying.

cc @sihil

### DIFF
--- a/cloudformation/riffraff.template
+++ b/cloudformation/riffraff.template
@@ -27,7 +27,69 @@
       },
       "DeletionPolicy" : "Retain"
     },
-
+    "DelegateAccessArtifactBucketPolicy": {
+      "Type": "AWS::S3::BucketPolicy",
+      "Properties": {
+        "Bucket": { "Ref": "ArtifactBucket" },
+        "PolicyDocument": {
+          "Statement": {
+            "Action": [
+              "s3:Get*",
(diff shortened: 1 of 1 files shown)


rich-nguyen:
Perfect, this is exactly what I need (I think). I can then make my own credentials, and upload to the riff raff buckets?

philwills:
That's the idea.

sihil:
Yes - this is what we do with the pan domain auth config. :+1:

sihil:
Having sat with @rich-nguyen troubleshooting earlier - I wonder if we would have been better off adopting an assume role strategy here. S3 behaves unexpectedly if you put objects from another account, you must explicitly set an ACL on every put request.

Any reason for not adopting AssumeRole here @philwills?

philwills:
Just to round-up the resolution of this, we've changed sbt-riffraff-artifact to transfer ownership to the bucket owner: https://github.com/guardian/sbt-riffraff-artifact/blob/master/src/main/scala/com/gu/riffraff/artifact/RiffRaffArtifact.scala#L126

philwills:
When I say we, I actually mean @rich-nguyen

PR #244, opened 2015-10-12
Check for zero size asg.

kenoir:
Prevents:

> Unhandled exception in task TagCurrentInstancesWithTerminationTag Tag existing instances of the auto-scaling group for termination


philwills:
Hmm. I'm not sure that this check fits with the maxSize check in one conditional, as it feels like a different class of failure. 

However, it's probably better than the current situation, so I'm OK with giving it a :+1: and then trying to clarify later.

sihil:
Any thoughts about whether you still want to merge this @kenoir?

kenoir:
@sihil I still think it makes more sense than the current way of handling the situation when the `desiredCapacity` is 0.

sihil:
@kenoir I'm trying to make the ASG deploy work when an ASG size is set to 0. Do you recall what this was solving?

kenoir:
@sihil this change simply makes the error message more explicit.

sihil:
OK - so if I can make the error not happen then the world will be a happy place? 💭

kenoir:
That would be much better.

PR #245, opened 2015-10-16
Allow composer account access to buckets

philwills:
This has already been applied, so just an FYI.


jamespamplin:
Thanks @philwills, working for me.

Issue #246, opened 2015-10-30
Riff-Raff continuous deployment not firing...


rtyley:
@philwills,  @tomverran and myself noticed that Subscriptions Frontend is not getting continuously deploye :crying_cat_face: 

For instance, build #1173, started in TeamCity at 30 Oct 15 15:03 (and the previous #1170), did not fire in Riffraff:

#### RiffRaff history
https://riffraff.gutools.co.uk/deployment/history?projectName=Subscriptions%3A%3Afrontend#
![image](https://cloud.githubusercontent.com/assets/52038/10850396/4cc34d38-7f1d-11e5-8565-21d27728efad.png)

#### Teamcity
https://teamcity.gutools.co.uk/viewType.html?buildTypeId=Subscriptions_Frontend&branch_Subscriptions=%3Cdefault%3E&tab=buildTypeStatusDiv
![image](https://cloud.githubusercontent.com/assets/52038/10850407/58589b6c-7f1d-11e5-853b-885963524123.png)

#### RiffRaff CD is set
https://riffraff.gutools.co.uk/deployment/continuous/c4425e1a-9b84-411c-8b9e-ed82ee72b570/edit

![image](https://cloud.githubusercontent.com/assets/52038/10850441/a2ec44e4-7f1d-11e5-8b26-3225e863c48c.png)

philwills:
Good news, there's a correlated stack trace, so I have a clue about how to fix!

Bad news, it's from some really shonky not quite async code, so it might take a bit of work.

philwills:
I think this has largely been addressed by #247. In the longer term we could do with revisiting that logic, but I'm going to close this for now. Feel free to reopen if you disagree.

PR #247, opened 2015-11-02
CD deploys shouldn't propagate errors

philwills:
A future timing out or the like should not cause all CD to stop


cb372:
Is this a fix for #246 ? CD on my project (https://riffraff.gutools.co.uk/deployment/history?projectName=Content+Platforms%3A%3Abonobo&pageSize=20&page=1) seems to be broken at the moment as well.

philwills:
@cb372 It's an attempt to ameliorate rather than fix. That broke everything, whereas this would mean only the one deploy failed. 

I'm currently waiting for an opportunity to hard bounce RiffRaff to get it working again for the moment.

PR #249, opened 2015-11-09
Added Membership to Accout access

chrisjowen:
Added gu-membership account access to the riffraff buckets


philwills:
This has been applied.

PR #253, opened 2015-11-25
Lambda deployment type

gidsg:
Adds a deployment type `aws-lambda` which provides one deploy action, `updateLambda`, that runs Lambda Update Function Code using the package files which can be specified per stage.

CC @guardian/discussion @philwills

### DIFF
--- a/magenta-lib/src/main/scala/magenta/deployment_type/Lambda.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/Lambda.scala
@@ -8,14 +8,14 @@ object Lambda extends DeploymentType  {
   val name = "aws-lambda"
   val documentation =
     """
-      |Provides one deploy action, `updateLambda`, that runs Lambda Update Function Code using the package file which should be a single file named lambda.zip.
+      |Provides one deploy action, `updateLambda`, that runs Lambda Update Function Code using the package files which can be specified per stage.
     """.stripMargin
   
 
   //required configuration, you cannot upload without setting these
   val functions = Param[Map[String, Map[String, String]]]("functions",
     documentation =
-      """Map of Stage to Lambda function names.
+      """Map of Stage to Lambda functions. 'filename' field is optional and if not specified defaults to lambda.zip
         |e.g.
         |        "functions": {
         |          "CODE": {
--- a/magenta-lib/src/main/scala/magenta/deployment_type/Lambda.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/Lambda.scala
@@ -1,6 +1,7 @@
 package magenta.deployment_type
 
 import java.io.File
+import magenta.MessageBroker
 import magenta.tasks.UpdateLambda
 
 object Lambda extends DeploymentType  {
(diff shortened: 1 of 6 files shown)


kenoir:
Nice!

akash1810:
Nice! Was just thinking how we'd deploy lambdas on the Grid too!

magenta-lib/src/main/scala/magenta/deployment_type/Lambda.scala
```diff
@@ -0,0 +1,37 @@
...
+      """.stripMargin
+  ).default(Map.empty)
+
+  def perAppActions = {
+    case "updateLambda" => (pkg) => (resourceLookup, parameters, stack) => {
+      implicit val keyRing = resourceLookup.keyRing(parameters.stage, pkg.apps.toSet, stack)
+      val fNames = functionNames(pkg)
+      val stage = parameters.stage.name
+      val fName = fNames get stage
+      assert(fName.isDefined, s"functionName must be defined for stage $stage")
```
philwills: The normal pattern in RiffRaff is to use `MessageBroker.fail`, e.g. https://github.com/guardian/deploy/blob/master/magenta-lib/src/main/scala/magenta/tasks/AWS.scala#L122

magenta-lib/src/main/scala/magenta/tasks/AWS.scala
```diff
@@ -7,17 +7,55 @@ import com.amazonaws.services.ec2.AmazonEC2Client
...
 
 trait S3 extends AWS {
   def s3client(keyRing: KeyRing) = new AmazonS3Client(credentials(keyRing))
 }
 
+trait Lambda extends AWS {
+
+  def lambdaClient(implicit keyRing: KeyRing) = {
+    val client = new AWSLambdaClient(credentials(keyRing))
+    client.setEndpoint("lambda.eu-west-1.amazonaws.com")
```
philwills: Eek, we should stop hardcoding these at some point, but it's fine for the purpose of this PR.

magenta-lib/src/main/scala/magenta/tasks/AWS.scala
```diff
@@ -7,17 +7,55 @@ import com.amazonaws.services.ec2.AmazonEC2Client
...
+    val request = new UpdateFunctionCodeRequest
+    request.withFunctionName(functionName)
+    request.withZipFile(buffer)
+    request
+  }
+}
+
+object Lambda {
+  def lambdaUpdateFunctionCodeRequest(functionName: String, file: File): UpdateFunctionCodeRequest = {
+
```
philwills: How come we have this function in the trait and the object?

philwills:
Apart from those couple of minor points, this looks good.

janua:
@gidsg How does this work if you have a standalone project?

gidsg:
I talked with @janua IRL but please see example project https://github.com/guardian/discussion-platform/tree/riff-raff-lambda/cachebust

magenta-lib/src/main/scala/magenta/deployment_type/Lambda.scala
```diff
@@ -0,0 +1,37 @@
...
+    List(UpdateLambda(new File(pkg.srcDir.getPath + "/lambda.zip"), fName.get))
```
janua: @gidsg Would it be possible to put this into a parameter?

gidsg: What is the benefit of this v.s. extra config? A bit easier build script?

janua: More explicit instead of an arbitrary filename.

philwills: RiffRaff params can have defaults, so we can have some flexibility without demanding extra config.

gidsg:
@philwills I made the changes you suggested.

In addition, I have done a refactor with @janua which allows for the following:
* Optionally specify Lambda filenames
* Allow for different artefacts per stage. e.g. for @janua's use case of injecting different config per stage.

Please take a look and let me know if you are OK with these changes.

janua:
Yes!! :+1:

philwills:
:+1: great work @gidsg

PR #257, opened 2016-01-25
Update play-googleauth

adamnfish:
Bump to latest version of google auth library with forced domain check.


sihil:
:+1: Thanks. Seems that riff-raff is also failing to apply the authentication list so I'm going to look at that too.

PR #258, opened 2016-01-25
Check whether a user is authorised

sihil:
I discovered today that riff-raff has not been applying Authorisation correctly. I suspect this happened when it was converted to the play-googleauth library as the authorisation code was massively overhauled.

Anyhow - this one little trick applies the authorisation check.


adamnfish:
Great spot!

Writing critical auth logic at the top as a side effect and having it fall through to the rest of the code is poor from a defensive coding point of view though. It also makes it very easy to forget it (see also, before this PR). It's all extremely side-effecty already so :+1: since it is consistent, but it might be worth changing this to be more "happy path" at some point.

PR #259, opened 2016-01-28
Format exception lines nicely in the browser

janua:
Upon receiving an exception on the riffraff deploy screen, the exception lines are meant to print out with newlines in between, but this doesn't work and the browser just ignores the newlines. You receive just a big long string.

This wraps each exception line in a `span`.


sihil:
Ah yes - this used to work at some point (I'm sure `pre` used to honour newlines). 

Thanks for the fix :)  :+1:

PR #260, opened 2016-02-05
Use existing CloudFormation params

sihil:
The AWS SDK now makes it possible to specify in an update call to a cloudformation stack that existing values should be reused.

This PR changes the behaviour so that this flag is automatically set for any parameter in a cloudformation template that has neither a default value, a value provided by the user nor a value generated by riff-raff itself (eg. Stack/Stage).

This following screenshot illustrates this at work. The CertificateARN parameter has no default or anything specified so is passed in as UseExistingValue.

screen shot 2016-02-05 at 12 07 06

### DIFF
--- a/magenta-lib/src/main/scala/magenta/tasks/UpdateCloudFormationTask.scala
+++ b/magenta-lib/src/main/scala/magenta/tasks/UpdateCloudFormationTask.scala
@@ -11,23 +11,29 @@ import collection.convert.wrapAsScala._
 case class UpdateCloudFormationTask(
   cloudFormationStackName: String,
   template: Path,
-  parameters: Map[String, String],
+  parameters: Map[String, Option[String]],
   stage: Stage,
   stack: Stack,
   createStackIfAbsent:Boolean)(implicit val keyRing: KeyRing) extends Task {
+  
+  case class TemplateParameter(key:String, default:Boolean)
 
   def execute(stopFlag: => Boolean) = if (!stopFlag) {
-    val requiredParameters = CloudFormation.validateTemplate(template.string).getParameters.map(_.getParameterKey)
+    val templateParameters = CloudFormation.validateTemplate(template.string).getParameters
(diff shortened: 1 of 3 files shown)


magenta-lib/src/main/scala/magenta/tasks/UpdateCloudFormationTask.scala
```diff
@@ -8,6 +8,13 @@ import com.amazonaws.services.cloudformation.model._
 import scalax.file.Path
 import collection.convert.wrapAsScala._
 
+object UpdateCloudFormationTask {
+  trait ParameterValue
+  case class StringValue(value: String) extends ParameterValue
+  case class UseExistingValue() extends ParameterValue
```
philwills: I'd make this a `case object` and the trait `sealed` so that the compiler can do exhaustivity checking.

magenta-lib/src/main/scala/magenta/tasks/UpdateCloudFormationTask.scala
```diff
@@ -8,6 +8,13 @@ import com.amazonaws.services.cloudformation.model._
 import scalax.file.Path
 import collection.convert.wrapAsScala._
 
+object UpdateCloudFormationTask {
+  trait ParameterValue
+  case class StringValue(value: String) extends ParameterValue
```
philwills: Would `SpecifiedValue` be a more informative name for this?

philwills:
A + :100:  for adding some tests for this and :+1: to it generally give or take the minor style quibbles.

magenta-lib/src/main/scala/magenta/tasks/UpdateCloudFormationTask.scala
sihil: Yes, much better. thanks.

magenta-lib/src/main/scala/magenta/tasks/UpdateCloudFormationTask.scala
sihil: agreed.

PR #261, opened 2016-02-09
Add support for YAML CloudFormation templates

cb372:
If the template file has a `.yml` or `.yaml` extension, it is assumed to be in YAML format and RiffRaff will convert it to JSON on the fly. Otherwise it is assumed to be in JSON format and will be used as-is.

### DIFF
--- a/magenta-lib/src/main/scala/magenta/tasks/UpdateCloudFormationTask.scala
+++ b/magenta-lib/src/main/scala/magenta/tasks/UpdateCloudFormationTask.scala
@@ -1,6 +1,9 @@
 package magenta.tasks
 
 import com.amazonaws.AmazonServiceException
+import com.fasterxml.jackson.core.util.DefaultPrettyPrinter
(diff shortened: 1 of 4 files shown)


magenta-lib/src/main/scala/magenta/tasks/UpdateCloudFormationTask.scala
```diff
@@ -15,6 +18,26 @@ object UpdateCloudFormationTask {
...
 }
 
+object JsonConverter {
+
+  /**
+    * Return the template's content as JSON,
+    * converting it from YAML if necessary.
+    */
+  def convert(template: Path): String = template.extension match {
+    case Some("yml") | Some("yaml") =>
```
adamnfish: I didn't know that `|` syntax! Thank you!

jamespamplin:
:tada: nice one!

davidfurey:
Works in CODE.

cb372:
@davidfurey thanks for testing!

PR #262, opened 2016-02-11
Fix extension parsing

davidfurey:
Correctly parses the extension of filenames containing dots.

### DIFF
--- a/magenta-lib/src/main/scala/magenta/tasks/tasks.scala
+++ b/magenta-lib/src/main/scala/magenta/tasks/tasks.scala
@@ -147,17 +147,13 @@ case class S3Upload( stack: Stack,
     s"$stackName$stageName${toRelative(file)}"
   }
 
-  def contentTypeLookup(fileName:String) = {
-    import org.parboiled2._
-    import org.parboiled2.CharPredicate._
-    class ExtensionParser(val input: ParserInput) extends Parser {
-      def FileName = rule { oneOrMore(Visible -- '.') ~ '.' ~ Extension ~ EOI }
-      def Extension = rule { capture(oneOrMore(Visible)) ~> (s => mimeTypeMap.get(s)) }
-    }
-    new ExtensionParser(fileName).FileName.run().getOrElse(None)
-  }
+  def contentTypeLookup(fileName: String) =
+    fileExtension(fileName).flatMap(mimeTypeMap.get)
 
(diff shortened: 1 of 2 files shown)


sihil:
Your parser combinator magic is being torn to shreds @philwills. Are you just going to stand by and put up with this?

cb372:
Bring back the regex!

philwills:
I actually suggested it based on @adamnfish suggestion.

I know it's hard to believe, but occasionally, I can be convinced that a parser combinator isn't the best solution.

magenta-lib/src/main/scala/magenta/tasks/tasks.scala
```diff
@@ -147,18 +147,14 @@ case class S3Upload( stack: Stack,
...
-    }
-    new ExtensionParser(fileName).FileName.run().getOrElse(None)
-  }
+  def contentTypeLookup(fileName: String) =
+    fileExtension(fileName).flatMap(mimeTypeMap.get)
 
   def cacheControlLookup(fileName:String) = cacheControlPatterns.find(_.regex.findFirstMatchIn(fileName).isDefined).map(_.value)
 
+  private def fileExtension(fileName: String) =
+    fileName.split('.').drop(1).lastOption
```
philwills: Why do you need the `drop`, if you're doing `lastOption`

davidfurey: So that files without a dot do not return the filename as the extension

philwills: Ah, that makes sense. I forgot that `split` included the original if it didn't find the character.

sihil:
Treachery.

magenta-lib/src/main/scala/magenta/tasks/tasks.scala
adamnfish: Rather than writing your own stuff to handle filenames and then realising it is hard and buggy, why not use a lib to do it (e.g. `FilenameUtils`?)

davidfurey:
We did fix the original parboiled version, but it didn't seem very clear any more:
https://github.com/guardian/riff-raff/commit/005ac509a149ca45cacb7e5facb75171fb642db2

magenta-lib/src/main/scala/magenta/tasks/tasks.scala
davidfurey: It seemed like overkill to add another dependency just for this.

sihil:
:+1:

PR #264, opened 2016-02-29
Bump Casbah so that we work with MongoDB 3.2.1

sihil:
compose.io provides MongoDB 3.2.1 so the libraries we use must be compatible with that version. We're bumping to the latest version of Casbah as there is no reason to do otherwise.

### DIFF
--- a/riff-raff/build.sbt
+++ b/riff-raff/build.sbt
@@ -6,8 +6,6 @@ resolvers ++= Seq(
     "Typesafe Releases" at "http://repo.typesafe.com/typesafe/releases/"
 )
 
-ivyScala := ivyScala.value map { _.copy(overrideScalaVersion = true) }
-
 libraryDependencies ++= Seq(
   "com.gu" %% "management-play" % guardianManagementPlayVersion exclude("javassist", "javassist"), // http://code.google.com/p/reflections/issues/detail?id=140
   "com.gu" %% "management-logback" % guardianManagementVersion,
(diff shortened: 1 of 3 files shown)


riff-raff/build.sbt
```diff
@@ -6,13 +6,15 @@ resolvers ++= Seq(
     "Typesafe Releases" at "http://repo.typesafe.com/typesafe/releases/"
 )
 
+ivyScala := ivyScala.value map { _.copy(overrideScalaVersion = true) }
```
philwills: What's this?

philwills:
Otherwise +1

riff-raff/build.sbt
sihil: Good question. I've added it because a warning told me to :grimacing:.

I think the reason for the warning is that different libraries try to pull in different versions of scala and this overrides that. There might be a better approach, but a search of guardian repos showed this was common practice.

philwills: A quick Google suggests it's down to https://github.com/guardian/riff-raff/blob/sh-bump-casbah/build.sbt#L3

We should try bumping that to 2.11.7 and getting rid of this.

sihil: Ah - that does seem to work. So it's only when you pull in a library that is using a newer scala than your build. That makes it eminently fixable. It's not a great warning message. I avoided adding it for ages, but then I saw [others were using it](https://github.com/search?utf8=%E2%9C%93&q=ivyScala+user%3Aguardian&type=Code&ref=searchresults) and figured I'd join them (just realised that most of those are my own team).

PR #265, opened 2016-02-29
Ami cloudformation parameter

sihil:
Add a type that can update the AMI parameter of a cloudformation stack without having the template in your stack. We use Prism to do the AMI lookup in order to allow easy lookups across AWS accounts.


magenta-lib/src/main/scala/magenta/Lookup.scala
```diff
@@ -73,4 +74,6 @@ case class DeployInfoLookupShim(deployInfo: DeployInfo, secretProvider: SecretPr
   }
 
   def stages = deployInfo.knownHostStages
+
+  def getLatestAmi(region: String)(tags: Map[String, String]): Option[String] = ???
```
adamnfish: Why isn't this implemented?

related: Good to use the override keyword if you're overriding because you get the compile-time check that it actually overrides something.

magenta-lib/src/test/scala/magenta/deployment_type/CloudFormationTest.scala
```diff
@@ -21,32 +21,30 @@ class CloudFormationTest extends FlatSpec with Matchers {
...
+        inside(updateTask) {
+          case UpdateCloudFormationTask(stackName, path, userParams, amiParam, amiTags, _, stage, stack, ifAbsent) =>
+            stackName should be(cfnStackName)
+            path should be(Path(p.srcDir) \ Path.fromString( """cloud-formation/cfn.json"""))
+            userParams should be(Map.empty)
+            amiParam should be("AMI")
+            amiTags should be(Map.empty)
+            stage should be(PROD)
+            stack should be(NamedStack("cfn"))
+            ifAbsent should be(true)
```
adamnfish: scalatest has a nice have syntax which you may prefer? It let's you check multiple properties on a thing at once. It's a bit less typesafe (because have takes symbols which it uses to look up the value) but I tend to use it for testing multiple properties on a case class because in tests clarity of meaning is the most important thing.

```scala
inside(updateTask) { ufct =>
  ucft should have (
    'stackName (cfnStackName),
    'propertyName (value)
    // etc
  )
}
```

PR #266, opened 2016-03-01
Send webhook body as json if valid json

jharewinton:
Utilising the hacking skills of @philwills, this will set the body of webhooks to json if it fits a valid json format. This to allow passing parameters to Circle CI as they require the header to be set to "Content-Type: application/json"


sihil:
:+1:

PR #270, opened 2016-03-03
Self deployment and version visibility

sihil:
Make self deployment work inside the AWS environment.

This is connected with two other PRs: https://github.com/guardian/machine-images/pull/95 and https://github.com/guardian/deploy-tools-platform/pull/10. 

This PR also makes the build ID visible again at the bottom right of each page:
image

### DIFF
--- a/magenta-lib/src/main/scala/magenta/deployment_type/SelfDeploy.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/SelfDeploy.scala
@@ -1,0 +1,71 @@
+package magenta.deployment_type
+
+import java.io.File
+
+import magenta.tasks.{ChangeSwitch, S3Upload}
+
+object SelfDeploy extends DeploymentType with S3AclParams {
+  val name = "self-deploy"
+  val documentation =
+    """
+      |This is a special deployment type that can be used to self-deploy Riff-Raff.
+      |
+      |The approach is to upload a file to S3 and then hit a switch on the singleton instance.
+      |When the switch is hit it will shutdown with a specific exit code at the next
+      |opportunity (when it has finished deploys).
+    """.stripMargin
+
+  val bucket = Param[String]("bucket",
+    """
+      |S3 bucket name to upload artifact into.
+      |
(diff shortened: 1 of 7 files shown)


project/MagentaBuild.scala
```diff
@@ -41,4 +51,28 @@ object MagentaBuild extends Build {
   ) ++ net.virtualvoid.sbt.graph.Plugin.graphSettings
 
   val magentaVersion = "1.0"
+
+
+  implicit def string2Dequote(s: String): Object {val dequote: String} = new {
+    lazy val dequote = s.replace("\"", "")
+  }
+
+  def buildFile(outDir: File, branch: String, buildNum: String, vcsNum: String, s: TaskStreams) = {
```
philwills: What's the reason for not using https://github.com/sbt/sbt-buildinfo?

magenta-lib/src/main/scala/magenta/deployment_type/SelfDeploy.scala
```diff
@@ -0,0 +1,71 @@
...
+
+  val bucket = Param[String]("bucket",
+    """
+      |S3 bucket name to upload artifact into.
+      |
+      |The path in the bucket is `<stack>/<stage>/<packageName>/<fileName>`.
+    """.stripMargin
+  )
+
+  val prefixStage = Param[Boolean]("prefixStage",
```
philwills: We should probably try and cut-down on the c&p of these params some time.

philwills:
:+1:

project/MagentaBuild.scala
sihil: I wasn't aware of it. Will have a look at refactoring to use that instead.

magenta-lib/src/main/scala/magenta/deployment_type/SelfDeploy.scala
sihil: When I was reviewing this I thought I might actually get rid of them entirely for this deploy type. I can't see why the values would ever need to be customised.

PR #274, opened 2016-03-15
Fix yaml deployment

alexduf:
- this PR _should_ fix the deployment of yaml cloudformation templates
- it also adds bit more information in the case the JSON is invalid

It needs to be tested in a CODE environment, so this should be blocked until we have a CODE environment in AWS for riff-raff

cc @cb372

### DIFF
--- a/magenta-lib/src/main/scala/magenta/tasks/UpdateCloudFormationTask.scala
+++ b/magenta-lib/src/main/scala/magenta/tasks/UpdateCloudFormationTask.scala
@@ -83,14 +83,17 @@ case class UpdateCloudFormationTask(
 
     if (CloudFormation.describeStack(cloudFormationStackName).isDefined)
       try {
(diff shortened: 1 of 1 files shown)


sihil:
Is anyone using this conversion in PROD? If it's not being used heavily then I think it should be OK to deploy this to PROD to test until we have a chance to build a CODE environment.

alexduf:
we are for MAPI, not sure about anyone else. @cb372 do you use it?

cb372:
I see, it looks like part of my change was lost in a dodgy git merge when #265 was merged into master (see https://github.com/guardian/riff-raff/pull/265/files#diff-03de60f2ddca48309519c19d115429fcR86). And this PR reinstates my lost changes, so :+1: 

I haven't heard of anyone apart from MAPI using this feature yet. But either way, the feature is currently broken and I'm happy that this PR fixes it, so I say :shipit:

PR #280, opened 2016-04-15
Use DynamoDB for continuous deployment config

philwills:
cc @adamnfish @sihil 

One question is whether we should put the table definition in the associated cloudformation template or not. On Ophan we came to the conclusion that it was often more trouble than it's worth, but I'm happy to be convinced otherwise.

### DIFF
--- a/riff-raff/app/persistence/ContinuousDeploymentConfigRepository.scala
+++ b/riff-raff/app/persistence/ContinuousDeploymentConfigRepository.scala
@@ -1,0 +1,40 @@
+package persistence
+
+import java.util.UUID
+
+import cats.data.Validated
+import ci.{ContinuousDeploymentConfig, Trigger}
+import com.gu.scanamo.{DynamoFormat, Scanamo, TypeCoercionError}
+import com.gu.scanamo.syntax._
+import conf.Configuration
+import org.joda.time.DateTime
+
+object ContinuousDeploymentConfigRepository {
+  val tableName = "continuous-deployment-config"
+  val client = Configuration.continuousDeployment.dynamoClient
+
+  implicit val uuidFormat = DynamoFormat.xmap[UUID, String](
+    s => Validated.catchOnly[IllegalArgumentException](UUID.fromString(s)).leftMap(TypeCoercionError(_)).toValidatedNel
+  )(_.toString)
+
+  implicit val jodaStringFormat = DynamoFormat.xmap[DateTime, String](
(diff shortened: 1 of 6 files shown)


riff-raff/app/persistence/ContinuousDeploymentConfigRepository.scala
```diff
@@ -0,0 +1,40 @@
...
+import cats.data.Validated
+import ci.{ContinuousDeploymentConfig, Trigger}
+import com.gu.scanamo.{DynamoFormat, Scanamo, TypeCoercionError}
+import com.gu.scanamo.syntax._
+import conf.Configuration
+import org.joda.time.DateTime
+
+object ContinuousDeploymentConfigRepository {
+  val tableName = "continuous-deployment-config"
+  val client = Configuration.continuousDeployment.dynamoClient
```
adamnfish: I'm not a huge fan of doing IO from an Object constructor unless that Object is the 'edge' of the program (where it can't be helped). Better to take this as an arg so the dirty edge can do the IO of fetching it?

adamnfish:
I'm not totally on top of the scanamo use but looks good otherwise (my comment excepted).

It's nice to document the table structure, perhaps a separate CF template from the app (which I assume is where the hassles come from) that outputs the table which the app can take as a param?

riff-raff/app/persistence/ContinuousDeploymentConfigRepository.scala
philwills: I have mixed feelings and the the ScanamoFree interface is designed to suit that model. 

In this case, since I'm directly replacing an existing interface, this was by far the path of least resistance.

philwills:
See also https://github.com/guardian/deploy-tools-platform/pull/27

riff-raff/test/MigrateCDConfigs.scala
```diff
@@ -0,0 +1,7 @@
+import persistence.{ContinuousDeploymentConfigRepository, Persistence}
+
+object MigrateCDConfigs {
+  def main(args: Array[String]) {
+    Persistence.store.getContinuousDeploymentList().foreach(ContinuousDeploymentConfigRepository.setContinuousDeployment)
+  }
+}
```
sihil: Historically I've done this in the main app by checking a migration key using this collection in mongo: https://github.com/guardian/riff-raff/blob/master/riff-raff/app/persistence/datastore.scala#L64-L66

Then run the migration as a one off if the key is missing.

This would also work.

sihil:
That was less code than I expected.

👍 But I'll leave you and @adamnfish to argue over the purity levels. <spectating>.

PR #285, opened 2016-05-12
Make times in drop down respect BST

philwills:


riff-raff/app/controllers/DeployController.scala
```diff
@@ -149,12 +149,13 @@ object DeployController extends Controller with Logging with LoginActions {
...
 
+  val shortFormat = DateTimeFormat.forPattern("HH:mm d/M/yy").withZone(DateTimeZone.forID("Europe/London"))
+
   def autoCompleteBuild(project: String, term: String) = AuthAction {
     val possibleProjects = Builds.successfulBuilds(project).filter(
       p => p.number.contains(term) || p.branchName.contains(term)
     ).map { build =>
-      val formatter = DateTimeFormat.forPattern("HH:mm d/M/yy")
-      val label = "%s [%s] (%s)" format(build.number, build.branchName, formatter.print(build.startTime))
+      val label = "%s [%s] (%s)" format(build.number, build.branchName, shortFormat.print(build.startTime))
```
adamnfish: Why use format over string interpolation?

adamnfish:
:+1: Excellent

riff-raff/app/controllers/DeployController.scala
philwills: Good question. I suspect just because this is quite old code.

sihil:
Ideally this should respect the user's timezone using the locale in the request (times rendered locally using moment do this). I had a brief look at this, but it looked non-trivial with the way the templates were arranged.

philwills:
@sihil That does sound ideal, but I don't want the best to be the enemy of the good

PR #287, opened 2016-05-12
Add an S3 based lambda deployment type

sihil:
The existing lambda deployment task doesn't match what is available in CloudFormation (where the archive has to come form S3). As such I've knocked this together.

It is strongly opinionated in so far as it will upload the file provided to the bucket given using a generated path (`s3://$bucket/$stack/$stage/$package/$fileName`). The update action updates the lambda function with the name `$functionName$stage`. i.e. If `functionName` is `MyFunction-` and you are deploying to CODE then it will become `MyFunction-CODE`. If will update this function name (which must already exist) using a file in S3 generated in the same way as the upload function.


philwills:
I'm a bit nervous of having two different lambda deployment tasks rather than variants of one.

sihil:
I considered doing it as one task - but my experience in the past is that the tasks become complex. I also don't really like the way that the existing task encourages per stage configuration so thought I'd start over with something simpler.

philwills:
Should we see if we can convince existing Lambda deployers to move over to this model then?

sihil:
After further consideration (and finding that it wasn't too complex) I've merged this into one task. Have also caught up with @gidsg. Whilst I still consider stage-based artifacts to be an anti-pattern, the limitations in AWS Lambda means that it is one of the only reasonable ways to configure lambdas at the moment.

philwills:
👍

PR #290, opened 2016-05-17
Replace the S3Upload task with a new S3UploadV2 task

sihil:
The `S3Upload` task does a little too much internally which makes it hard to reason about and hard to use when you want to reliably work out where it has uploaded files. The replacement V2 doesn't do as much magic and moves the responsibility of mapping files to keys (or directories to key prefixes) to the callee.

At the moment I've only replaced the use of it in the lambda upload action, but would like to completely replace and remove `S3Upload`.

### DIFF
--- a/magenta-lib/src/main/scala/magenta/deployment_type/Lambda.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/Lambda.scala
@@ -2,8 +2,8 @@ package magenta.deployment_type
 
 import java.io.File
 
-import magenta.{DeploymentPackage, KeyRing, MessageBroker, Stack, Stage}
-import magenta.tasks.{S3Upload, UpdateLambda, UpdateS3Lambda}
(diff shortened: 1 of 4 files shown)


philwills:
I think that making the S3Upload task more tightly focused is a good idea. I'm less keen on the idea of having the two both in the code base at the same time. 

If that feels a bit hefty, maybe a BIG comment saying, if you are doing anything new, use the other one.

sihil:
How about I deprecate the old one? That way we'll have a bunch of compiler warnings and it's more likely people will notice. I will also add an issue and an item to my todo list.

philwills:
Yea, I had forgotten that we don't have fatal warnings on (we should), but given that we don't that seems a reasonable approach.

sihil:
OK - I've done that. Any further comments?

philwills:
A belated :shipit:

Issue #291, opened 2016-05-18
Remove S3Upload from magenta-lib


sihil:
PR #290 introduces a new S3UploadV2 task that simplifies the S3Upload task and hopes to make it easier to reuse. The PR deprecates the existing task but does not go so far as to switch over all of the existing uses to the new task. This is left as a follow up PR once the new task has had a few days to settle.

philwills:
I consider #292 to close this

PR #292, opened 2016-05-21
Switch out use of S3Upload for S3UploadV2 throughout

sihil:
This addresses #291 by switching all uses over to the cleaner S3UploadV2 and deleting the unused task. All of the tests pass, but I think this still needs some care when rolling out - it impacts the `self-deploy` (i.e. Riff-Raff's deploy), `aws-s3`, `autoscaling`, `elasticsearch`.

(It also impacts the `uploadArtifacts` actions on the `executable-jar-webapp`, `native-packager-webapp` and `jetty-webapp` deployment types - although I very much doubt that any of those are still in use).


magenta-lib/src/main/scala/magenta/deployment_type/S3.scala
```diff
@@ -106,18 +107,18 @@ object S3 extends DeploymentType with S3AclParams {
         assert(data.isDefined, s"Cannot find resource value for ${bucketResource(pkg)} (${pkg.apps.head} in ${parameters.stage.name})")
         data.get.value
       }
+      val prefix:String = S3UploadV2.prefixGenerator(
+        stack = if (prefixStack(pkg)) Some(stack) else None,
+        stage = if (prefixStage(pkg)) Some(parameters.stage) else None,
+        packageName = if (prefixPackage(pkg)) Some(pkg.name) else None
+      )
```
jennysivapalan: This looks similar to the logic in `Autoscaling.scala`, is it worth `S3UploadV2` having a method that has `pkg` as the parameter and return the result of `prefixGenerator`?

sihil: It should be identical. The problem is that it also needs the parameters from the class, so it's quite difficult to dry out without passing in seven parameters. I could extract it into a param trait.

sihil: After more thought I've decided not to bother DRYing this out. It is minimal code duplication.

jennysivapalan: Agreed - don't want to make you do more work for a few lines!

sihil:
I've tested this locally with autoscaling deploys and will test the self deploy as soon as I've pushed out this PR.

Any thoughts @philwills

philwills:
Any reason to keep the V2 suffix? Looks like this could just be the new S3Upload.

sihil:
Nope - just renamed it.

philwills:
👍 👍 👍

PR #293, opened 2016-05-23
Logging without thread local

sihil:
This is a bash at removing the dependency on thread local logging. The approach is to pass around a `DeployLogger` (née `MessageBroker`) everywhere that needs it. This is largely used during task resolution and then task execution.

I've got as far as all of the test suites passing - but this has not been tested outside of that and if memory serves me right, logging has historically been somewhat brittle. Having said that, the changes made here feel generally good.

### DIFF
--- a/magenta-cli/src/main/scala/magenta/cli/Main.scala
+++ b/magenta-cli/src/main/scala/magenta/cli/Main.scala
@@ -21,7 +21,7 @@ import scalax.file.Path
 object Main extends scala.App {
 
   var taskList: List[TaskDetail] = _
-  MessageBroker.messages.subscribe(wrapper => {
+  DeployLogger.messages.subscribe(wrapper => {
     val indent = "  " * (wrapper.stack.messages.size - 1)
     wrapper.stack.top match {
       case Verbose(message) => if (Config.verbose) Console.out.println(indent + message)
@@ -80,7 +80,7 @@ object Main extends scala.App {
     def localArtifactDir_=(dir: Option[File]) {
       dir.foreach { f =>
         if (!f.exists() || !f.isDirectory) sys.error("Directory not found.")
-        MessageBroker.info("WARN: Ignoring <project> and <build>; using local artifact directory of " + f.getAbsolutePath)
+        System.err.println("WARN: Ignoring <project> and <build>; using local artifact directory of " + f.getAbsolutePath)
       }
 
       _localArtifactDir = dir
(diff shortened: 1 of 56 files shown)


philwills:
I'm not going to claim to have looked at all this, but a massive 👏 for attempting it.  

One quibble, is that I'm not convinced by using the name `logger` without any further context for this. I think that's too easily assumed to be some kind of slf4j or the like object.

sihil:
Yeah - I'd agree with that. I thought `DeployLogger` was better than `MessageBroker` - but it did cross my mind that the shortening to `logger` lost the context. Better suggestions are more than welcome.

philwills:
`DeployReporter`?

sihil:
I've renamed as suggested by @philwills.

Also I've just run it locally to deploy Prism. Seems to have worked first time - so now I'd like some serious feedback as rebasing this is a nightmare!

screencapture-localhost-9000-deployment-view-8323593f-c480-403d-aaf7-578c97964fa0-1464213316292

magenta-cli/src/main/scala/magenta/cli/Main.scala
```diff
@@ -169,48 +169,47 @@ object Main extends scala.App {
...
-                  secret <- Config.secretKey
-                } yield new BasicAWSCredentials(key, secret)).getOrElse(null)
-
-                override def refresh(): Unit = {}
-              }, new DefaultAWSCredentialsProviderChain())
-            ))
-          }
+        val deployId = UUID.randomUUID()
+        implicit val rootLogger = DeployReporter.rootReporterFor(deployId, parameters)
+        rootLogger.info("[using %s build %s]" format (programName, programVersion))
```
philwills: Should probably rename this.

magenta-lib/src/main/scala/magenta/deployment_type/AmiCloudFormationParameter.scala
```diff
@@ -28,7 +28,7 @@ object AmiCloudFormationParameter extends DeploymentType {
   ).default("AMI")
 
   override def perAppActions = {
-    case "update" => pkg => (lookup, parameters, stack) => {
+    case "update" => pkg => (logger, lookup, parameters, stack) => {
```
philwills: reporter

magenta-lib/src/main/scala/magenta/deployment_type/AutoScaling.scala
```diff
@@ -85,7 +85,7 @@ object AutoScaling  extends DeploymentType with S3AclParams {
   ).default(true)
 
   def perAppActions = {
-    case "deploy" => (pkg) => (lookup, parameters, stack) => {
+    case "deploy" => (pkg) => (logger, lookup, parameters, stack) => {
```
philwills: reporter

magenta-lib/src/main/scala/magenta/deployment_type/CloudFormation.scala
```diff
@@ -62,7 +62,7 @@ object CloudFormation extends DeploymentType {
   ).default("AMI")
 
   override def perAppActions = {
-    case "updateStack" => pkg => (lookup, parameters, stack) => {
+    case "updateStack" => pkg => (logger, lookup, parameters, stack) => {
```
philwills: etc.

magenta-lib/src/main/scala/magenta/deployment_type/DeploymentType.scala
```diff
@@ -17,27 +17,27 @@ trait DeploymentType {
   val paramsList = mutable.Map.empty[String, Param[_]]
   lazy val params = paramsList.values.toSeq
   def perAppActions: PartialFunction[String, DeploymentPackage =>
-    (Lookup, DeployParameters, Stack) => List[Task]]
+    (DeployReporter, Lookup, DeployParameters, Stack) => List[Task]]
```
philwills: What's the reason for adding this to the start, rather than the end of the list?

philwills:
There are lots of things I notice that we can improve looking through this, but this gets rid of one of my most disliked things about RiffRaff, so I'm definitely in the let's 🚢  now, tweak later crowd for this one.

👏 🎬 👏

magenta-lib/src/main/scala/magenta/deployment_type/CloudFormation.scala
sihil: I really thought I'd found all of these... apparently my late night find/replace skillz are lacking.

magenta-lib/src/main/scala/magenta/deployment_type/DeploymentType.scala
sihil: My vague ordering here is parameters that are more riff-raff related to the left and more specific deploy related to the right. Platform => Specific Deploy. I'm not wedded to it though.

sihil:
Yes - the reporter can definitely do with some simplification - perhaps by redefining events for observers.

sihil:
Also - do we want to:
 a. give the CLI some love
 b. remove the CLI entirely

rtyley:
image

PR #295, opened 2016-05-23
Make sure we only upload once

sihil:


philwills:
Feels like we could reasonably test this, but 👍

PR #296, opened 2016-05-24
Cater for case with zero length prefix

sihil:
The restructured `S3Upload` task broke mobile's static file deploy as we inject a superfluous `/` into the key when a zero length prefix is provided (see https://riffraff.gutools.co.uk/deployment/view/1cbf487f-89c7-46c9-8cbe-2b5facc27eed?verbose=1).

This fix special cases the zero length key.


alexduf:
:+1: Thanks @sihil !

PR #300, opened 2016-05-31
Add CSRF validation for all POST endpoints

sihil:


cb372:
Why not use the global CSRF filter? That way there's no worry about forgetting to add `CSRFCheck` somewhere. The filter will automatically add a CSRF to every request and do a CSRF check on every POST.

sihil:
Because there are some API endpoints that I don't want to protect. Unless
you can bypass it for a set of endpoints.

On Tue, 31 May 2016 17:08 Chris Birchall, <notifications@github.com> wrote:

> Why not use the global CSRF filter? That way there's no worry about
> forgetting to add CSRFCheck somewhere. The filter will automatically add
> a CSRF to every request and do a CSRF check on every POST.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/guardian/riff-raff/pull/300#issuecomment-222736682>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/ABLd8kKAiBEUMnieZDVqIQizibMTvwX8ks5qHFwzgaJpZM4IqoZC>
> .
>

-- 

------------------------------
This e-mail and all attachments are confidential and may also be 
privileged. If you are not the named recipient, please notify the sender 
and delete the e-mail and all attachments immediately. Do not disclose the 
contents to another person. You may not use the information for any 
purpose, or store, or copy, it in any way.  Guardian News & Media Limited 
is not liable for any computer viruses or other material transmitted with 
or as part of this e-mail. You should employ virus checking software.
 
Guardian News & Media Limited is a member of Guardian Media Group plc. Registered 
Office: PO Box 68164, Kings Place, 90 York Way, London, N1P 2AP.  Registered 
in England Number 908396

cb372:
I see. It's a shame Riff Raff is not on a newer version of Play. In recent versions `CSRFConfig` has [a `shouldProtect` predicate](https://github.com/playframework/playframework/blob/2.5.3/framework/src/play-filters-helpers/src/main/scala/play/filters/csrf/csrf.scala#L50) that you can use in cases like this.

cb372:
You could write your own filter that extends [CSRFFilter](https://github.com/playframework/playframework/blob/2.3.7/framework/src/play-filters-helpers/src/main/scala/play/filters/csrf/CSRFFilter.scala) and adds a check for "does the URL start with `/api`". But it might be more trouble than it's worth.

sihil:
Yep - as and when it is moved to 2.5 it makes sense to make the CSRF filter easier to maintain. For now though this fixes a potentially serious security problem that I'd like to resolve.

adamnfish:
I recommend viewing the diff with the "ignore whitespace" flag.

https://github.com/guardian/riff-raff/pull/300/files?w=1

jennysivapalan:
👍

PR #302, opened 2016-06-03
Enable a configurable `pathPrefixResource` for `aws-s3` upload deployments

rtyley:
The parameter is quite similar to the `bucketResource` key, but changes the path, rather than the bucket, that the resources get deployed to.

This will allow us to deploy files into differing locations in the `gdn-cdn` S3 bucket, ie:

https://interactive.guim.co.uk/2016/05/brexit-companion/
https://interactive.guim.co.uk/testing/2016/05/brexit-companion/

...the addition of `/testing/` appears to be a convention on the Visuals team, eg: https://github.com/guardian/interactive-brexit-companion/blob/530c4ad/Gruntfile.js#L125-L151

https://riffraff.gutools.co.uk/deployinfo/data

cc @sihil @DiegoVazquezNanini @annebyrne


rtyley:
I've added the config we'd want to use to s3://prism-data/prism-data.json, ie:

image

philwills:
This looks good, but it would be nice to have a test demonstrating the new behaviour.

rtyley:
> This looks good, but it would be nice to have a test demonstrating the new behaviour.

Done with 9731a25!

philwills:
👍

PR #303, opened 2016-06-06
Support looking up CloudFormation stack by tags

cb372:
Add a param to `UpdateAmiCloudFormationParameterTask` to specify tags for looking up the CF stack. `Stack` and `Stage` are automatically added to the list of tag filters if they are not already present, so usually only the `App` tag will need to be specified.

This more flexible way of looking up the CF stack is needed by CAPI because some of our stacks don't use the `<stack>-my-cf-stack-<stage>` naming scheme. In particular we have some nested CF stacks with auto-generated random IDs appended to their names.

@sihil @adamnfish @LATaylor-guardian

### DIFF
--- a/magenta-lib/src/main/scala/magenta/deployment_type/AmiCloudFormationParameter.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/AmiCloudFormationParameter.scala
@@ -1,5 +1,6 @@
 package magenta.deployment_type
 
+import magenta.tasks.UpdateCloudFormationTask.{LookupByName, LookupByTags}
 import magenta.tasks.{CheckUpdateEventsTask, UpdateAmiCloudFormationParameterTask}
 
 object AmiCloudFormationParameter extends DeploymentType {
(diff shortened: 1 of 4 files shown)


magenta-lib/src/main/scala/magenta/deployment_type/AmiCloudFormationParameter.scala
```diff
@@ -31,21 +40,35 @@ object AmiCloudFormationParameter extends DeploymentType {
...
-      val stageName = Some(parameters.stage.name).filter(_ => appendStageToCloudFormationStackName(pkg))
-      val cloudFormationStackNameParts = Seq(stackName, Some(cloudFormationStackName(pkg)), stageName).flatten
-      val fullCloudFormationStackName = cloudFormationStackNameParts.mkString("-")
+      // Based on the supplied params, decide whether to look up the CF stack by name or by tags
+      val cloudFormationStackLookupStrategy = {
+        val tagsMap = cloudFormationStackTags(pkg)
+        if (tagsMap.nonEmpty) {
+          val stackAndStage = Map("Stage" -> parameters.stage.name) ++ stack.nameOption.map("Stack" -> _)
+          val withStackAndStageTags = stackAndStage.foldLeft[Map[String, String]](tagsMap){
+            // Don't overwrite any Stack or Stage tags that the user has specified
```
philwills: Could you achieve the same thing by starting with `stackAndStage` and then just adding `tagsMap`?

philwills:
👍 👍 

Using tags makes a lot more sense to me.

magenta-lib/src/main/scala/magenta/deployment_type/AmiCloudFormationParameter.scala
cb372: You are a genius!

sihil:
I'd prefer it if this was consistent with the way the tags are looked up for ASGs (i.e. infer stack, app and stage from the `deploy.json` and deploy parameters).

I appreciate that this isn't ideal - but it works and I'd rather keep it consistent and review all of the tag behaviour later than create an inconsistent behaviour. It was in fact a conversation around adding exactly this feature that was the impetus for creating the Riff-Raff future meeting next week :)

sihil:
Superseded by #385

PR #304, opened 2016-06-16
Fix S3 deploy paths, removing double-slashes ('//')

rtyley:
With this change, when a `pathPrefixResource` is found, the `prefixStage`, `prefixPackage` and `prefixStack` keys are ignored and only the looked-up prefix is used, ie:

`/<pathPrefix>/<filePathAndName>`

The old, broken paths had `//` in them (eg `s3://gdn-cdn/testing/2016/05/brexit-companion//embed/embed.html`):

image

This follows on from #302, which introduced `pathPrefixResource` configuration.

cc @philwills @SiAdcock


philwills:
👍

rtyley:
Deploy looks good to me - @SiAdcock what do you think!?

PR #306, opened 2016-06-27
Prefix the lambda with stack name if prefixStack is set to true

jennysivapalan:
To be able work with multiple stacks we need to be able to prepend the stack name to a lambda function name. This will only be set if `prefixStack` is set to true. 

@sihil I've added a test as per our discussion.


sihil:
👍

PR #307, opened 2016-07-11
Make the project filter on deploy history page work reliably

cb372:
i.e. prevent the Hash of Doom from breaking the project name filter.

screen shot 2016-07-11 at 2 31 41 pm

I'm still not sure where that hash comes from, but this should make the URL manipulation resilient to its presence.

You're welcome, every Riff Raff user ever.


adamnfish:
OMG +10^6

LATaylor-guardian:
@cb372 Bets its on a link where the href="#" and it has an onClick!

@tomrf1 Your wish was granted.

@cb372 WHY are you working on your day off?!

philwills:
👍 👍 👍 👍 👍

sihil:
💯
It seems to appear when going through the google auth dance. Thanks for fixing :)

PR #308, opened 2016-07-11
Consistently display timestamps in London time, UK locale

cb372:
Date formatting was quite inconsistent across the app, e.g.:

* Deploy start time was in London time (actually browser locale's time, I think) on the history page, but the deploy logs were in UTC.
* API keys, etc. had their dates formatted in US style (mm/dd/yy)

There is an argument for displaying all timestamps in UTC, but given that nearly all users are in London, I think London time is more convenient. As long as we are consistent (which we now should be), it doesn't really matter which one we pick.

### DIFF
--- a/magenta-lib/src/main/scala/magenta/reporting.scala
+++ b/magenta-lib/src/main/scala/magenta/reporting.scala
@@ -1,8 +1,8 @@
 package magenta
 
-import org.joda.time.DateTime
+import org.joda.time.{DateTime, DateTimeZone}
 import org.joda.time.format.DateTimeFormat
-import java.util.UUID
+import java.util.{UUID, Locale}
 
 object RunState extends Enumeration {
   type State = Value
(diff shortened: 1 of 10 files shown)


adamnfish:
Frontend has a nice approach to this. You could print all dates/times in a span displayed as UTC and include the raw timestamp as a data attribute. Then add a piece of JavaScript that runs through all the dates on the page and converts them to local time.

Maybe not worth it here though, up to you.

cb372:
I think I'll just merge it like this (assuming it looks ok to @sihil and @philwills), as I don't really want to spend any more time on this. I was merely scratching an itch.

philwills:
Yep, local would be lovely, but consistency is a more pressing issue. 👍

cb372:
Deployed the branch and it seems to do what I expect. Merging.

PR #309, opened 2016-07-15
Remove CLI tool

sihil:
The CLI tool is never used any more and is very unlikely to actually work. There are other approaches to deploying code when Riff-Raff is unavailable.


philwills:
👍 👍

PR #310, opened 2016-07-15
Remove legacy deployment types

sihil:
I'm confident that we no longer use any of these deployment types: ExecutableJarWebapp, JettyWebapp, FileCopy, Django, RPM, NativePackagerWebapp

As a result the code that supports them can be deleted along with the underlying tasks. We can also drop support for RemoteShell commands (SSH commands) and with that we can simplify the KeyRing to no longer hold an SSH key.


philwills:
🔥 👍 🔥

PR #311, opened 2016-07-15
Eliminate the artifact.zip deployable format

sihil:
The `artifact.zip` file was originally intended as a way of packaging all of the files for a deployment into a single moveable file. However this is no longer desirable. Typically a build spends ~ 30s - 1m building `artifacts.zip` files from a `deploy.json` and then uploading them to S3. The first thing Riff-Raff does is download the file and unzip it on local disk - which also takes time.

Instead we can upload what would previously have been the contents of the tarball directly into the S3 location it would have occupied. This will mean that we can:
 - reduce the pipeline latency (no more zipping and unzipping)
 - further reduce local state on the Riff-Raff server (tasks will operate directly on the files in S3)
 - increase the speed of previews (they will no longer need to download the whole tarball just to get to `deploy.json`)

This PR sets out to do just that by modelling an `S3Artifact` that knows where the `deploy.json` can be found and can also provide `S3Package` instances for each of the package locations in S3. These are modelled on top of `S3Location`, `S3Path` and `S3Object` that provide simple and generic S3 concepts. These replace all uses of files throughout the codebase.

In the tasks where we uploaded files, we now open `InputStream`s from the source S3 objects and upload them directly into S3 or Fastly etc. Due to the need to work with two sets of AWS credentials (one for access to the artifact and another set for access to a project's AWS account) we cannot do copies directly within AWS but transfer data via the Riff-Raff instance.

In order to make the transition easy we will continue to support the legacy `artifact.zip` format. Rather than support both throughout Riff-Raff - we have a simple conversion that is applied when a `deploy.json` file cannot be found but an `artifact.zip` file can be. Riff-Raff will download the legacy tarball, explode it locally and then re-upload the contents into the same location. Once done, the deploy will kick off. This will slow Riff-Raff down until the `sbt-riffraff-artifact` plugin supports the new format. In addition, we do not (for now) delete the legacy tarball to allow for Riff-Raff to be rolled back.

Left to do:
 - [x] At least some testing


sihil:
I've flushed out obvious bugs for this and am confident that the core changes and the S3Upload task work as expected. I've not exhaustively tried the other tasks that upload files using other mechanisms (e.g. Fastly, CloudFormation and Lambda deployment types).

Definitely worth reviewing at this stage.

magenta-lib/src/main/scala/magenta/artifact/ArtifactHelper.scala
```diff
@@ -0,0 +1,30 @@
+package magenta.artifact
+
+import com.amazonaws.services.s3.AmazonS3
+import com.amazonaws.services.s3.model.{ListObjectsV2Request, S3ObjectSummary}
+
+import scala.collection.JavaConverters._
+
+object ArtifactHelper {
+  def listObjects(path: S3Location)(implicit s3Client: AmazonS3) = {
```
philwills: This looks quite a lot like https://github.com/guardian/riff-raff/blob/master/riff-raff/app/ci/S3Location.scala#L25

Is it worth pulling them out. Also I'm not a fan of names of the form `XyHelper`

sihil: Ha - totally hadn't noticed that. Mostly only touched magenta.

Would you be happy for that to be extracted into the magenta library if I can make it work?

magenta-lib/src/main/scala/magenta/artifact/S3Location.scala
```diff
@@ -0,0 +1,89 @@
...
+import com.gu.management.Loggable
+import magenta.{Build, DeployReporter}
+
+trait S3Location {
+  def bucket: String
+  def prefix: String
+  def prefixElements: List[String] = prefix.split("/").toList
+  def fileName:String = prefixElements.last
+  def extension:Option[String] = if (fileName.contains(".")) Some(fileName.split('.').last) else None
+  def string()(implicit client:AmazonS3):Option[String] = {
```
cb372: Could rename this method to give more of a clue that it actually downloads data from S3.

Actually it doesn't seem to be used anywhere, so could be deleted?

sihil: I've done that (it was used in one place).

magenta-lib/src/main/scala/magenta/json/JsonInputFile.scala
```diff
@@ -38,24 +43,26 @@ case class JsonRecipe(
 object JsonReader {
   private implicit val formats = DefaultFormats
 
-  def parse(f: File)(implicit reporter: DeployReporter): Project = {
+  def parse(artifact: S3Artifact)(implicit s3Client: AmazonS3Client, reporter: DeployReporter): Project = {
     try {
-      parse(Source.fromFile(f).mkString, f.getAbsoluteFile.getParentFile)
+      val deployJson = s3Client.getObject(artifact.bucket, s"${artifact.key}/${artifact.deployObject}")
+      val deployJsonInputStream = deployJson.getObjectContent
+      parse(Source.fromInputStream(deployJsonInputStream).mkString, artifact)
```
cb372: Fun (but little-known) fact: `scala.io.Source` does not handle closing the resource for you, so this is a leak.

magenta-lib/src/main/scala/magenta/tasks/FastlyTasks.scala
```diff
@@ -70,14 +72,15 @@ case class UpdateFastlyConfig(pkg: DeploymentPackage)(implicit val keyRing: KeyR
...
-        if (file.getName.endsWith(".vcl")) {
-          reporter.info(s"Uploading ${file.getName}")
-          val vcl = scala.io.Source.fromFile(file.getAbsolutePath).mkString
-          block(client.vclUpload(versionNumber, vcl, file.getName, file.getName))
+      ArtifactHelper.listObjects(s3Package)(artifactClient).map { obj =>
+        if (obj.extension.contains("vcl")) {
+          val fileName = obj.relativeTo(s3Package)
+          val stream = artifactClient.getObject(obj.bucket, obj.key).getObjectContent
+          reporter.info(s"Uploading $fileName")
+          val vcl = scala.io.Source.fromInputStream(stream).mkString
```
cb372: another leak here

magenta-lib/src/main/scala/magenta/artifact/ArtifactHelper.scala
philwills: Yep, absolutely.

sihil: Done - the Riff-Raff play app now shares the common S3 wrappers in `magenta-lib`

magenta-lib/src/main/scala/magenta/artifact/S3Location.scala
```diff
@@ -0,0 +1,131 @@
...
+
+  def buildPrefix(build: Build): String = {
+    s"${build.projectName}/${build.id}"
+  }
+
+  /*
+  This is essentially the same as the apply method in function, but in this case we actually confirm that the
+  deployObject exists. If it doesn't then we try to convert a legacy artifacts.zip file into the new format.
+   */
+  def withLegacyFallback(build: Build, bucket: String)(implicit client: AmazonS3, reporter: DeployReporter): S3Artifact = {
```
philwills: I'd expect `withLegacyFallback` to take a function or a `Try` and then recover from the failure case, rather than doing both the happy and sad cases itself.

Also, I'd probably err away from calling it `Legacy`. Maybe something like `ZipBundle`?

magenta-lib/src/main/scala/magenta/json/JsonInputFile.scala
sihil: Thanks Chris - good spot. Actually found a couple of other places this was done. Have fixed them all.

magenta-lib/src/main/scala/magenta/artifact/S3Location.scala
sihil: Any better?

philwills:
👍

philwills:
Let's give it a go.

PR #313, opened 2016-07-22
Refactor deployment type by introducing resources and target classes

sihil:
This will make it less intrusive to make other more significant changes (adding more resources to this case class is easier than changing every deployment type).

### DIFF
--- a/magenta-lib/src/main/scala/magenta/Project.scala
+++ b/magenta-lib/src/main/scala/magenta/Project.scala
@@ -1,10 +1,11 @@
 package magenta
 
-import json.{DeployInfoHost, DeployInfoJsonInputFile}
-import tasks.Task
-import collection.SortedSet
 import java.util.UUID
+
+import magenta.json.{DeployInfoHost, DeployInfoJsonInputFile}
+import magenta.tasks.Task
 import org.joda.time.DateTime
+
 import scala.math.Ordering.OptionOrdering
 
 object DeployInfo {
@@ -130,6 +131,13 @@ object HostList {
   implicit def hostListAsListOfHosts(hostList: HostList) = hostList.hosts
 }
 
+case class DeploymentResources(reporter: DeployReporter, lookup: Lookup) {
+  def assembleKeyring(target: DeployTarget, pkg: DeploymentPackage): KeyRing =
+    lookup.keyRing(target.parameters.stage, pkg.apps.toSet, target.stack)
+}
+
+case class DeployTarget(parameters: DeployParameters, stack: Stack)
+
 /*
  An action represents a step within a recipe. It isn't executable
  until it's resolved against a particular host.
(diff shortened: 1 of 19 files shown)


magenta-lib/src/main/scala/magenta/Project.scala
```diff
@@ -130,14 +131,18 @@ object HostList {
   implicit def hostListAsListOfHosts(hostList: HostList) = hostList.hosts
 }
 
+case class RuntimeResources(reporter: DeployReporter, lookup: Lookup, parameters: DeployParameters, stack: Stack) {
```
philwills: I'm not keen on the name and feel that partly that's because there are two different types of thing here. 

To me, the reporter and the lookup are things a task might need to perform an action, whereas the params and stack determine what the deployment tries to do.

magenta-lib/src/main/scala/magenta/deployment_type/S3.scala
```diff
@@ -109,32 +109,32 @@ object S3 extends DeploymentType with S3AclParams {
...
-    case "uploadStaticFiles" => (pkg) => (reporter, resourceLookup, parameters, stack) => {
+    case "uploadStaticFiles" => (pkg) => (runtime) => {
       def resourceLookupFor(resource: Param[String]): Option[Datum] = {
         resource.get(pkg).flatMap { resourceName =>
           assert(pkg.apps.size == 1, s"The $name package type, in conjunction with ${resource.name}, only be used when exactly one app is specified - you have [${pkg.apps.map(_.name).mkString(",")}]")
-          val dataLookup = resourceLookup.data
+          val dataLookup = runtime.lookup.data
           val app = pkg.apps.head
-          val datumOpt = dataLookup.datum(resourceName, app, parameters.stage, stack)
+          val datumOpt = dataLookup.datum(resourceName, app, runtime.parameters.stage, runtime.stack)
```
philwills: To attempt to bolster my earlier point, there are lots of places like this where DeployParameters and Stack appear together without the Lookup and Reporter.

philwills:
I think this is a good idea, but as mentioned I don't think they all belong together in one blob.

magenta-lib/src/main/scala/magenta/Project.scala
sihil: Yep - that did occur to me there were two things. You'd advocate for two case classes? `Resources` and `???`.

philwills: The best I've been able to think of is `ResolvedDeployParameters`, or `PerStackDeployParameters`. I'd probably err towards  the first.

sihil:
I've gone for `RuntimeResources` and `DeployTarget` although happy to rename the case class again if you don't think that makes sense. Let me know what you think (before I rebase it onto the next PR).

philwills:
I don't like `Runtime`. Maybe `DeploymentResources`?

philwills:
`DeployTarget` sounds good to me.

sihil:
Voila!

philwills:
👍

PR #314, opened 2016-07-27
Avoid buffering in memory when downloading and uploading files

sihil:
Magenta-lib was buffering the whole file in memory when downloading artifacts and uploading to S3. With the recent changes memory pressure will have increased somewhat - this removes the needless use of memory when handling streams.


johnduffell:
good idea 👍

PR #315, opened 2016-07-27
Use the AWS mimetype map to replicate prior behaviour

sihil:
When we uploaded from files, the AWS code would lookup in an internal map from the extension of the file being uploaded. When uploading using an input stream it defaults everything that isn't explicitly set to application/octet-stream. This change inspects the target key and sets the content type from the lookup file.


satterly:
👍

PR #316, opened 2016-07-27
Don't log more than 10 files in detail

sihil:
When we are uploading thousands of files this behaviour puts a strain on the DB. A handful of files is useful for debugging, but logging everything is superfluous so lets just do the first 10.


satterly:
👍

PR #317, opened 2016-08-10
Add parameter to set custom headers on S3 upload

piuccio:
Similarly to cacheControl, allow setting any custom header.

My use case it to set an [S3 redirect](http://docs.aws.amazon.com/AmazonS3/latest/dev/how-to-page-redirect.html) for discussion-frontend as @jamespamplin suggested.

@sihil @philwills forgive my scala skills, that's the best I could come up with.

I have absolutely no idea on how to run riff raff locally (and I'd like to avoid doing that), but at least it compiles, tests pass and I've added a test for that use case.

### DIFF
--- a/magenta-lib/src/main/scala/magenta/deployment_type/S3.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/S3.scala
@@ -96,6 +96,32 @@ object S3 extends DeploymentType with S3AclParams {
     """.stripMargin
   ).default(Map.empty)
 
+  val headers = Param[List[PatternList]]("headers",
+    """
+      |Set custom headers for the uploaded files.
+      |headers parameter is set to an array of JSON objects, each of which have `pattern` and `list` keys.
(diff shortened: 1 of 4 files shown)


sihil:
I'm not against this, but isn't your intention to use this for a Location redirect header? That implies that you will dynamically write a `deploy.json` file to contain the hashed target. Or am I missing your intention here?

If my assumptions are correct then I'd advocate a different approach. Maybe something similar to the Apache `asis` mode? Or an accompanying `.metadata.json` file? Dynamically changing the `deploy.json` for each artifact feels like an anti-pattern to me. This file is meant to describe a deploy process and IMHO should be static.

Happy to chat further about what you're trying to do.

jamespamplin:
@sihil raises some good points. Keeping `deploy.json` static makes sense. An accompanying metadata file could be a good alternative. Although I'd suggest perhaps `deploy.metadata.json` as a filename to associate it to the riff-raff `deploy.json`.

piuccio:
Thanks for the comments, I'm closing this and implement something that doesn't require riffraff changes

PR #318, opened 2016-08-12
Add a way of instantly seeing the associated deploy.json

sihil:
It's pretty difficult to view the deploy.json without accessing the S3 bucket. I'd like to expand this to a debug view for a build, but this is an easy starting point that only took a few minutes but will none the less be useful.

screen shot 2016-08-12 at 17 29 37


philwills:
Can confirm this is already useful.

PR #319, opened 2016-08-16
adds a more understandable message to the cloudformation deployment type

alexduf:
I tried to add a unit test but it was getting endless as I had to mock the s3 client.
cc @sihil


sihil:
👍

PR #320, opened 2016-08-19
Switch to a graph of deployments instead of simple task list

sihil:
This is an initial implementation of parallel deployment. The resolver now returns a directed acyclic graph (DAG) of lists of tasks instead of a simple list of tasks. This DAG contains a `StartNode` and an `EndNode` so that the graph can be walked. The nodes in the graph are `DeploymentNode`s which represent a list of tasks that should be executed in series along with a label and priority.

Previously the FSM in the `DeployCoordinator` used to step through the list of tasks. When it was notified that a task was complete it would kick off the next task and so on until it completed. The DAG version starts by finding the `StartMarker` and then starts all direct successors. Upon receiving a notification that a task has completed it will kick off any direct successors that have no uncompleted predecessors. It does this until no tasks are left to execute.

There has been a good deal of changes under the hood. The `TaskRunner` class has now been changed to `DeploymentRunner`. The latter is now responsible for executing a list of tasks rather than a single task. The `DeployRunState` class has now been turned into an Actor (`DeployGroupRunner`) that sits between the `DeployCoordinator` and the new `DeploymentRunner`s. Initial attempts did not do this and they resulted in code that was hard to write and hard to reason about.

In this initial implementation we only parallelise the deployment of apps with multiple stacks. If there are multiple stacks then it will be resolved into a DAG with parallelised elements, in all other cases a DAG is still created but it is equivalent to a list.

TODO:
 - [x] Add tests around the FSM?
 - [x] Add tests around the `TaskGraph` object
 - [x] Eliminate the non-deterministic nature of `TaskGraph.toTaskGraph`
 - [x] Consider / test unhappy paths
 - [x] Add tests of the resolver graph
 - [x] Consider converting `DeployRunState` to an actor
 - [x] Review the entire implementation to see if it can be done more cleanly
 - [x] Review the number of available threads
 - [x] Switch from scala-graph to our own simple implementation
 - [x] Review whether the `DeploymentGraph` should focus on nodes instead of edges
 - [x] Review the supervisor strategies and logging of actor failures


philwills:
💣 💥

riff-raff/app/deployment/actors.scala
```diff
@@ -115,209 +114,302 @@ class DeployMetricsActor extends Actor with Logging {
...
         }
-        DeployReporter.withFailureHandling(deployReporter) { safeReporter =>
-          safeReporter.taskContext(task.task) { taskReporter =>
-            task.task.execute(taskReporter, stopFlagAsker)
+        deployReporter.taskContext(task.task) { taskReporter =>
+          //task.task.execute(taskReporter, stopFlagAsker)
+          // TODO: remove fake work here
+          task.task match {
+            case DoubleSize(_, _, stack) if stack.nameOption == Some("flexible") =>
+              //taskReporter.fail("I'm failing here to see what happens...")
```
philwills: I'm assuming this commented out stuff should go.

philwills:
I'm definitely going to need you to talk me through the graph stuff in this, as there's a lot I'm unfamiliar with. I'm also generally nervous of anything with a `scalax` package, as they seem to have often been rather abandoned.

riff-raff/app/deployment/actors.scala
sihil: Uh, yeah. That's the third time I've checked it in 😖

sihil:
The graph library seems to continue to have releases, albeit infrequently. If there is another graph library then I'm happy to switch to that. Or even implement a simple library - it turns out I'm not using a huge amount of the library itself (e.g. I feel like I should be able to use the traversal stuff rather than roll my own, but it's not quite as useful as you'd hope).

cb372:
I am also a bit wary about that graph library. It looks like it is still being actively developed (last commit was two weeks ago: https://github.com/scala-graph/scala-graph) but I get a strange feeling about it.

magenta-lib/src/main/scala/magenta/tasks/package.scala
```diff
@@ -0,0 +1,94 @@
...
+  /** Default (immutable) directed acyclic `Graph`. */
+  type TaskGraph = Graph[TaskNode, LDiEdge]
+  /** Companion module for default (immutable) directed acyclic `Graph`. */
+  object TaskGraph extends CompanionAlias[LDiEdge](dagConstraint) {
+    private def toDiEdges[T](nodes: List[T]): Set[LDiEdge[T]] = {
+      nodes match {
+        case Nil => Set.empty
+        case single :: Nil => Set.empty
+        case first :: second :: tail =>
+          toDiEdges(second :: tail) + (first ~+> second)(None)
```
cb372: wtf does all this mean?

philwills: That was my initial reaction too.

magenta-lib/src/main/scala/magenta/graph/DeploymentGraph.scala
```diff
@@ -0,0 +1,90 @@
...
+    val deploymentNode = DeploymentNode(tasks, pathName, priority)
+    DeploymentGraph(StartNode() ~> deploymentNode, deploymentNode ~> EndNode())
+  }
+
+  def apply(edges: Edge*): DeploymentGraph = DeploymentGraph(edges.toSet)
+
+  val empty = DeploymentGraph(StartNode() ~> EndNode())
+}
+
+case class DeploymentGraph(edges: Set[Edge]) {
```
philwills: I can't help feeling that a lot of this would be simpler if our view was oriented around nodes, rather than edges.

magenta-lib/src/main/scala/magenta/graph/DeploymentGraph.scala
```diff
@@ -0,0 +1,90 @@
+package magenta.graph
+
+import magenta.tasks.Task
+
+sealed trait Node {
+  def ~>(to: Node) = Edge(this, to)
+  def maybePriority: Option[Int] = None
+}
+case class StartNode() extends Node
```
philwills: Any reason these aren't case objects?

philwills:
This definitely feels like a significant improvement in understandability.

magenta-lib/src/main/scala/magenta/graph/DeploymentGraph.scala
sihil: I had some issues with pattern matching on case objects, but turns out those issues don't seem to affect this particular case.

sihil:
OK - think this is properly ready to go. Planning to push it out first thing tomorrow morning and see what happens.

riff-raff/app/deployment/actors.scala
```diff
@@ -115,209 +110,310 @@ class DeployMetricsActor extends Actor with Logging {
...
+    case DeploymentFailed(deployment, exception) =>
+      log.debug("Deployment failed")
+      markFailed(deployment)
+      if (isExecuting) {
+        log.debug("Failed during deployment but others still running - deferring clean up")
+      } else {
+        cleanup
+      }
+
+    case Terminated(actor) =>
```
cb372: I'm pretty sure you will receive this message in both the "child actor stopped cleanly by calling context.stop" and the "child actor exploded in a fireball" cases. So you might want to add a guard to only call `failRootContext` if `rootContextClosed == false`.

riff-raff/app/deployment/actors.scala
```diff
@@ -115,209 +110,310 @@ class DeployMetricsActor extends Actor with Logging {
...
+          cleanup
+        }
+
+      case None =>
+        elseBlock
+    }
+  }
+
+  @scala.throws[Exception](classOf[Exception])
+  override def postStop(): Unit = {
```
cb372: This will also be called in both the happy and sad cases

riff-raff/app/deployment/actors.scala
```diff
@@ -115,209 +110,310 @@ class DeployMetricsActor extends Actor with Logging {
...
+      context
+    }
+  }
+
+  private def runDeployments(deployments: List[DeploymentNode]) = {
+    log.debug(s"Running next deployments: $deployments")
+    try {
+      honourStopFlag(rootReporter) {
+        deployments.foreach { deployment =>
+          val deploymentRunner = context.watch(deploymentRunnerFactory(context, s"${record.uuid}-${deployment.pathName}-${deployment.priority}"))
```
cb372: I don't think you need to call `context.watch` because you are the parent actor. You should automatically receive a `Terminate` message when the child dies.

riff-raff/app/deployment/actors.scala
sihil: Yep - that's fine. I record if a context has already been failed and check in the method. I'll rename the method or make the check explicit.

riff-raff/app/deployment/actors.scala
sihil: I've now changed the child to not stop itself (as I was getting dead letters). They will be terminated when the parent is stopped. The check you suggest is already done in the method, but I'll revisit it to improve clarity.

sihil:
I ran this for a short while this morning (from 11:45 until 12:08) 😞 

 - Somehow the self deploy to rollback didn't work - possibly it thought it was still deploying when it wasn't
 - The dual stack deploys I tried all failed spectacularly due to hitting AWS throttling limits (for an example see https://riffraff.gutools.co.uk/deployment/view/1fd912bd-ff46-4a75-83c0-a143f8afd2ec)

I've rolled back (with a manual restart) and now want to take stock.

PR #322, opened 2016-08-24
Update riffraff.template

fredex42:
@akash1810  @gidsg  - Requesting access to RiffRaff for the Multimedia AWS account.  Main reason is to be able to deploy the in-development live stream tool and content delivery for atoms.


akash1810:
🙌 👍 assuming this is the only place the account number is needed (ping @sihil)

sihil:
👍 You'll need to create some credentials in your AWS account too.

fredex42:
yep got them built :-)

PR #326, opened 2016-09-05
Big sbt refactor

cb372:
More preparation for the Play upgrade.

* `project/MagentaBuild.scala` -> `build.sbt`
* Get rid of the per-project build.sbt files
* Get rid of a few unneeded deps and resolvers
* General cleanup/simplification


build.sbt
```diff
@@ -1 +1,77 @@
...
+    buildInfoOptions += BuildInfoOption.BuildTime,
+    buildInfoPackage := "riffraff",
+
+    resolvers += "Brian Clapper Bintray" at "http://dl.bintray.com/bmc/maven",
+    libraryDependencies ++= riffRaffDeps,
+
+    packageName in Universal := normalizedName.value,
+    topLevelDirectory in Universal := Some(normalizedName.value),
+    riffRaffPackageType := (packageZipTarball in Universal).value,
+    riffRaffBuildIdentifier := sys.env.getOrElse("TRAVIS_BUILD_NUMBER", "DEV"), // TODO Travis??
```
cb372: @philwills I'm confused. The plugin is obviously picking up the build number from TeamCity successfully, but I have no idea how.

build.sbt
```diff
@@ -1 +1,77 @@
...
+    ivyXML := {
+      <dependencies>
+        <exclude org="commons-logging"><!-- Conflicts with acl-over-slf4j in Play. --> </exclude>
+        <exclude org="oauth.signpost"><!-- Conflicts with play-googleauth--></exclude>
+        <exclude org="org.springframework"><!-- Because I don't like it. --></exclude>
+        <exclude org="xpp3"></exclude>
+      </dependencies>
+    },
+
+    unmanagedClasspath in Test += Attributed.blank(baseDirectory.value / "test"), // TODO what's this for?
```
cb372: @philwills any idea?

build.sbt
```diff
@@ -1 +1,77 @@
...
+    fork in Test := false,
+
+    includeFilter in (Assets, LessKeys.less) := "*.less",
+
+    magenta := riffRaffArtifact.value
+  ))
+
+// TODO what's this about?
+lazy val magenta = taskKey[File]("Alias to riffRaffArtifact for TeamCity compatibility")
```
cb372: @philwills also this

philwills: That was just something I put in so that I didn't have to sync up changing the build in Teamcity with merging a PR. No reason not to change the build to use `riffRaffUpload` and then get rid of this.

build.sbt
philwills: Ancient uber-jar munging. I can't remember if we even deploy as an uberjar any more. I'd certainly not recommend it.

build.sbt
philwills: I'm guessing we're not actually using these settings at all and that this is only working due to the teamcity-s3-plugin and it's magic.

cb372:
@philwills Thanks your for comments. I'll try switching TeamCity to `riffRaffUpload`.

Current status: failing on TeamCity, but the TC logs give no clue why. *sigh*.

philwills:
I blame the person who wrote the SBT plugin.

cb372:
I've deployed this branch to PROD and it seems to work fine. Also checked that you can build with a cold Ivy cache. Merging.

PR #327, opened 2016-09-05
Make graph implementation generic and add new methods

sihil:
In preparation for overhauling the resolver I've made the new graph type I built generic and made it somewhat more fully featured. 

Making it generic makes it possible to map from a graph of one type to a graph of a different type - for example this makes it possible to resolve a graph of recipes as the first step and then map the recipes into deployments as a second step rather than doing everything at once.

I've also moved priority onto the edge instead of the node. Turns out that it makes merging graphs like the following much easier as nodes are easily discoverable within the edges and nodes without having to workaround priorities. Other methods are easier to implement too.

This was originally done to deal with cases like the below. Whilst this is contrived it does deal with cases where a deployment is duplicated and needs to be de-duplicated. Previously the join methods would effectively duplicate the nodes (with different priority).

```
S     S     +-+ S +-+
            |       |
+     +     |       |
|     |     v       |
|     |             |
v     v     1       |
                    |
1     2     +       |
            |       v
+     +     |
|  +  |  =  +-----> 2 +-----+
|     |                     |
v     v             +       |
                    |       v
2     3             |
                    |       3
+     +             |
|     |             |       +
|     |             |       |
v     v             |       |
                    +-> E <-+
E     E
```

### DIFF
--- a/magenta-lib/src/main/scala/magenta/graph/Graph.scala
+++ b/magenta-lib/src/main/scala/magenta/graph/Graph.scala
@@ -2,18 +2,18 @@ package magenta.graph
 
 sealed trait Node[+T] {
   def ~>[R >: T](to: Node[R]): Edge[R] = Edge(this, to)
-  def maybePriority: Option[Int] = None
   def maybeValue: Option[T] = None
 }
 case object StartNode extends Node[Nothing]
-case class MidNode[T](value: T, priority: Int = 1) extends Node[T] {
-  override def maybePriority: Option[Int] = Some(priority)
-  def incPriority(n: Int): MidNode[T] = this.copy(priority = priority + n)
+case class MidNode[T](value: T) extends Node[T] {
   override def maybeValue: Option[T] = Some(value)
 }
 case object EndNode extends Node[Nothing]
 
-case class Edge[+T](from: Node[T], to: Node[T])
+case class Edge[+T](from: Node[T], to: Node[T], priority: Int = 1) {
+  def incPriority(n: Int) = this.copy(priority = priority + n)
+  def ~=[R >: T](other: Edge[R]) = from == other.from && to == other.to
+}
 
 object Graph {
   def apply[T](edges: Edge[T]*): Graph[T] = Graph(edges.toSet)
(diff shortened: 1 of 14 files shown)


cb372:
Sorry, failing build is my fault. My sbt changes have caused the subproject name to change from `riff-raff` to `riffraff`, so I updated the TeamCity build to reflect that. I was hoping nobody would do any work on Riff Raff until I merged my branch!

cb372:
I've now merged my branch. Please rebase against master and your build should pass.

magenta-lib/src/main/scala/magenta/graph/SimpleGraph.scala
```diff
@@ -0,0 +1,101 @@
...
+
+case class Edge[+T](from: Node[T], to: Node[T])
+
+object SimpleGraph {
+  def apply[T](edges: Edge[T]*): SimpleGraph[T] = SimpleGraph(edges.toSet)
+
+  def empty[T] = SimpleGraph[T](StartNode ~> EndNode)
+}
+
+case class SimpleGraph[T](edges: Set[Edge[T]]) {
```
cb372: My prediction: one day this class will grow into a 1000-line monster and you will rue the day you called it `SimpleGraph`

sihil: I've renamed it :)

magenta-lib/src/main/scala/magenta/graph/package.scala
```diff
@@ -1,15 +1,15 @@
 package magenta
 
 package object graph {
-  implicit class RichNodeSet(nodes: Set[Node]) {
-    def filterDeploymentNodes: Set[DeploymentNode] = nodes.flatMap{
-      case deploymentNode:DeploymentNode => Some(deploymentNode)
+  implicit class RichNodeSet[T](nodes: Set[Node[T]]) {
+    def filterMidNodes: Set[MidNode[T]] = nodes.flatMap{
```
cb372: Could be slightly simplified to `nodes.collect { case mn: MidNode[T] => mn }`

magenta-lib/src/main/scala/magenta/graph/Graph.scala
```diff
@@ -0,0 +1,101 @@
...
+      ((if (!nodes.contains(StartNode)) Some("No start node") else None) ::
+        (if (!nodes.contains(EndNode)) Some("No end node") else None) :: Nil).flatten
+    if (errors.isEmpty) Right(true) else Left(errors)
+  }
+  assert(isValid.isRight, s"Graph isn't valid: ${isValid.left.get.mkString(", ")}")
+
+  def get(node: T): MidNode[T] = dataNodes.find(_.value == node).get
+  def successors(node: Node[T]): Set[Node[T]] = edges.filter(_.from == node).map(_.to)
+  def predecessors(node: Node[T]): Set[Node[T]] = edges.filter(_.to == node).map(_.from)
+  def successorNodes(node: Node[T]): List[MidNode[T]] = {
```
cb372: I don't quite understand the naming or the use case of this method

magenta-lib/src/main/scala/magenta/graph/Graph.scala
```diff
@@ -0,0 +1,101 @@
...
+    successors(node).filterMidNodes.toList.sortBy(_.priority)
+  }
+
+  def replace(node: Node[T], withNode: Node[T]): Graph[T] = {
+    assert(nodes.contains(node), "Node to replace not found in graph")
```
cb372: Throwing an exception here seems a bit brutal. Is this check necessary?

If the node is not present, you could just return the original graph unchanged. And maybe return a `(Graph[T], Boolean)` tuple to signal whether a replacement took place, if that information is useful.

cb372:
Looks fine to me apart from a few stylistic comments

magenta-lib/src/main/scala/magenta/graph/package.scala
sihil: Oh man - that's a great function. Thanks for the introduction to the world of `collect`.

magenta-lib/src/main/scala/magenta/graph/Graph.scala
sihil: That's a fair point. I'll remove the assertion - it will work anyway and wind up returning the original graph as you say.

magenta-lib/src/main/scala/magenta/graph/Graph.scala
sihil: The idea is that it returns the successors in order of node priority.

In my recent restructure (to move priority back onto the edges - which makes it a lot easier to merge graphs as the nodes are comparable) I've also taken the opportunity to rename the method to orderedSuccessors.

sihil:
I've done a whole bunch more work on this to move priority back onto edges so might be worth another look. I think this is cleaner and a saner way to think about priorities - certainly made it easier to fix the join methods to behave correctly for more complex cases.

PR #328, opened 2016-09-06
Make the continous deployments Dynamo table name configurable

cb372:
Corresponding CloudFormation change: https://github.com/guardian/deploy-tools-platform/pull/38

### DIFF
--- a/riff-raff/app/conf/context.scala
+++ b/riff-raff/app/conf/context.scala
@@ -65,7 +65,7 @@ class Configuration(val application: String, val webappConfDirectory: String = "
       credentialsProviderChain(None, None),
       new ClientConfiguration()
     )
-    val dynamoTableName = configuration.getStringProperty("continuousDeployment.dynamoTable").getOrException("Dynamo table name for continous deployment configs not configured")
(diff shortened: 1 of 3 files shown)


sihil:
For mongo there is a 'collection prefix' that was used for allowing PROD/CODE to share the same database. Should we re-use that sort of pattern here?

cb372:
I think if we add any more Dynamo tables, we should just follow the simple convention that the stage name is always appended to the table name (e.g. `my-table-PROD` and `my-table-CODE`), so we won't need to list each table name individually in the config file.

Unfortunately this table was created without following that convention so we need special treatment here.

philwills:
I'm not completely convinced there's a great advantage to having separate tables for different environments, but if you do go down that route, it's easy enough to migrated data of this size from one table to another and I'd prefer that to having a one off exception.

cb372:
Closing. As discussed offline, I'll make a new PR that makes Riff Raff use the convention of appending the stage name to all Dynamo tables. 

I'll also do the CloudFormation dance to create a new `continuous-deployment-config` table that follows the convention and migrate the data to it.

PR #329, opened 2016-09-06
Modify the resolver to create graphs more intelligently

sihil:
Start to more deeply leverage the fact that we can now run things in parallel. This now builds graphs of deployments based on the recipe dependencies in `deploy.json`. I've not (yet?) gone so far as to split out host groups - the need to do so has mostly gone (we rarely target hosts now). However if you have a static files recipe and also a number of applications that all depend on the static files then it will now do the expected thing of creating a graph of start ~> static and then static ~> app1, static ~> app2 etc. Once the static deployment is completed all of the appX deployments will kick off.

To do this I've:
 - added `toGraph` to the `RecipeTasksNode` to build a graph from tree of nodes
 - split the resolver methods down into smaller pieces
 - made various improvements to the graph library to behave correctly in a number of corner cases when merging or flattening graphs
 - added methods for gaining access to the graph edges
 - changed `map` to work with underlying types instead of the wrappers
 - added an easy to read toString method on graph class, which is helpful for troubleshooting and logging
 - added numerous tests to cover more corner cases

Todo:
 - [ ] Testing locally or in CODE (there are not so many config files that will make use of this yet, so might need to create one)

### DIFF
--- a/magenta-lib/src/main/scala/magenta/graph/Graph.scala
+++ b/magenta-lib/src/main/scala/magenta/graph/Graph.scala
@@ -23,6 +23,7 @@ object Graph {
 
 case class Graph[T](edges: Set[Edge[T]]) {
   val nodes = edges.map(_.from) ++ edges.map(_.to)
+  val underlyingNodes = nodes.flatMap(_.maybeValue)
   val dataNodes = nodes.filterMidNodes
 
   val isValid: Either[List[String], Boolean] = {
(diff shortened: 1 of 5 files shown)


magenta-lib/src/main/scala/magenta/graph/Graph.scala
```diff
@@ -74,29 +80,49 @@ case class Graph[T](edges: Set[Edge[T]]) {
...
     Graph(mergedEdges)
   }
 
-  def toList: List[T] = {
-    def traverseFrom(node: Node[T], visited: Set[Node[T]]): List[Node[T]] = {
-      val predecessors: Set[Node[T]] = this.predecessors(node)
-      if ((predecessors -- visited).nonEmpty) {
-        // if there are some predecessors of this node that we haven't yet visited then return empty list - we'll be back
-        Nil
+  lazy val nodeList: List[Node[T]] = {
```
sihil: This is now a more typical topological sort of the graph - I switched this from recording visited nodes to recording traversed edges which is more important in many shapes of graphs.

magenta-lib/src/main/scala/magenta/graph/Graph.scala
```diff
@@ -74,29 +80,49 @@ case class Graph[T](edges: Set[Edge[T]]) {
   }
 
   def joinSeries(other: Graph[T]): Graph[T] = {
```
sihil: This was pretty broken in the way it modified (or didn't) priorities when gluing graphs together end to end. It feels like there should be a simpler way to express this, but it is at least passing a suite of tests.

PR #330, opened 2016-09-07
Use the "table-name-$STAGE" convention for Dynamo table names

cb372:
Look up the stage using the instance's tags.

@sihil @philwills

### DIFF
--- a/riff-raff/app/conf/context.scala
+++ b/riff-raff/app/conf/context.scala
@@ -9,8 +9,6 @@ import com.gu.management._
 import logback.LogbackLevelPage
 import com.gu.management.play.{Management => PlayManagement}
 import com.gu.conf.ConfigurationFactory
-import java.io.File
-
 import magenta._
 import controllers.{Logging, routes}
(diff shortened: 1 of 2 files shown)


sihil:
LGTM 👍

cb372:
I've migrated the CD configs to the new Dynamo table, and checked that the stage lookup code works in CODE.

philwills:
This is good, but I wonder whether for local work we want to use DynamoDB local, rather than DEV tables.

cb372:
Good point. I was going to add an "if stage == DEV then use CODE" fudge when calculating the table name, but I forgot. I'm happy either doing that or using Dynamo local.

sihil:
Ooo DynamoDB local is in maven. That's a cute way to actually test things.

cb372:
For now I've added code to my Play 2.5 branch to use the CODE tables when running on developer machines. We can switch to Dynamo local later if necessary.

philwills:
Sounds reasonable. 

@sihil for the purpose of tests, as opposed to spinning things up to look at locally, I'd argue for https://github.com/localytics/sbt-dynamodb

PR #336, opened 2016-09-12
Retry failed S3 uploads when using InputStream

sihil:
The retry logic in the AWS SDK is flawed when uploading an InputStream that cannot be reset. 

This moves the retry logic for this case outside of the SDK call so that the InputStream can also be reset. The internal retry logic is disabled by passing in a `ClientConfiguration` with a no retry policy. We continue to use the AWS retry and backoff logic from the client configuration - you'll note that the `originalRequest` is not passed in as we do not have it and neither is it actually used.

### DIFF
--- a/magenta-lib/src/main/scala/magenta/package.scala
+++ b/magenta-lib/src/main/scala/magenta/package.scala
@@ -2,6 +2,11 @@ package magenta
 
 import java.io.Closeable
 
+import com.amazonaws.{AmazonClientException, ClientConfiguration}
+
+import scala.annotation.tailrec
+import scala.util.{Failure, Success, Try}
+
 object `package` {
   def transpose[A](xs: Seq[Seq[A]]): Seq[Seq[A]] = xs.filter(_.nonEmpty) match {
     case Nil => Nil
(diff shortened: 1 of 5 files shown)


philwills:
Still not sure why writing to files is such a bad idea, but that aside, this looks sane.

adamnfish:
Since it's written as a nice separate re-usable chunk, is it not the sort of thing that's worth putting a test on?

sihil:
If you write to a file you are limiting yourself to the bandwidth of the disk. In addition you have to have finished writing the file to disk before you can start uploading the file so therefore you have to have downloaded it before you can upload it rather than streaming the two together.

I'm not against writing out to disk but it will at least double the time for upload operations.

philwills:
Yea, I've just rarely seen that be a limiting factor, so I'd tend towards reliability and simplicity.

sihil:
I think it can add 20 seconds or so to some deploys when they are shifting large files around. That approaches 10% of deploy time which seems like an optimisation worth making. If this was super flakey I'd agree, but we've only seen a handful of issues.

If it continues to be a problem then we can revert to a simpler approach.

sihil:
I've added a handful of tests @adamnfish

philwills:
Slightly terrified that we have something that takes 10% of time doing S3 download/upload from within the same region, but 👍  especially for the tests.

sihil:
They are large files. An example I just looked at took 24 seconds to shift a 104Mb file. It would be nearing double that if we had to write it locally (I assume the same time to download and upload which might be incorrect).

PR #338, opened 2016-09-14
Add deprecation warnings for deploy.json configs

sihil:
It's been an aim of mine for a while to have a way of pushing people to migrate from deprecated configurations. I've added a reporter message type of warning which is highlighted beautifully and can be used to give deprecation warnings.

I've added a handful of things to highlight as deprecation warnings:
 - `publicReadAcl=true` on an autoscaling deploy
 - Not uploading via S3 when updating lambdas
 - Not having `defaultStacks` in a deploy.json
 - Using the artifacts.zip format

An example:
screen shot 2016-09-14 at 11 29 34


sihil:
There is definitely no irony in this:
screen shot 2016-09-14 at 15 06 12

philwills:
Well those colours would certainly encourage me to get rid of them.

philwills:
More generally, I think this is a good idea. Particularly the fact that it provides information on a resolution.

PR #339, opened 2016-09-14
Use Dynamo for Hook Configs

philwills:


philwills:
There's quite a bit more can be deleted once this has a 👍 and I've migrated the data to the PROD table.

PR #340, opened 2016-09-14
Bump sbt-riffraff-artifact plugin to use new format

sihil:
We should really migrate ourselves onto the non-legacy artifact format.


sihil:
I tried to build this into a deployable last night and found it didn't work in TeamCity. Either we can move to riffRaffUpload or we need to fix the riffRaffNotifyTeamcity task as suggested in https://github.com/guardian/sbt-riffraff-artifact/pull/27

sihil:
This now works due to https://github.com/guardian/sbt-riffraff-artifact/pull/29

PR #342, opened 2016-09-16
Parser for riff-raff.yaml proposal

sihil:
This implements a parser for the `riff-raff.yaml` [proposal](https://docs.google.com/document/d/1zUQz8cHj8WbcZ-JgbFhtWteFgYtiTCmW771SELxZ5S8/edit).

It consists of two parts, a RiffRaffYamlReader and a DeploymentResolver. The reader turns the YAML in to a set of case classes that map directly from the YAML format. The resolver takes the representation provided by the first and produces a list of deployments by applying the templates and defaults.

None of this is actually hooked up to anything else, but I think makes sense to review on its own.

Reader
-------
The reader converts the YAML to JSON and then uses the Play JSON library. I tried using some YAML libraries to no avail. Someone is welcome to try again if they so desire, but this works well given that the format is a subset of JSON.

Resolver
---------
The resolver deals with applying templates and correctly populating the fields of a deployment according the prioritisation rules for values that come from application defaults, global values, templates and deployments. The resolver returns a list of either of errors or deployments. Errors are returned for missing templates and missing dependencies. This could be extended in time.

### DIFF
--- a/magenta-lib/src/main/scala/magenta/input/DeploymentResolver.scala
+++ b/magenta-lib/src/main/scala/magenta/input/DeploymentResolver.scala
@@ -9,6 +9,7 @@ object DeploymentResolver {
       for {
         templated <- applyTemplates(label, rawDeployment, config.templates).right
         deployment <- resolveDeployment(label, templated, config.stacks, config.regions).right
+        _ <- validateDependencies(label, deployment, config.deployments).right
       } yield deployment
     }
   }
@@ -30,8 +31,6 @@ object DeploymentResolver {
         app               = templated.app.getOrElse(label),
         contentDirectory  = templated.contentDirectory.getOrElse(label),
         dependencies      = templated.dependencies.getOrElse(Nil),
-        // TODO? do we need to validate required parameters here,
-        // or is that up to the deployment type to do later?
         parameters        = templated.parameters.getOrElse(Map.empty)
       )
     }
@@ -62,6 +61,18 @@ object DeploymentResolver {
             parameters        = Some(resolvedParent.parameters.getOrElse(Map.empty) ++ template.parameters.getOrElse(Map.empty))
           )
         }
+    }
+  }
+
+  /**
+    * Ensures that when deployments have named dependencies, deployments with those names exists.
+    */
+  private[input] def validateDependencies(label: String, deployment: Deployment, allDeployments: List[(String, DeploymentOrTemplate)]): Either[(String, String), Deployment] = {
+    deployment.dependencies.filterNot(allDeployments.map(_._1).contains) match {
+      case Nil =>
+        Right(deployment)
+      case missingDependencies =>
+        Left(label -> missingDependencies.mkString(s"Missing deployment dependencies ", ", ", ""))
     }
   }
 }
--- a/magenta-lib/src/main/scala/magenta/input/DeploymentResolver.scala
+++ b/magenta-lib/src/main/scala/magenta/input/DeploymentResolver.scala
@@ -4,7 +4,7 @@ package magenta.input
 object DeploymentResolver {
   val DEFAULT_REGIONS = List("eu-west-1")
 
-  def resolve(config: RiffRaffDeployConfig): List[Either[(String, String), Deployment]] = {
+  def resolve(config: RiffRaffDeployConfig): List[Either[ConfigError, Deployment]] = {
     config.deployments.map { case (label, rawDeployment) =>
       for {
         templated <- applyTemplates(label, rawDeployment, config.templates).right
@@ -18,10 +18,10 @@ object DeploymentResolver {
     * Validates and resolves a templated deployment by merging its
     * deployment attributes with any globally defined properties.
     */
-  private[input] def resolveDeployment(label: String, templated: DeploymentOrTemplate, globalStacks: Option[List[String]], globalRegions: Option[List[String]]): Either[(String, String), Deployment] = {
+  private[input] def resolveDeployment(label: String, templated: DeploymentOrTemplate, globalStacks: Option[List[String]], globalRegions: Option[List[String]]): Either[ConfigError, Deployment] = {
     for {
-      deploymentType <- templated.`type`.toRight(label -> "No type field provided").right
-      stacks <- templated.stacks.orElse(globalStacks).toRight(label -> "No stacks provided").right
+      deploymentType <- templated.`type`.toRight(ConfigError(label, "No type field provided")).right
+      stacks <- templated.stacks.orElse(globalStacks).toRight(ConfigError(label, "No stacks provided")).right
     } yield {
       Deployment(
         name              = label,
@@ -40,14 +40,14 @@ object DeploymentResolver {
     * Recursively apply named templates by merging the provided
     * deployment template with named parent templates.
     */
-  private[input] def applyTemplates(templateName: String, template: DeploymentOrTemplate, templates: Option[Map[String, DeploymentOrTemplate]]): Either[(String, String), DeploymentOrTemplate] = {
+  private[input] def applyTemplates(templateName: String, template: DeploymentOrTemplate, templates: Option[Map[String, DeploymentOrTemplate]]): Either[ConfigError, DeploymentOrTemplate] = {
     template.template match {
       case None =>
         Right(template)
       case Some(parentTemplateName) =>
         for {
           parentTemplate <- templates.flatMap(_.get(parentTemplateName))
-            .toRight(templateName -> s"Template with name $parentTemplateName does not exist").right
+            .toRight(ConfigError(templateName, s"Template with name $parentTemplateName does not exist")).right
           resolvedParent <- applyTemplates(parentTemplateName, parentTemplate, templates).right
         } yield {
           DeploymentOrTemplate(
@@ -67,13 +67,13 @@ object DeploymentResolver {
   /**
     * Ensures that when deployments have named dependencies, deployments with those names exists.
     */
-  private[input] def validateDependencies(label: String, deployment: Deployment, allDeployments: List[(String, DeploymentOrTemplate)]): Either[(String, String), Deployment] = {
+  private[input] def validateDependencies(label: String, deployment: Deployment, allDeployments: List[(String, DeploymentOrTemplate)]): Either[ConfigError, Deployment] = {
     val allDeploymentNames = allDeployments.map { case (name, _) => name }
     deployment.dependencies.filterNot(allDeploymentNames.contains) match {
       case Nil =>
         Right(deployment)
       case missingDependencies =>
-        Left(label -> missingDependencies.mkString(s"Missing deployment dependencies ", ", ", ""))
+        Left(ConfigError(label, missingDependencies.mkString(s"Missing deployment dependencies ", ", ", "")))
     }
   }
 }
--- a/magenta-lib/src/main/scala/magenta/input/DeploymentResolver.scala
+++ b/magenta-lib/src/main/scala/magenta/input/DeploymentResolver.scala
@@ -1,0 +1,68 @@
+package magenta.input
+
+import play.api.libs.json.JsValue
+
+case class Deployment(
+  name: String,
+  `type`: String,
+  stacks: List[String],
+  regions: List[String],
+  dependencies: List[String],
+  app: String,
+  contentDirectory: String,
+  parameters: Map[String, JsValue]
+)
+
+object DeploymentResolver {
+  val DEFAULT_REGIONS = List("eu-west-1")
+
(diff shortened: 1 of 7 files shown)


LATaylor-guardian:
Question: Why the need for YAML?

magenta-lib/src/test/scala/magenta/input/DeploymentResolverTest.scala
```diff
@@ -0,0 +1,298 @@
...
+import play.api.libs.json.{JsNumber, JsString, JsValue}
+
+class DeploymentResolverTest extends FlatSpec with ShouldMatchers {
+  "DeploymentResolver" should "parse a simple deployment with defaults" in {
+    val yaml = RiffRaffYaml(None, None, None,
+      Map("test" -> deploymentType("testType").withParameters("testParam" -> JsString("testValue")).withStacks("testStack")))
+    val deployments = DeploymentResolver.resolve(yaml)
+    deployments.size should be (1)
+    deployments.head.isRight should be(true)
+    val deployment = deployments.head.right.get
```
adamnfish: scalatest can check multiple properties on an object at once.

deployments should have (
  'size (1),
  'type ("testType"),
  'stacks (List("testStack")),
  ... etc
)

It's much less verbose and (I think) easier to take in at a glance.

I think you'll still need to do other property tests that don't check a simple value (like `contain` / `isRight`) separately.

sihil: nice :)

sihil:
@LATaylor-guardian I've added the proposal link to the main body. YAML is easier to work with as you can have comments etc.

magenta-lib/src/main/scala/magenta/input/DeploymentResolver.scala
```diff
@@ -0,0 +1,75 @@
...
+)
+
+object DeploymentResolver {
+  val DEFAULT_REGIONS = List("eu-west-1")
+
+  def getTemplate(name: String, context: String, templates: Option[Map[String, DeploymentOrTemplate]]): Either[List[String], DeploymentOrTemplate] = {
+    templates.flatMap(_.get(name)).fold[Either[List[String], DeploymentOrTemplate]] {
+      Left(List(s"Template with name $name (specified in $context) does not exist"))
+    } { template =>
+      Right(template)
```
adamnfish: Probably clearer to write this as

```scala
templates.flatMap(_.get(name)).toRight(List(s"Template with name $name (specified in $context) does not exist"))
```

adamnfish: Also, at that point it may be worth inlining this function.

adamnfish:
I think I'd quite like to see a few tests that go end-to-end. the unit tests are great for testing the logic is ok but in terms of documenting intent I think it would be helpful to see a few examples of "this yaml produces precisely the following deployment(s)". They'd be providing the same info as the examples in the document where given yaml gets parsed to a more complete set of yaml.

Also, the `DeploymentResolver` stuff looks a bit complex at the moment, various concerns all mixed together. I'm going to try staring at it for a bit to see if it could be expressed more simply.

philwills:
This looks like a reasonable starting point to me.

magenta-lib/src/main/scala/magenta/input/DeploymentResolver.scala
```diff
@@ -0,0 +1,68 @@
...
+  parameters: Map[String, JsValue]
+)
+
+object DeploymentResolver {
+  val DEFAULT_REGIONS = List("eu-west-1")
+
+  def resolveDeployment(name: String, deployment: DeploymentOrTemplate, yaml: RiffRaffYaml): Either[List[String], DeploymentOrTemplate] = {
+    val missingDeps = yaml.missingDependencies(deployment.dependencies.getOrElse(Nil))
+    if (missingDeps.nonEmpty)
+      Left(missingDeps.map(dep => s"Missing dependency $dep in $name"))
```
adamnfish: Pretty sure this is catching the same problem as the `"Template with name ...` error message further down?

magenta-lib/src/main/scala/magenta/input/DeploymentResolver.scala
```diff
@@ -0,0 +1,68 @@
...
+      }
+    }
+  }
+
+  def resolve(yaml: RiffRaffYaml): List[Either[List[String], Deployment]] = {
+    yaml.deployments.map { case (name, deployment) =>
+      resolveDeployment(name, deployment, yaml).right.map { resolved =>
+        Deployment(
+          name = name,
+          `type` = resolved.`type`.get,
```
adamnfish: Shame to have the Either type for passing errors, but then not actually use it when there are problems. I can't see anything that's guarding this `.get` so presumably you'll just get exceptions at the moment?

magenta-lib/src/main/scala/magenta/input/DeploymentResolver.scala
```diff
@@ -0,0 +1,68 @@
...
+        case deployment @ DeploymentOrTemplate(Some(deploymentType), None, stacks, regions, _, _, _, _) =>
+          // terminating case - apply any default stacks and regions
+          val deploymentWithGlobalDefaults = deployment.copy(stacks = stacks.orElse(yaml.stacks), regions = regions.orElse(yaml.regions))
+          Right(deploymentWithGlobalDefaults)
+
+        case template @ DeploymentOrTemplate(None, Some(templateName), stacks, regions, app, contentDirectory, dependencies, parameters) =>
+          // template case - recursively resolve
+          for {
+            template <- yaml.templates.flatMap(_.get(templateName))
+              .toRight(List(s"Template with name $templateName (specified in $name) does not exist")).right
```
adamnfish: I think it might be nice to pass this context ('specified in') as part of the error type, rather than embedding it in the strings. Means whatever client is showing the errors has the opportunity to do something clever. I think using `(String, String)` as the left type (for `context -> error`) would be nicer. With this design I don't think we'll ever use List[String] for errors and it should be a fairly easy change if the need arises.

magenta-lib/src/test/scala/magenta/input/DeploymentResolverTest.scala
```diff
@@ -0,0 +1,302 @@
...
+      Some(Map(
+        "nestedTemplate" -> deploymentType("testType").withDependencies("nested-dep"),
+        "testTemplate" -> deploymentTemplate("nestedTemplate")
+      )),
+      Map("test" -> deploymentTemplate("nonExistentTemplate"))
+    )
+    val deployments = DeploymentResolver.resolve(yaml)
+    deployments.size should be (1)
+    deployments.head.isLeft should be(true)
+    val errors = deployments.head.left.get
```
adamnfish: scalatest has a useful trait called `EitherValues`

magenta-lib/src/test/scala/magenta/input/DeploymentResolverTest.scala
```diff
@@ -0,0 +1,302 @@
...
+    errors should be(List("Template with name nonExistentTemplate (specified in test) does not exist"))
+  }
+
+  it should "report missing dependencies" in {
```
adamnfish: As with the logic that produces the errors, this seems to be testing the same thing as the above test?

magenta-lib/src/main/scala/magenta/input/DeploymentResolver.scala
sihil: No - the template with name error catches when the `template` field is referencing a non-existent template in the `templates` section. This is catching when any of the `dependencies` lists are referring to a deployment that doesn't exist. They are similar but not the same.

magenta-lib/src/main/scala/magenta/input/DeploymentResolver.scala
sihil: Agreed. I like the change in your PR - although wonder if it still needs to be a list.

magenta-lib/src/main/scala/magenta/input/DeploymentResolver.scala
sihil: The base case of the recursive `resolveDeployment` returns a `DeploymentOrTemplate` that has `type=Some(x)`.

sihil:
@philwills - @adamnfish and I have honed this somewhat. fancy another look before it gets merged?

magenta-lib/src/main/scala/magenta/input/DeploymentResolver.scala
```diff
@@ -0,0 +1,78 @@
+package magenta.input
+
+
+object DeploymentResolver {
+  val DEFAULT_REGIONS = List("eu-west-1")
+
+  def resolve(config: RiffRaffDeployConfig): List[Either[(String, String), Deployment]] = {
```
philwills: I'd definitely prefer some more descriptive type than `(String, String)` here.

magenta-lib/src/main/scala/magenta/input/DeploymentResolver.scala
```diff
@@ -0,0 +1,78 @@
...
+          )
+        }
+    }
+  }
+
+  /**
+    * Ensures that when deployments have named dependencies, deployments with those names exists.
+    */
+  private[input] def validateDependencies(label: String, deployment: Deployment, allDeployments: List[(String, DeploymentOrTemplate)]): Either[(String, String), Deployment] = {
+    deployment.dependencies.filterNot(allDeployments.map(_._1).contains) match {
```
philwills: The use of `_._1` makes this a bit cryptic.

magenta-lib/src/test/scala/magenta/input/RiffRaffYamlReaderTest.scala
```diff
@@ -0,0 +1,73 @@
+package magenta.input
+
+import org.scalatest.{FlatSpec, ShouldMatchers}
+import play.api.libs.json.{JsArray, JsNumber, JsString, Json}
+
+class RiffRaffYamlReaderTest extends FlatSpec with ShouldMatchers{
```
philwills: Is this still needed now the other tests work on YAML?

adamnfish: This is required to narrow down failures if the parser code is broken. Having tests of the logic depend on external functionality isn't ideal, but with this additional coverage problems emerging from that coupling are less severe.

magenta-lib/src/main/scala/magenta/input/DeploymentResolver.scala
adamnfish: Yeah, I agree. I tried it written out first and the extra noise doesn't really help much. I'd hoped the intent of this line was clear enough that the `_1` was ok.

That said, I'm always nervous about using `_n` so I'm very happy to change it if there's a feeling that it isn't clear enough here.

sihil: I've extracted it out of the `validateDependencies` method and used an extractor to make it clearer.

magenta-lib/src/test/scala/magenta/input/RiffRaffYamlReaderTest.scala
sihil: I think these tests would make any problems with the YAML reading easier to understand so I'm tempted to keep them.

magenta-lib/src/main/scala/magenta/input/DeploymentResolver.scala
sihil: A case class? There was some thought about replacing Either with Validated from cats, but possibly a little later once we understand what we are doing a little better.

adamnfish: Using simple built in types is nice though. There's definitely a point where this becomes complex enough for it to warrant the cost of using a case class everywhere to get named properties. I don't think 'two strings' is quite at that point (although I concede as the co-author my opinion is biased here). I'll give it a try though, see how it looks.

magenta-lib/src/test/scala/magenta/input/RiffRaffYamlReaderTest.scala
philwills: OK. I'm not vehemently against them.

sihil:
I don't like the new reviewing functionality. Lack of automatic updates or warnings makes it hard for two people to review at once... 💩

magenta-lib/src/main/scala/magenta/input/DeploymentResolver.scala
philwills: `Either` is going to be right-biased in 2.12, so cats' `Xor` is going away.

I really don't know what those `String`s are, so would definitely value a case class. 

If in doubt, stick a type on it.

PR #343, opened 2016-09-16
Suggest 0.9.2, not 0.9.1

philwills:


sihil:
There is another pr that does this. As well as improve the check, but I'm happy to do just this for now.

sihil:
:+1:

PR #344, opened 2016-09-16
Update to the riff-raff.yaml proposal

adamnfish:
This commit addresses the points I raised in my review of [the `riff-raff.yaml` proposal PR](https://github.com/guardian/riff-raff/pull/342).
* handles missing required fields in deployments
* makes use of scalatest's EitherValues trait
* passes a context in the Left of the Either to enable display of which block the error refers to

It also:
* simplifies the DeploymentResolver logic
* separates the models from the business logic
* documents the DeploymentOrTemplate model

### DIFF
--- a/magenta-lib/src/main/scala/magenta/input/DeploymentResolver.scala
+++ b/magenta-lib/src/main/scala/magenta/input/DeploymentResolver.scala
@@ -1,68 +1,67 @@
 package magenta.input
 
-import play.api.libs.json.JsValue
-
-case class Deployment(
-  name: String,
-  `type`: String,
-  stacks: List[String],
-  regions: List[String],
-  dependencies: List[String],
-  app: String,
-  contentDirectory: String,
-  parameters: Map[String, JsValue]
-)
 
 object DeploymentResolver {
   val DEFAULT_REGIONS = List("eu-west-1")
 
-  def resolveDeployment(name: String, deployment: DeploymentOrTemplate, yaml: RiffRaffYaml): Either[List[String], DeploymentOrTemplate] = {
-    val missingDeps = yaml.missingDependencies(deployment.dependencies.getOrElse(Nil))
-    if (missingDeps.nonEmpty)
-      Left(missingDeps.map(dep => s"Missing dependency $dep in $name"))
-    else {
-      deployment match {
-        case deployment @ DeploymentOrTemplate(Some(deploymentType), None, stacks, regions, _, _, _, _) =>
-          // terminating case - apply any default stacks and regions
-          val deploymentWithGlobalDefaults = deployment.copy(stacks = stacks.orElse(yaml.stacks), regions = regions.orElse(yaml.regions))
-          Right(deploymentWithGlobalDefaults)
-
-        case template @ DeploymentOrTemplate(None, Some(templateName), stacks, regions, app, contentDirectory, dependencies, parameters) =>
-          // template case - recursively resolve
-          for {
-            template <- yaml.templates.flatMap(_.get(templateName))
-              .toRight(List(s"Template with name $templateName (specified in $name) does not exist")).right
-            resolved <- resolveDeployment(templateName, template, yaml).right
-          } yield {
-            DeploymentOrTemplate(
-              resolved.`type`,
-              None,
-              stacks.orElse(resolved.stacks),
-              regions.orElse(resolved.regions),
-              app.orElse(resolved.app),
-              contentDirectory.orElse(resolved.contentDirectory),
-              dependencies.orElse(resolved.dependencies),
-              Some(resolved.parameters.getOrElse(Map.empty) ++ parameters.getOrElse(Map.empty))
-            )
-          }
-      }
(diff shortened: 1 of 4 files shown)


magenta-lib/src/main/scala/magenta/input/DeploymentResolver.scala
```diff
@@ -1,68 +1,67 @@
...
+    } yield {
+      Deployment(
+        name              = label,
+        `type`            = deploymentType,
+        stacks            = stacks,
+        regions           = templated.regions.orElse(globalRegions).getOrElse(DEFAULT_REGIONS),
+        app               = templated.app.getOrElse(label),
+        contentDirectory  = templated.contentDirectory.getOrElse(label),
+        dependencies      = templated.dependencies.getOrElse(Nil),
+        // TODO? do we need to validate required parameters here,
```
adamnfish: Any thoughts on this?

magenta-lib/src/main/scala/magenta/input/models.scala
```diff
@@ -0,0 +1,55 @@
+package magenta.input
+
+import play.api.libs.json.{JsValue, Json, Reads}
+
+
+case class RiffRaffDeployConfig(
```
adamnfish: There's nothing yaml specific about this class, so I've proposed changing its name. Not particularly wedded to this one, suggestions welcome.

magenta-lib/src/main/scala/magenta/input/DeploymentResolver.scala
sihil: I think the earlier the better for this. I think at this stage we should validate everything that doesn't require knowledge of available types or valid parameters. The dependencies don't require that knowledge so I think this is the right place to do it.

There are definitely more validations to do later, but I don't think this is one of them.

sihil: Totally mis-read that. No - I don't think we should validate deployment types or parameters here. At this stage we are just creating deployment types. We should validate that dependencies exist.

magenta-lib/src/test/scala/magenta/input/DeploymentResolverTest.scala
```diff
@@ -259,27 +249,22 @@ class DeploymentResolverTest extends FlatSpec with ShouldMatchers {
     )
     val deployments = DeploymentResolver.resolve(yaml)
     deployments.size should be (1)
-    deployments.head.isLeft should be(true)
-    val errors = deployments.head.left.get
-    errors should be(List("Template with name nonExistentTemplate (specified in test) does not exist"))
+    deployments.head.left.value should be("test" -> "Template with name nonExistentTemplate does not exist")
   }
 
-  it should "report missing dependencies" in {
```
sihil: Think we should keep this test.

magenta-lib/src/main/scala/magenta/input/models.scala
sihil: That name is fine by me.

magenta-lib/src/main/scala/magenta/input/models.scala
```diff
@@ -0,0 +1,55 @@
...
+  *
+  * @param `type`           The type of deployment to perform (e.g. autoscaling, s3).
+  * @param template         Name of the custom deploy template to use for this deployment.
+  * @param stacks           Stack tags to apply to this deployment. The deployment will be executed once for each stack.
+  * @param regions          A list of the regions in which this deploy will be executed. Defaults to just 'eu-west-1'
+  * @param app              The `app` tag to use for this deployment. By default the deployment's key is used.
+  * @param contentDirectory The path where this deployment is found in the build output. Defaults to app.
+  * @param dependencies     This deployment's execution will be delayed until all named dependencies have completed. (Default empty)
+  * @param parameters       Provides additional parameters to the deployment type. Refer to the deployment types to see what is required.
+  */
```
sihil: 👍 for the docs. ultimately we need to work out how to present this to users.

magenta-lib/src/test/scala/magenta/input/DeploymentResolverTest.scala
adamnfish: How does it differ to the one above?

"Give a useful message if it fails to resolve a dependency"

magenta-lib/src/main/scala/magenta/input/DeploymentResolver.scala
adamnfish: Yep, that was my thought as well, just wanted to check it here. I'll take the comment out in the office.

adamnfish:
Oops, I only tested the `lib` project locally. Presumably it's the class name change, I'll fix the build in the office.

magenta-lib/src/test/scala/magenta/input/DeploymentResolverTest.scala
sihil: The former is about templates and the latter is about dependencies. Two different concepts and I think it's useful to distinguish between the errors.

PR #345, opened 2016-09-20
Restructure classes and naming

sihil:
This is a housekeeping PR that:
 - separates the classes in `actors.scala` into their own files
 - renames a handful of concepts so they are clearer and less likely to clash


sihil:
I'm also not overly enthused by `Tasks` - I'd prefer something like `TaskBundle` or `DeploymentTasks`. @adamnfish has strong opinions about naming thin wrappers. 🚀

adamnfish:
`DeploymentTasks` is ok, I'm not sure what `TaskBundle` adds to Tasks apart from characters.

PR #347, opened 2016-09-20
Eliminate YAML conversion

sihil:
Use the new native YAML support rather than our poor mans conversion.


cb372:
[I](https://github.com/guardian/riff-raff/pull/261) was months ahead of my time...

PR #351, opened 2016-09-22
Fix post deploy hooks

sihil:
A couple of problems have combined to mean that post deploy hooks haven't been working for a while.
 - Some actor wiring was slightly off since the Play 2.5 upgrade
 - The role permissions needed to grant access to the index - done in https://github.com/guardian/deploy-tools-platform/pull/46

I've also tidied up the `Lifecycle` traits.

### DIFF
--- a/riff-raff/app/notification/hooks.scala
+++ b/riff-raff/app/notification/hooks.scala
@@ -85,13 +85,10 @@ object HookConfig {
     HookConfig(UUID.randomUUID(), projectName, stage, url, enabled, new DateTime(), updatedBy)
 }
 
-object HooksClient extends Lifecycle with Logging {
-  trait Event
-  case class Finished(uuid: UUID, params: DeployParameters)
-
+class HooksClient(wsClient: WSClient) extends Lifecycle with Logging {
   lazy val system = ActorSystem("notify")
   val actor = try {
-    Some(system.actorOf(Props[HooksClient], "hook-client"))
+    Some(system.actorOf(Props(classOf[HooksClientActor], wsClient), "hook-client"))
   } catch {
     case t:Throwable =>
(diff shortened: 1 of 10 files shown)


riff-raff/app/notification/hooks.scala
```diff
@@ -85,17 +85,18 @@ object HookConfig {
...
-object HooksClient extends LifecycleWithoutApp with Logging {
-  trait Event
-  case class Finished(uuid: UUID, params: DeployParameters)
-
+class HooksClient(wsClient: WSClient) extends Lifecycle with Logging {
   lazy val system = ActorSystem("notify")
   val actor = try {
-    Some(system.actorOf(Props[HooksClient], "hook-client"))
-  } catch { case t:Throwable => None }
+    Some(system.actorOf(Props(classOf[HooksClientActor], wsClient), "hook-client"))
```
cb372: You could avoid the problem of this breaking silently when the constructor is refactored by using the alternative form of `Props`:

```
system.actorOf(Props(new HooksClientActor(wsClient))
```

I think there are a few other instances of `Props` elsewhere in the codebase that could be refactored in the same way.

cb372:
Looks fine apart from that 👍

riff-raff/app/notification/hooks.scala
sihil: A nice idea, but my understanding was that use of that form is discouraged - see this post for example: http://www.cakesolutions.net/teamblogs/understanding-akkas-recommended-practice-for-actor-creation-in-scala

sihil: I'm happy to switch everything over if we conclude that is a good thing to do (I'll add it to tonights agenda 😉). In the meantime I'm going to merge this as it is.

PR #352, opened 2016-09-22
Add the ability to deploy lambda cross-regions

mchv:
The associated changes add the capability to deploy a lambda in multiple regions which can be specified in the deployment configuration file.

If not provided, `eu-west-1` region (the current default) is used.

### DIFF
--- a/magenta-lib/src/main/scala/magenta/deployment_type/Lambda.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/Lambda.scala
@@ -1,7 +1,9 @@
 package magenta.deployment_type
 
 import java.io.File
+import scala.util.Try
 
+import com.amazonaws.regions.{Region, Regions}
 import magenta.artifact.S3Path
 import magenta.{DeployParameters, DeployReporter, DeploymentPackage, KeyRing, Stack, Stage}
 import magenta.tasks.{S3Upload, UpdateLambda, UpdateS3Lambda}
@@ -21,6 +23,14 @@ object Lambda extends DeploymentType  {
       |powerful `functions` parameter. This lets you bind a specific file to a specific function for any given stage.
       |As a result you can bundle stage specific configuration into the respective files.
     """.stripMargin
+
+  val regionsParam = Param[List[String]]("regions",
+    documentation = 
+      """
+      |One or more AWS region name where the lambda should be deployed - If this is not specified then `eu-west-1` will be used.
+      """.stripMargin,
+      defaultValue = Some(List("eu-west-1"))
+  )
 
   val bucketParam = Param[String]("bucket",
     documentation =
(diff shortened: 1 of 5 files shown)


magenta-lib/src/main/scala/magenta/deployment_type/Lambda.scala
```diff
@@ -22,6 +22,13 @@ object Lambda extends DeploymentType  {
       |As a result you can bundle stage specific configuration into the respective files.
     """.stripMargin
 
+  val regionsParam = Param[List[String]]("regions",
+    documentation = 
+      """
+      |One or more AWS region name where the lambda should be deployed - If this is not specified then `eu-west-1` will be used.
+      """.stripMargin
+  )
```
sihil: `Param` has a `withDefault` method which would be a better place to deal with the `eu-west-1` default.

magenta-lib/src/main/scala/magenta/deployment_type/Lambda.scala
```diff
@@ -69,15 +76,25 @@ object Lambda extends DeploymentType  {
   def lambdaToProcess(pkg: DeploymentPackage, stage: String, stack: Stack)(reporter: DeployReporter): List[LambdaFunction] = {
     val bucketOption = bucketParam.get(pkg)
     if (bucketOption.isEmpty) reporter.warning(s"DEPRECATED: Uploading directly to lambda is deprecated (it is dangerous for CloudFormed lambdas). Specify the bucket parameter and call both uploadLambda and updateLambda to upload via S3.")
+    
+    val regions = regionsParam.get(pkg).map(_.map(Region)).getOrElse(List(DefaultRegion))
```
sihil: As mentioned above the default can be dealt with before you get here

magenta-lib/src/main/scala/magenta/tasks/AWS.scala
```diff
@@ -18,15 +18,21 @@ import magenta.{App, DeployReporter, DeploymentPackage, KeyRing, NamedStack, Sta
...
   def s3client(keyRing: KeyRing, config: ClientConfiguration = clientConfiguration) =
     new AmazonS3Client(provider(keyRing), config)
 }
 
 trait Lambda extends AWS {
-  def lambdaClient(implicit keyRing: KeyRing) = {
+  def lambdaClient(region: Region)(implicit keyRing: KeyRing) = {
     val client = new AWSLambdaClient(provider(keyRing), clientConfiguration)
-    client.setEndpoint("lambda.eu-west-1.amazonaws.com")
+    client.setEndpoint(s"lambda.${region}.amazonaws.com")
```
sihil: I'd rather we switch to using the `client.setRegion` notation rather than encoding endpoints in Riff-Raff when the SDK now knows it. I think you can do `.withRegion` on the client.

magenta-lib/src/main/scala/magenta/tasks/AWS.scala
```diff
@@ -18,15 +18,21 @@ import magenta.{App, DeployReporter, DeploymentPackage, KeyRing, NamedStack, Sta
 
 import scala.collection.JavaConversions._
 
+sealed case class Region(val host: String) {
+  override def toString = host
+}
```
sihil: Is the `Region` object in the AWS SDK an appropriate alternative here?

magenta-lib/src/main/scala/magenta/tasks/AWS.scala
```diff
@@ -18,15 +18,21 @@ import magenta.{App, DeployReporter, DeploymentPackage, KeyRing, NamedStack, Sta
...
+
+object DefaultRegion extends Region("eu-west-1")
```
sihil: This should probably be in the config file.

mchv:
@sihil thanks for the review I will update code with your suggestions

mchv:
@sihil udpated the code, hopefully look better

magenta-lib/src/main/scala/magenta/deployment_type/Lambda.scala
```diff
@@ -69,19 +79,39 @@ object Lambda extends DeploymentType  {
...
+
       case _ => reporter.fail("Must specify one of 'functions' or 'functionNames' parameters")
     }
   }
 
+  private def toRegion(name: String)(reporter: DeployReporter): Option[Region] = {
+    /* sdk can throw IllegalArgumentException if region name is unknown */
+    val region = Try(Region.getRegion(Regions.fromName(name))).toOption
+    if (region.isEmpty) {
+       reporter.warning(s"${name} is not a recognised AWS region and has been ignored.")
```
sihil: Nice 👍

PR #353, opened 2016-09-23
Build a deployment graph from a list of deployments

sihil:
From a list of deployments with dependencies we should build a graph of deployments. This is achieved using a graph builder that creates all of the necessary edges to fulfil the deployments. We create three sets of edges:
 - from the `StartNode` to all nodes that have no listed dependencies (i.e. nothing goes before them)
 - from a dependency node to a dependent node (i.e. each edge that is expressed by an entry in a deployment's dependencies list)
 - from any node on which no nodes depend to the `EndNode`

These three sets joined together creates a complete graph for the Deployments.

It was realised that building a graph at this stage only makes sense if we can make later transformations easy. As such this PR also implements a `flatMap` on `Graph` that allows individual nodes to be replaced with multiple nodes. This makes sense for splitting out actions for example or even stacks and regions - although this is likely to be done in completely different graphs which are later joined together.

Finally, this PR also improves the DSL for creating edges so that priorities can be more cleanly expressed - mostly done during tests.

### DIFF
--- a/magenta-lib/src/main/scala/magenta/graph/Graph.scala
+++ b/magenta-lib/src/main/scala/magenta/graph/Graph.scala
@@ -22,9 +22,12 @@ case class Edge[+T](from: Node[T], to: Node[T], priority: Int = 1) {
 object Graph {
   def apply[T](edges: Edge[T]*): Graph[T] = Graph(edges.toSet)
 
-  def apply[T](singleton: T): Graph[T] = Graph(StartNode ~> MidNode(singleton), MidNode(singleton) ~> EndNode)
+  def apply[T](singleton: T): Graph[T] = {
+    val node = MidNode(singleton)
+    Graph(StartNode ~> node, node ~> EndNode)
+  }
 
-  def from[T](values: T*): Graph[T] = {
+  def from[T](values: Seq[T]): Graph[T] = {
     val nodes: Seq[Node[T]] = StartNode +: values.map(MidNode.apply) :+ EndNode
     val edges = nodes.sliding(2).map{ window => window.head ~> window.tail.head }.toSet
     Graph(edges)
@@ -101,10 +104,10 @@ case class Graph[T](edges: Set[Edge[T]]) {
     * End or Mid nodes with multiple nodes. The incoming and outgoing edges of each node of this graph will be replaced
     * by edges that map to the start and end nodes of the graph returned from f.
     *
-    * An identity mapping for flatMap would be start and end nodes to empty map and mid nodes to a single value graph
+    * An identity mapping for flatMap would be start and end nodes to empty graph and mid nodes to a single value graph
     * containing the value of the node.
     *
-    * It is not possible to map a mid node to an empty node.
+    * It is not possible to map a mid node to an empty graph.
     * @param f the function to apply to each element
     * @return a new graph with the graphs substituted
     */
--- a/magenta-lib/src/main/scala/magenta/graph/Graph.scala
+++ b/magenta-lib/src/main/scala/magenta/graph/Graph.scala
@@ -2,6 +2,7 @@ package magenta.graph
 
 sealed trait Node[+T] {
   def ~>[R >: T](to: Node[R]): Edge[R] = Edge(this, to)
+  def ~[R >: T](priority: Int): EdgeBuilder[R] = EdgeBuilder(this, priority)
   def maybeValue: Option[T] = None
 }
 case object StartNode extends Node[Nothing]
@@ -10,21 +11,50 @@ case class MidNode[T](value: T) extends Node[T] {
 }
 case object EndNode extends Node[Nothing]
 
+case class EdgeBuilder[+T](from: Node[T], priority: Int) {
+  def ~>[R >: T](to: Node[R]): Edge[R] = Edge(from, to, priority)
+}
+
 case class Edge[+T](from: Node[T], to: Node[T], priority: Int = 1) {
-  def incPriority(n: Int) = this.copy(priority = priority + n)
   def ~=[R >: T](other: Edge[R]) = from == other.from && to == other.to
 }
 
 object Graph {
   def apply[T](edges: Edge[T]*): Graph[T] = Graph(edges.toSet)
 
+  def apply[T](singleton: T): Graph[T] = Graph(StartNode ~> MidNode(singleton), MidNode(singleton) ~> EndNode)
+
+  def from[T](values: T*): Graph[T] = {
+    val nodes: Seq[Node[T]] = StartNode +: values.map(MidNode.apply) :+ EndNode
+    val edges = nodes.sliding(2).map{ window => window.head ~> window.tail.head }.toSet
+    Graph(edges)
+  }
+
   def empty[T] = Graph[T](StartNode ~> EndNode)
+
+  /**
+    * Generate the set of edges that are needed to join the end of the from graph to the start of the to graph
+ *
+    * @param from The graph whose end you wish to join
+    * @param to The graph whose start you wish to join
+    * @return Set of joining edges that would join these two graphs together
+    */
+  private[graph] def joiningEdges[R](from: Graph[R], to: Graph[R]): Set[Edge[R]] = {
+    val fromEndEdges = from.incoming(EndNode)
+    val toStartEdges = to.orderedOutgoing(StartNode)
+    fromEndEdges.foldLeft(Set.empty[Edge[R]]) { case (acc, endEdge) =>
+      val existing = from.orderedOutgoing(endEdge.from)
+      val targetEdges = toStartEdges.map(startEdge => Edge(endEdge.from, startEdge.to))
+      val replacement = existing.replace(endEdge, targetEdges).reprioritise
+      acc -- existing ++ replacement
+    }
+  }
 }
 
 case class Graph[T](edges: Set[Edge[T]]) {
   val nodes = edges.map(_.from) ++ edges.map(_.to)
-  val underlyingNodes = nodes.flatMap(_.maybeValue)
-  val dataNodes = nodes.filterMidNodes
+  private val underlyingNodes = nodes.flatMap(_.maybeValue)
+  private val dataNodes = nodes.filterMidNodes
 
   val isValid: Either[List[String], Boolean] = {
     val errors =
@@ -66,6 +96,50 @@ case class Graph[T](edges: Set[Edge[T]]) {
     Graph(newEdges)
   }
 
+  /**
+    * Flat map keeps the shape of this graph whilst mapping each node to a graph. This means one can replace the Start,
+    * End or Mid nodes with multiple nodes. The incoming and outgoing edges of each node of this graph will be replaced
+    * by edges that map to the start and end nodes of the graph returned from f.
+    *
+    * An identity mapping for flatMap would be start and end nodes to empty map and mid nodes to a single value graph
+    * containing the value of the node.
+    *
+    * It is not possible to map a mid node to an empty node.
+    * @param f the function to apply to each element
+    * @return a new graph with the graphs substituted
+    */
+  def flatMap[R](f: Node[T] => Graph[R]): Graph[R] = {
+    val nodeToGraphMap: Map[Node[T], Graph[R]] = nodes.map(n => n -> f(n)).toMap
+    // find all of the edges that we won't adjoin to another graph
+    val allInternalEdges = nodeToGraphMap.flatMap{
+      case (StartNode, graph) => graph.edges -- graph.incoming(EndNode)
+      case (MidNode(_), graph) =>
+        assert(!graph.isEmpty, "It is not possible to flatMap a MidNode to an empty graph")
+        graph.edges -- graph.incoming(EndNode) -- graph.outgoing(StartNode)
+      case (EndNode, graph) => graph.edges -- graph.outgoing(StartNode)
+    }.toSet
+
+    // sort the edges by priority - this means we can re-calculate the priorities correctly
+    val joiningEdges = edges.toList.sortBy(_.priority)
+      .foldLeft(Set.empty[Edge[R]]) { case(acc, oldEdge) =>
+
+      val fromGraph = nodeToGraphMap(oldEdge.from)
+      val toGraph = nodeToGraphMap(oldEdge.to)
+      // get the joining edges
+      val newEdges = Graph.joiningEdges(fromGraph, toGraph)
+        // deduplicate any edges that we've already generated
+        .filterNot(e1 => acc.exists(e1 ~= _))
+        // fix up the priorities
+        .map{ newEdge =>
+          val maxPriority = acc.filter(_.from == newEdge.from).map(_.priority).reduceOption(Math.max).getOrElse(0)
+          newEdge.copy(priority = newEdge.priority + maxPriority)
+        }
+      acc ++ newEdges
+    }
+
+    Graph(allInternalEdges ++ joiningEdges)
+  }
+
   def joinParallel(other: Graph[T]): Graph[T] = {
     if (other.isEmpty) this
     else if (isEmpty) other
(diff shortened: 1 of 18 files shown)


magenta-lib/src/main/scala/magenta/graph/Graph.scala
```diff
@@ -10,21 +11,50 @@ case class MidNode[T](value: T) extends Node[T] {
...
+
 case class Edge[+T](from: Node[T], to: Node[T], priority: Int = 1) {
-  def incPriority(n: Int) = this.copy(priority = priority + n)
   def ~=[R >: T](other: Edge[R]) = from == other.from && to == other.to
 }
 
 object Graph {
   def apply[T](edges: Edge[T]*): Graph[T] = Graph(edges.toSet)
 
+  def apply[T](singleton: T): Graph[T] = Graph(StartNode ~> MidNode(singleton), MidNode(singleton) ~> EndNode)
```
adamnfish: If there's any chance the "set-like" behaviour of the graph will change one day it'd be better to write this as something like:

```scala
val node = MidNode(singleton)
Graph(StartNode ~> node, node ~> EndNode)
```

If that's a fundamental property of this graph then no worries, although I that isn't super clear from the source files at the moment.

magenta-lib/src/main/scala/magenta/graph/Graph.scala
```diff
@@ -66,6 +96,50 @@ case class Graph[T](edges: Set[Edge[T]]) {
...
+
+      val fromGraph = nodeToGraphMap(oldEdge.from)
+      val toGraph = nodeToGraphMap(oldEdge.to)
+      // get the joining edges
+      val newEdges = Graph.joiningEdges(fromGraph, toGraph)
+        // deduplicate any edges that we've already generated
+        .filterNot(e1 => acc.exists(e1 ~= _))
+        // fix up the priorities
+        .map{ newEdge =>
+          val maxPriority = acc.filter(_.from == newEdge.from).map(_.priority).reduceOption(Math.max).getOrElse(0)
```
adamnfish: Would it be simpler to replace the map and following statements with `.maxBy(_.priority.getOrElse(0))`?

magenta-lib/src/main/scala/magenta/graph/Graph.scala
```diff
@@ -66,6 +96,50 @@ case class Graph[T](edges: Set[Edge[T]]) {
...
 
+  /**
+    * Flat map keeps the shape of this graph whilst mapping each node to a graph. This means one can replace the Start,
+    * End or Mid nodes with multiple nodes. The incoming and outgoing edges of each node of this graph will be replaced
+    * by edges that map to the start and end nodes of the graph returned from f.
+    *
+    * An identity mapping for flatMap would be start and end nodes to empty map and mid nodes to a single value graph
+    * containing the value of the node.
+    *
+    * It is not possible to map a mid node to an empty node.
```
adamnfish: Why not?

magenta-lib/src/main/scala/magenta/graph/Graph.scala
```diff
@@ -66,6 +96,50 @@ case class Graph[T](edges: Set[Edge[T]]) {
     Graph(newEdges)
   }
 
+  /**
+    * Flat map keeps the shape of this graph whilst mapping each node to a graph. This means one can replace the Start,
+    * End or Mid nodes with multiple nodes. The incoming and outgoing edges of each node of this graph will be replaced
+    * by edges that map to the start and end nodes of the graph returned from f.
+    *
+    * An identity mapping for flatMap would be start and end nodes to empty map and mid nodes to a single value graph
```
adamnfish: What does "empty map" mean in this context?

magenta-lib/src/main/scala/magenta/graph/Graph.scala
```diff
@@ -113,13 +180,17 @@ case class Graph[T](edges: Set[Edge[T]]) {
   lazy val toList: List[T] = nodeList.flatMap(_.maybeValue)
 
   override def toString: String = {
```
adamnfish: Would an output like the `tree` command's be easier to visualise than separate nodes and edges?

If I remember (and see) correctly, 'priority' really means order, so it's captured by an ordered representation without needing to be displayed separately?

magenta-lib/src/main/scala/magenta/graph/Graph.scala
```diff
@@ -10,21 +11,50 @@ case class MidNode[T](value: T) extends Node[T] {
...
-  def incPriority(n: Int) = this.copy(priority = priority + n)
   def ~=[R >: T](other: Edge[R]) = from == other.from && to == other.to
 }
 
 object Graph {
   def apply[T](edges: Edge[T]*): Graph[T] = Graph(edges.toSet)
 
+  def apply[T](singleton: T): Graph[T] = Graph(StartNode ~> MidNode(singleton), MidNode(singleton) ~> EndNode)
+
+  def from[T](values: T*): Graph[T] = {
```
adamnfish: This is only ever called with a list. Would it be better to take a list instead of varargs?

magenta-lib/src/main/scala/magenta/graph/Graph.scala
adamnfish: Ah, maybe you're talking about the `nodeToGraphMap`? That's an implementation detail, so probably not helpful for the docstring?

magenta-lib/src/main/scala/magenta/graph/Graph.scala
```diff
@@ -66,6 +96,50 @@ case class Graph[T](edges: Set[Edge[T]]) {
...
+    * @param f the function to apply to each element
+    * @return a new graph with the graphs substituted
+    */
+  def flatMap[R](f: Node[T] => Graph[R]): Graph[R] = {
+    val nodeToGraphMap: Map[Node[T], Graph[R]] = nodes.map(n => n -> f(n)).toMap
+    // find all of the edges that we won't adjoin to another graph
+    val allInternalEdges = nodeToGraphMap.flatMap{
+      case (StartNode, graph) => graph.edges -- graph.incoming(EndNode)
+      case (MidNode(_), graph) =>
+        assert(!graph.isEmpty, "It is not possible to flatMap a MidNode to an empty graph")
```
adamnfish: Be nice if this was enforced by the compiler instead of being a runtime error.

Seems to me that mapping a node to an empty Graph should remove it from the Graph though, I think that's ok at a high level?

magenta-lib/src/main/scala/magenta/input/resolver/DeploymentGraphBuilder.scala
```diff
@@ -0,0 +1,46 @@
...
+
+  private[resolver] def startEdges(deployments: List[Deployment]): List[Edge[Deployment]] = {
+    val targetNodes = for {
+      deployment <- deployments
+      if !deployment.dependencies.exists(deployments.names.contains)
+    } yield MidNode(deployment)
+    edges(StartNode, targetNodes)
+  }
+
+  private[resolver] def midEdges(deployments: List[Deployment]): List[Edge[Deployment]] = {
```
adamnfish: "mid" is a bit overloaded here, what with it being `MidNode` as well.

magenta-lib/src/main/scala/magenta/input/resolver/DeploymentGraphBuilder.scala
```diff
@@ -0,0 +1,46 @@
+package magenta.input.resolver
+
+import magenta.graph.{Edge, EndNode, Graph, MidNode, Node, StartNode}
+import magenta.input.Deployment
+
+object DeploymentGraphBuilder {
```
adamnfish: I fail to see how this captures a dependency chain. Might need you to take me through it.

A depends on B, which depends on C, which depends on D. You only have three levels of node, so this would produce:

A ~> B
A ~> C
B ~> D
C ~> D

But not
B ~> C

I'm sure I've missed something.

adamnfish: Rather than choosing "three levels" arbitrarily, wouldn't it be better to follow the dependencies, recursively building a graph as you go.

magenta-lib/src/test/scala/magenta/input/resolver/DeploymentTasksGraphBuilderTest.scala
```diff
@@ -0,0 +1,62 @@
...
+    val deployment3 = deployment.copy(name="bob3", dependencies = List("bob"))
+    val graph = DeploymentGraphBuilder.buildGraph(List(deployment, deployment2, deployment3))
+    graph shouldBe Graph(
+      StartNode ~> MidNode(deployment),
+      MidNode(deployment) ~> MidNode(deployment2), MidNode(deployment2) ~> EndNode,
+      MidNode(deployment) ~2~> MidNode(deployment3), MidNode(deployment3) ~> EndNode
+    )
+  }
+
+  it should "build a simple graph when there are missing dependencies" in {
```
adamnfish: Is this because it needs to behave this way if a user deselected parts of the deployment during preview?

magenta-lib/src/test/scala/magenta/input/resolver/DeploymentTasksGraphBuilderTest.scala
```diff
@@ -0,0 +1,62 @@
+package magenta.input.resolver
+
+import magenta.graph.{EndNode, Graph, MidNode, StartNode}
+import magenta.input.Deployment
+import org.scalatest.{FlatSpec, ShouldMatchers}
+
+class DeploymentTasksGraphBuilderTest extends FlatSpec with ShouldMatchers {
```
adamnfish: I think this needs a test for when there is a long chain of dependencies (see my comment above about the implementation of the graph builder).

magenta-lib/src/main/scala/magenta/graph/Graph.scala
```diff
@@ -66,6 +96,50 @@ case class Graph[T](edges: Set[Edge[T]]) {
...
+      val fromGraph = nodeToGraphMap(oldEdge.from)
+      val toGraph = nodeToGraphMap(oldEdge.to)
+      // get the joining edges
+      val newEdges = Graph.joiningEdges(fromGraph, toGraph)
+        // deduplicate any edges that we've already generated
+        .filterNot(e1 => acc.exists(e1 ~= _))
+        // fix up the priorities
+        .map{ newEdge =>
+          val maxPriority = acc.filter(_.from == newEdge.from).map(_.priority).reduceOption(Math.max).getOrElse(0)
+          newEdge.copy(priority = newEdge.priority + maxPriority)
```
adamnfish: Why doesn't this make use of the existing reprioritise stuff?

It's not clear from "fix up the priorities" what you expect this to do when merging trees with clashing priorities?

magenta-lib/src/main/scala/magenta/graph/Graph.scala
```diff
@@ -66,6 +96,50 @@ case class Graph[T](edges: Set[Edge[T]]) {
...
+    *
+    * An identity mapping for flatMap would be start and end nodes to empty map and mid nodes to a single value graph
+    * containing the value of the node.
+    *
+    * It is not possible to map a mid node to an empty node.
+    * @param f the function to apply to each element
+    * @return a new graph with the graphs substituted
+    */
+  def flatMap[R](f: Node[T] => Graph[R]): Graph[R] = {
+    val nodeToGraphMap: Map[Node[T], Graph[R]] = nodes.map(n => n -> f(n)).toMap
```
adamnfish: Is the fact that you're using this here a sign that it would be a better representation than `Set[Edge]`?

magenta-lib/src/main/scala/magenta/graph/Graph.scala
sihil: Yep - that's fair enough. I don't think it ever will, but it would be clearer to abstract that out.

magenta-lib/src/main/scala/magenta/graph/Graph.scala
sihil: That should read "empty graph". There is a test that does exactly this.

magenta-lib/src/main/scala/magenta/graph/Graph.scala
sihil: Too hard. If you think it's reasonably easy to do then happy to chat about it.

magenta-lib/src/main/scala/magenta/graph/Graph.scala
sihil: `Graph[R]`? You have to have the context of a start and end node as that is where we join in the incoming and outgoing edges for the existing node.

magenta-lib/src/main/scala/magenta/graph/Graph.scala
sihil: Yes!

magenta-lib/src/main/scala/magenta/graph/Graph.scala
sihil: Yeah - try to actually do that though. Maybe it's easier than I think but I chose to do this because it's significantly simpler not to be able to remove nodes.

magenta-lib/src/main/scala/magenta/graph/Graph.scala
sihil: Reprioritise assumes we have all of the edges to reprioritise. In this case we only have the new edges that are the result of processing one old edge. Other edges that leave the same node can already have been accumulated - in which case we use the max priority to offset the existing priorities.

It might be possible to reuse the reprioritise call but I think the code would be more complex (we have to sort the sets into lists first).

magenta-lib/src/main/scala/magenta/graph/Graph.scala
sihil: Possibly. This was a big improvement over the existing case class output without the complexity of having to work out how to render a graph with ascii art.

It's useful to have the priority here - there is no reason why you can't have the same priority twice which could result in `toString` output that looks the same but two case classes that fail equality.

magenta-lib/src/main/scala/magenta/input/resolver/DeploymentGraphBuilder.scala
sihil: Yes - you've missed something. The three levels are really:
 - edges that go from start to a node
 - edges that go from a node to the end
 - any other edge that goes from a node to a node

It can happily handle multiple layers of dependencies. It might be possible to do as you say - let's chat IRL.

magenta-lib/src/main/scala/magenta/input/resolver/DeploymentGraphBuilder.scala
sihil: It relates directly to `MidNode`. This creates any edges that go from a `MidNode` to another `MidNode`.

magenta-lib/src/test/scala/magenta/input/resolver/DeploymentTasksGraphBuilderTest.scala
sihil: Yes - exactly. This should never happen unless the deployments are filtered.

magenta-lib/src/test/scala/magenta/input/resolver/DeploymentTasksGraphBuilderTest.scala
sihil: Yes - good idea. I've added a test with five levels of dependencies.

magenta-lib/src/main/scala/magenta/input/resolver/DeploymentGraphBuilder.scala
sihil: It's another approach. You wind up building edges multiple times with that approach as there are potentially many paths through the graph. You have to start at the StartNode and build onwards recursing down. think you might have to pass through an accumulator of visited nodes as well.

I just had a bash and it gets complicated very quickly. Let's chat IRL.

sihil:
@adamnfish checkout the new beautiful graph builder implementation.

sihil:
I think I've made all the changes that you asked for so perhaps you could have another look this afternoon?

magenta-lib/src/main/scala/magenta/input/resolver/DeploymentGraphBuilder.scala
```diff
@@ -0,0 +1,27 @@
...
+
+  private[resolver] def allEdges(deployments: List[Deployment]) = {
+    val deploymentNodes = deployments.map(MidNode.apply)
+    for {
+      from <- StartNode :: deploymentNodes
+      edge <- {
+        val targets = from match {
+          case StartNode => deploymentNodes.filterNot(_.value.dependencies.exists(deployments.map(_.name).contains))
+          case MidNode(deployment) => deploymentNodes.filter(_.value.dependencies.contains(deployment.name))
+          case EndNode => throw new IllegalStateException("EndNode is not valid as the source of an edge")
```
sihil: Yep - I know :( the compiler minds. I considered using `???` but that seemed dirtier.

philwills:
Sorry for the delay replying, but this isn't something which I can review in dribs and drabs. Could I have a chat at some point, as there are a couple of things I'm unclear on.

sihil:
On @philwills request I've fully documented the Graph class. Feedback welcome although part of me would prefer others to contribute if it's not clear.

Whilst documenting I realised that `MidNode` was a stupid name and have switched it out for `ValueNode`.

PR #354, opened 2016-09-23
Switch input file parsing to consistently use Play Json

sihil:
Remove Json4s from the codebase so we more consistently use Play Json.

This aligns the types used between the two different parameter systems and eliminates more from the codebase. 

The Play Json parser is stricter than whatever Json4s uses by default as it no longer allows trailing commas. These were used liberally in the test classes but I haven't found an example in real life.

In addition I've removed the magic extractor that converted numbers represented as Json strings back into numbers. This used to be for `port` but I'm pretty sure we don't need it anymore. Might be worth some comms when we ship this.


magenta-lib/src/main/scala/magenta/deployment_type/S3.scala
```diff
@@ -1,11 +1,37 @@
...
+import play.api.data.validation.ValidationError
+import play.api.libs.json._
+
+case class PatternValue(pattern: String, value: String) {
+  lazy val regex = pattern.r
+}
+object PatternValue {
+  implicit val reads = Json.reads[PatternValue]
+
+  implicit val patternValueReads = new Reads[List[PatternValue]] {
```
philwills: It feels like there's enough going on here that a dedicated test would be worth it.

sihil: Sounds like a plan. I'll have a look at doing that.

Are you happy with removing the JValueExtractor stuff entirely? Felt like it duplicated `Reads`.

philwills: Yea.  I'm happy to get rid.

sihil: Great. Some happy and unhappy tests added. I've also pulled out the custom reads class into a separate package.

philwills:
👍 👍 👍

PR #355, opened 2016-09-27
Report fastly validation response

mchv:
Report fastly validation response so a user could self understand the issue if not the one expected.


mchv:
Closing as we are working on a better fix with @sihil

PR #356, opened 2016-09-27
Fix JSON parsing of fastly response and add unhappy case warning

sihil:
#354 didn't correctly fix up the Fastly status check. This fixes the error and also adds a warning check.


mchv:
👍

PR #357, opened 2016-10-03
Regions as first class concept

sihil:
I'm afraid this bundles a handful of changes together (be glad I stopped here). I've tried to separate it into two commits to aid reviewing - but I've not managed to separate them cleanly enough to make it worth splitting this PR.

The first commit is concerned with:
 - More consistent use of DeploymentResources
 - `S3Artifact` is split into `S3YamlArtifact` and `S3JsonArtifact`
 - `S3Package` is eliminated

The second commit is all about making Regions a first class concept within Riff-Raff which is required now that regions can be configured in the new config file.

### DIFF
--- a/magenta-lib/src/main/scala/magenta/Project.scala
+++ b/magenta-lib/src/main/scala/magenta/Project.scala
@@ -72,7 +72,7 @@ case class DeploymentResources(reporter: DeployReporter, lookup: Lookup, artifac
     lookup.keyRing(target.parameters.stage, pkg.apps.toSet, target.stack)
 }
 
-case class DeployTarget(parameters: DeployParameters, stack: Stack)
+case class DeployTarget(parameters: DeployParameters, stack: Stack, region: Region)
 
 /*
  An action represents a step within a recipe. It isn't executable
@@ -117,6 +117,8 @@ case object UnnamedStack extends Stack {
   override def nameOption = None
 }
 
+case class Region(name: String) extends AnyVal
+
 case class Deployer(name: String)
 
 case class DeployParameters(
@@ -128,8 +130,8 @@ case class DeployParameters(
                              hostList: List[String] = Nil
                              ) {
 
-  def toDeployContext(uuid: UUID, project: Project, resources: DeploymentResources): DeployContext = {
-    DeployContext(uuid, this, project, resources)
+  def toDeployContext(uuid: UUID, project: Project, resources: DeploymentResources, region: Region): DeployContext = {
+    DeployContext(uuid, this, project, resources, region)
   }
 
   def matchingHost(hostName:String) = hostList.isEmpty || hostList.contains(hostName)
--- a/magenta-lib/src/main/scala/magenta/Project.scala
+++ b/magenta-lib/src/main/scala/magenta/Project.scala
@@ -128,11 +128,4 @@ case class DeployParameters(
                              recipe: RecipeName = DefaultRecipe(),
                              stacks: Seq[NamedStack] = Seq(),
                              hostList: List[String] = Nil
-                             ) {
-
-  def toDeployContext(uuid: UUID, project: Project, resources: DeploymentResources, region: Region): DeployContext = {
-    DeployContext(uuid, this, project, resources, region)
-  }
-
-  def matchingHost(hostName:String) = hostList.isEmpty || hostList.contains(hostName)
-}
+                             )
--- a/magenta-lib/src/main/scala/magenta/Project.scala
+++ b/magenta-lib/src/main/scala/magenta/Project.scala
@@ -128,8 +128,8 @@ case class DeployParameters(
                              hostList: List[String] = Nil
                              ) {
 
-  def toDeployContext(uuid: UUID, project: Project, resourceLookup: Lookup, reporter: DeployReporter, artifactClient: AmazonS3): DeployContext = {
-    DeployContext(uuid, this, project, resourceLookup, reporter, artifactClient)
+  def toDeployContext(uuid: UUID, project: Project, resources: DeploymentResources): DeployContext = {
+    DeployContext(uuid, this, project, resources)
   }
 
   def matchingHost(hostName:String) = hostList.isEmpty || hostList.contains(hostName)
--- a/magenta-lib/src/main/scala/magenta/deployment_type/AutoScaling.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/AutoScaling.scala
@@ -92,17 +92,17 @@ object AutoScaling  extends DeploymentType with S3AclParams {
       val parameters = target.parameters
       val stack = target.stack
       List(
-        CheckForStabilization(pkg, parameters.stage, stack),
-        CheckGroupSize(pkg, parameters.stage, stack),
-        SuspendAlarmNotifications(pkg, parameters.stage, stack),
-        TagCurrentInstancesWithTerminationTag(pkg, parameters.stage, stack),
-        DoubleSize(pkg, parameters.stage, stack),
+        CheckForStabilization(pkg, parameters.stage, stack, target.region),
+        CheckGroupSize(pkg, parameters.stage, stack, target.region),
+        SuspendAlarmNotifications(pkg, parameters.stage, stack, target.region),
+        TagCurrentInstancesWithTerminationTag(pkg, parameters.stage, stack, target.region),
+        DoubleSize(pkg, parameters.stage, stack, target.region),
         HealthcheckGrace(healthcheckGrace(pkg) * 1000),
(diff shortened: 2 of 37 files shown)


magenta-lib/src/main/scala/magenta/tasks/AWS.scala
```diff
@@ -47,28 +43,23 @@ trait Lambda extends AWS {
...
-    val client = new AmazonAutoScalingClient(provider(keyRing), clientConfiguration)
-    client.setEndpoint("autoscaling.eu-west-1.amazonaws.com")
-    client
-  }
+object ASG extends AWS {
+  def makeAsgClient(keyRing: KeyRing, region: Region): AmazonAutoScalingClient =
+    new AmazonAutoScalingClient(provider(keyRing), clientConfiguration).withRegion(awsRegion(region))
 
-  def desiredCapacity(name: String, capacity: Int)(implicit keyRing: KeyRing) =
+  def desiredCapacity(name: String, capacity: Int)(implicit client: AmazonAutoScalingClient) =
```
adamnfish: As a vague rule of thumb I use implicit args when I'm making a library and I want it to be easy for a third party to call. I use real args internally because they are clearer, unless it's something where there's a clear benefit. In these cases I'm not sure that writing `(implicit client: AmazonAutoScalingClient)` everywhere is so much better than writing `, client)` everywhere instead. No big deal.

magenta-lib/src/test/scala/magenta/deployment_type/AutoScalingTest.scala
```diff
@@ -24,18 +26,18 @@ class AutoScalingTest extends FlatSpec with Matchers {
...
     val p = DeploymentPackage("app", app, data, "asg-elb", S3Path("artifact-bucket", "test/123/app"))
 
-    AutoScaling.actions("deploy")(p)(DeploymentResources(reporter, lookupEmpty, artifactClient), DeployTarget(parameters(), UnnamedStack)) should be (List(
-      CheckForStabilization(p, PROD, UnnamedStack),
-      CheckGroupSize(p, PROD, UnnamedStack),
-      SuspendAlarmNotifications(p, PROD, UnnamedStack),
-      TagCurrentInstancesWithTerminationTag(p, PROD, UnnamedStack),
-      DoubleSize(p, PROD, UnnamedStack),
+    AutoScaling.actions("deploy")(p)(DeploymentResources(reporter, lookupEmpty, artifactClient), DeployTarget(parameters(), UnnamedStack, region)) should be (List(
+      CheckForStabilization(p, PROD, UnnamedStack, Region("eu-west-1")),
```
adamnfish: There's a `val` for that.

riff-raff/app/conf/context.scala
```diff
@@ -159,8 +163,16 @@ class Configuration(val application: String, val webappConfDirectory: String = "
...
       lazy val credentialsProvider = credentialsProviderChain(accessKey, secretKey)
+      lazy val regionName = configuration.getStringProperty("tag.aws.region", "eu-west-1")
+      lazy val region = awsRegion(regionName)
+      implicit lazy val client = new AmazonS3Client(credentialsProvider).withRegion(region)
+    }
+  }
+
+  object target {
+    object aws {
+      lazy val defaultRegionName = configuration.getStringProperty("target.aws.defaultRegion", "eu-west-1")
```
adamnfish: Do you need to fall back to a default here? Why not just make it required to reduce confusion when someone sets this up in ~~redacted~~^H^H^H "a new region" next year.

magenta-lib/src/main/scala/magenta/deployment_type/AutoScaling.scala
```diff
@@ -123,6 +123,7 @@ object AutoScaling  extends DeploymentType with S3AclParams {
         )
       List(
         S3Upload(
+          target.region,
```
sihil: That's an interesting philosophical question. This would mean that you get multiple uploads at present. The config lets you avoid this as you can override the region - but you are right, it's probably not helpful.

magenta-lib/src/main/scala/magenta/deployment_type/Lambda.scala
```diff
@@ -76,19 +74,19 @@ object Lambda extends DeploymentType  {
 
   def defaultActions = List("uploadLambda", "updateLambda")
 
-  def lambdaToProcess(pkg: DeploymentPackage, stage: String, stack: Stack)(reporter: DeployReporter): List[LambdaFunction] = {
+  def lambdaToProcess(pkg: DeploymentPackage, stage: String, stack: Stack, targetRegion: Region)(reporter: DeployReporter): List[LambdaFunction] = {
     val bucketOption = bucketParam.get(pkg)
     if (bucketOption.isEmpty) reporter.warning(s"DEPRECATED: Uploading directly to lambda is deprecated (it is dangerous for CloudFormed lambdas). Specify the bucket parameter and call both uploadLambda and updateLambda to upload via S3.")
     
-    val regions: List[Region] = regionsParam(pkg).flatMap(name => toRegion(name)(reporter))
+    val regions: List[Region] = regionsParam.get(pkg).map(_.filter(regionExists(_)(reporter)).map(Region)).getOrElse(List(targetRegion))
```
sihil: I was retaining the existing behaviour, but actually changing it makes a lot of sense.

riff-raff/app/conf/context.scala
sihil: There is no field for region in deploy.json, but open to making it mandatory in riff-raff.yaml.

magenta-lib/src/main/scala/magenta/Project.scala
```diff
@@ -128,8 +128,8 @@ case class DeployParameters(
                              hostList: List[String] = Nil
                              ) {
 
-  def toDeployContext(uuid: UUID, project: Project, resourceLookup: Lookup, reporter: DeployReporter, artifactClient: AmazonS3): DeployContext = {
-    DeployContext(uuid, this, project, resourceLookup, reporter, artifactClient)
+  def toDeployContext(uuid: UUID, project: Project, resources: DeploymentResources): DeployContext = {
```
sihil: Removed

riff-raff/app/deployment/actors/DeployGroupRunner.scala
```diff
@@ -154,14 +156,35 @@ class DeployGroupRunner(
...
+
+      val context: DeployContext = riffRaffYamlString.map { yaml =>
+//        val graph = Resolver.resolve(yaml, resources, record.parameters, DeploymentType.all, riffRaffYaml)
+//        graph.right.map(DeployContext(record.uuid, record.parameters, _)) match {
+//          case Left(errors) =>
+//            errors.errors.foreach(error => safeReporter.info(s"${error.context}: ${error.message}"))
+//            safeReporter.fail(s"Failed to successfully resolve the deployment: ${errors.errors.size} errors")
+//          case Right(success) => success
+//        }
+        ???
```
sihil: Changed this for a clearer `fail` for now.

magenta-lib/src/test/scala/magenta/deployment_type/AutoScalingTest.scala
sihil: I made it explicit as it is a test.

magenta-lib/src/main/scala/magenta/tasks/AWS.scala
sihil: Phil had similar concerns. Removed use of implicits.

sihil:
And yes - `S3Path` already existed - so this was simply getting rid of a case class that wasn't really adding much.

magenta-lib/src/main/scala/magenta/Project.scala
```diff
@@ -117,6 +117,8 @@ case object UnnamedStack extends Stack {
   override def nameOption = None
 }
 
+case class Region(name: String) extends AnyVal
```
philwills: What's the reason not to use `com.amazonaws.regions.Region`?

magenta-lib/src/main/scala/magenta/deployment_type/AutoScaling.scala
philwills: I can see arguments both ways. If you're trying to be across multiple regions for resiliency, then having the artifact in both regions seems like a good thing.

magenta-lib/src/main/scala/magenta/Project.scala
sihil: Mainly having a sensible `toString` method given that the region will be inside tasks. I did start off using the AWS classes but found, for example, that testing was harder as the AWS region will not so easily pattern match.

PR #359, opened 2016-10-04
Tie together all of the loose ends

sihil:
- Add a resolver to build a DeploymentTasks from a Deployment
 - Complete the top level Resolver
 - Facilitate different defaults for deploy.json / riff-raff.yaml (including documentation)

### DIFF
--- a/magenta-lib/src/main/scala/magenta/input/resolver/Resolver.scala
+++ b/magenta-lib/src/main/scala/magenta/input/resolver/Resolver.scala
@@ -52,10 +52,10 @@ object Resolver {
   }
 
   private[resolver] def filterDeployments(deployments: List[Deployment], stack: String, region: String): List[Deployment] = {
-    deployments.flatMap {
-      case d@Deployment(_, _, stacks, regions, _, _, _, _, _) if stacks.contains(stack) && regions.contains(region) =>
-        Some(d.copy(stacks = List(stack), regions = List(region)))
-      case _ => None
+    deployments.flatMap { deployment =>
+      if (deployment.stacks.contains(stack) && deployment.regions.contains(region))
+        Some(deployment.copy(stacks = List(stack), regions = List(region)))
+      else None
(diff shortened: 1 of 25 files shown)


magenta-lib/src/main/scala/magenta/input/resolver/Resolver.scala
```diff
@@ -1,34 +1,61 @@
...
+        Right(deploymentTaskGraph.map(_.right.get))
+      }
+    }
+  }
 
-    // now resolve each deployment into tasks
-    //val deploymentTaskGraph = DeploymentTaskResolver.resolveGraph(flattenedGraph)
+  private[resolver] def filterDeployments(deployments: List[Deployment], stack: String, region: String): List[Deployment] = {
+    deployments.flatMap {
+      case d@Deployment(_, _, stacks, regions, _, _, _, _, _) if stacks.contains(stack) && regions.contains(region) =>
```
philwills: That is one ugly match `case`

magenta-lib/src/main/scala/magenta/input/resolver/TaskResolver.scala
```diff
@@ -0,0 +1,44 @@
...
+    DeploymentPackage(
+      name = deployment.name,
+      pkgApps = Seq(App(deployment.app)),
+      pkgSpecificData = deployment.parameters,
+      deploymentTypeName = deployment.`type`,
+      s3Package = S3Path(artifact, deployment.contentDirectory)
+    )
+  }
+
+  private def mkLabel(name: String, actions: List[String], regions: List[String], stacks: List[String]): String = {
```
philwills: It feels strange that this printing logic is here, but I haven't got an immediate great solution for an alternative.

sihil: Agreed. On both counts.

Issue #335, opened 2016-09-09
Apply a code formatter


cb372:
The code is a royal mess.

sbt-scalariform is a classic, but I hear [scalafmt](https://olafurpg.github.io/scalafmt/) is where the cool kids are at these days.

philwills:
I have a particular personal animus towards code formatters. They always stop me doing things I explicitly want to do and historically I've had issues with them resulting in SBT/IntelliJ churn.

Things may be much improved now, but I've never been able to last more than a week with one before getting frustrated and ripping it out.

cb372:
Even if you're a philistine who hates nice things, could we at least run a formatter over the code *once*? After that you're free to mess it all up again.

philwills:
Yea, that sounds a reasonable compromise.

philwills:
See also #434

AWare:
I don't think this is likely to happen anytime soon, and we've lost enough people that the history is more valuable than when we last decided the history was too valuable to lose.

PR #365, opened 2016-10-05
Install scalafmt and reformat all source

sihil:
Tackle #335.

Whilst I've installed scalafmt as a plugin it is not wired up to run automatically as my reading of #335 indicates that @philwills would be disappointed if this were the case.

I'm happy to remove the plugin entirely if that is deemed better.


philwills:
I'm fairly convinced that a fair chunk of this is less readable, but I'm willing to be overruled.

magenta-lib/src/main/scala/magenta/deployment_type/AmiCloudFormationParameter.scala
```diff
@@ -11,44 +11,48 @@ object AmiCloudFormationParameter extends DeploymentType {
...
-  val amiTags = Param[Map[String,String]]("amiTags",
-    documentation = "Specify the set of tags to use to find the latest AMI"
-  )
-  val amiParameter = Param[String]("amiParameter",
-    documentation = "The CloudFormation parameter name for the AMI"
-  ).default("AMI")
+  val cloudFormationStackName =
+    Param[String]("cloudFormationStackName", documentation = "The name of the CloudFormation stack to update")
+      .defaultFromPackage(_.name)
+  val prependStackToCloudFormationStackName = Param[Boolean](
```
philwills: I really don't believe 5 lines is better than 3 here.

sihil: Fair point.

sihil:
I'm happy to close this, but thought I'd give it a bash in a moment of enthusiasm.

philwills:
Maybe it's worth trying this and then going back an fixing up things we don't like afterwards.  It's always easy to focus on the bad things in a change rather than the good and it may well be that the balance is in favour of this.

PR #366, opened 2016-10-05
I must never use a default parameter for a critical flag

sihil:
The field on this case class was never set in the case where it _wasn't_ a `legacyConfig`. Thus everything was treated as if it was a legacy configuration. This is a lesson to not use a default parameter in order to avoid updating all of the tests.


adamnfish:
:+1: and :+1: with a security hat on for the lesson ;-)

PR #367, opened 2016-10-05
Numerous parameter validation improvements

sihil:
In the new riff-raff.yaml world we:
 - Error fast when a required parameter is missing
 - Error when a parameter is supplied that is not used by the deployment type
 - Log a warning when a parameter is explicitly set to the default value of the parameter

### DIFF
--- a/magenta-lib/src/main/scala/magenta/deployment_type/DeploymentType.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/DeploymentType.scala
@@ -53,12 +53,18 @@ case class Param[T](name: String,
 
   def get(pkg: DeploymentPackage)(implicit reads: Reads[T]): Option[T] =
     pkg.pkgSpecificData.get(name).flatMap(jsValue => Json.fromJson[T](jsValue).asOpt)
-  def apply(pkg: DeploymentPackage)(implicit reads: Reads[T], manifest: Manifest[T]): T =
-    get(pkg).orElse(defaultValue).orElse(defaultValueFromPackage.map(_(pkg))).getOrElse{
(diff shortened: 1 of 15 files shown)


magenta-lib/src/main/scala/magenta/deployment_type/DeploymentType.scala
```diff
@@ -53,12 +53,18 @@ case class Param[T](name: String,
 
   def get(pkg: DeploymentPackage)(implicit reads: Reads[T]): Option[T] =
     pkg.pkgSpecificData.get(name).flatMap(jsValue => Json.fromJson[T](jsValue).asOpt)
-  def apply(pkg: DeploymentPackage)(implicit reads: Reads[T], manifest: Manifest[T]): T =
-    get(pkg).orElse(defaultValue).orElse(defaultValueFromPackage.map(_(pkg))).getOrElse{
+  def apply(pkg: DeploymentPackage)(implicit reporter: DeployReporter, reads: Reads[T], manifest: Manifest[T]): T = {
```
philwills: I know I'm a broken record on this, but why make this implicit?

sihil: No strong reason, only it's called in a lot of places and it makes the code less expressive. It's a shame that the reporter is not already available here.

philwills: To me the noise of `implicit val reporter = resources.reporter` in lots of places hurts expressiveness more than passing it through as a parameter.

sihil: Missed this (only just refreshed my page). Have removed the implicit.

PR #370, opened 2016-10-07
When parsing the YAML check for extra fields

sihil:
If a field is provided that we are not using then this is probably a bad thing. In many cases it will be because someone has made a typo (e.g. they may have used `template` instead of `templates`). This is another opportunity to fail fast to make life easier for users.

### DIFF
--- a/magenta-lib/src/main/scala/magenta/input/RiffRaffYamlReader.scala
+++ b/magenta-lib/src/main/scala/magenta/input/RiffRaffYamlReader.scala
@@ -43,7 +43,8 @@ object RiffRaffYamlReader {
       case JsError(errors :: tail) =>
         val nelErrors = NEL(errors, tail)
         Invalid(ConfigErrors(nelErrors.map{ case (path, validationErrors) =>
-          ConfigError(s"Parsing $path", validationErrors.map(ve => ve.message).mkString(", "))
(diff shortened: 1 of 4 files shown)


magenta-lib/src/main/scala/magenta/deployment_type/DeploymentType.scala
```diff
@@ -53,7 +53,7 @@ case class Param[T](name: String,
 
   def get(pkg: DeploymentPackage)(implicit reads: Reads[T]): Option[T] =
     pkg.pkgSpecificData.get(name).flatMap(jsValue => Json.fromJson[T](jsValue).asOpt)
-  def apply(pkg: DeploymentPackage, reporter: DeployReporter)(implicit reads: Reads[T], manifest: Manifest[T]): T = {
+  def apply(pkg: DeploymentPackage, reporter: DeployReporter)(implicit reads: Reads[T]): T = {
```
sihil: Sneaking the removal of this unused implicit into this (noticed it whilst looking up reflection)

philwills:
Seems like an excellent idea to me.

sihil:
Is that a +1 @philwills? I'm also tempted to import https://github.com/rockymadden/stringmetric and suggest corrections :)

philwills:
👍 👍 👍 

stringmetric does look cool

PR #371, opened 2016-10-10
Enable desktop notification upon deployment

alexduf:
The user will be prompted to allow or reject notifications on the first deployment. Users can disable notifications later by modifying their browser authorisations.

The appearance will vary depending on your system.

screenshot from 2016-10-10 18-13-33

### DIFF
--- a/riff-raff/app/assets/javascripts/notifications.coffee
+++ b/riff-raff/app/assets/javascripts/notifications.coffee
@@ -1,7 +1,4 @@
-$ ->
-  $("#enable-notifications").click (e) ->
-    e.preventDefault()
-    Notification.requestPermission()
+Notification.requestPermission()
 
 intervalId = null
 
@@ -14,7 +11,7 @@ checkStatus = () ->
     disableCheck()
 
 if $('[data-run-state]').length == 0
-  intervalId = setInterval checkStatus, 1000
+  intervalId = setInterval checkStatus, 800
 
 disableCheck = ->
   clearInterval(intervalId) if intervalId?
--- a/riff-raff/app/assets/javascripts/notifications.coffee
+++ b/riff-raff/app/assets/javascripts/notifications.coffee
@@ -6,8 +6,8 @@ checkStatus = () ->
   if $('[data-run-state]').length != 0
     buildName = window.riffraff.buildName
     switch $('[data-run-state]').data('run-state')
-      when 'Failed' then new Notification('Riffraff', {body: 'Deployment of ' + buildName + ' failed!'})
-      when 'Completed' then new Notification('Riffraff', {body: 'Deployment of ' + buildName + ' is finished'})
+      when 'Failed' then new Notification('Riffraff', {body: 'Deployment of ' + buildName + ' has failed!'})
+      when 'Completed' then new Notification('Riffraff', {body: 'Deployment of ' + buildName + ' has finished'})
     disableCheck()
 
 if $('[data-run-state]').length == 0
--- a/riff-raff/app/assets/javascripts/notifications.coffee
+++ b/riff-raff/app/assets/javascripts/notifications.coffee
@@ -1,16 +1,17 @@
-Notification.requestPermission()
-
 intervalId = null
 
 checkStatus = () ->
   if $('[data-run-state]').length != 0
     buildName = window.riffraff.buildName
+    stage = window.riffraff.stage
+    buildId = window.riffraff.buildId
     switch $('[data-run-state]').data('run-state')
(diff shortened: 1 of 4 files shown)


alexduf:
The implementation is... 🙄 not beautiful, but it should do the job. Let me know if there's something I should fix

riff-raff/app/assets/javascripts/notifications.coffee
```diff
@@ -0,0 +1,17 @@
...
+checkStatus = () ->
+  if $('[data-run-state]').length != 0
+    buildName = window.riffraff.buildName
+    switch $('[data-run-state]').data('run-state')
+      when 'Failed' then new Notification('Riffraff', {body: 'Deployment of ' + buildName + ' failed!'})
+      when 'Completed' then new Notification('Riffraff', {body: 'Deployment of ' + buildName + ' is finished'})
+    disableCheck()
+
+if $('[data-run-state]').length == 0
+  intervalId = setInterval checkStatus, 800
```
sihil: 800? Is that an arbitrary choice?

riff-raff/app/views/deploy/logSummary.scala.html
```diff
@@ -5,10 +5,10 @@
     }
     @defining(record.state) { state =>
         @state match {
-            case RunState.NotRunning => { <div class="alert alert-info"><p class="lead">Waiting to run</p></div> }
-            case RunState.Completed => { <div class="alert alert-success"><p class="lead">Completed</p></div> }
+            case RunState.NotRunning => { <div class="alert alert-info" data-run-state="NotRunning"><p class="lead">Waiting to run</p></div> }
```
sihil: Won't the fact that the run-state is non-zero length disable the check whilst a deploy is waiting to run?

riff-raff/app/assets/javascripts/notifications.coffee
```diff
@@ -0,0 +1,17 @@
...
+
+intervalId = null
+
+checkStatus = () ->
+  if $('[data-run-state]').length != 0
+    buildName = window.riffraff.buildName
+    switch $('[data-run-state]').data('run-state')
+      when 'Failed' then new Notification('Riffraff', {body: 'Deployment of ' + buildName + ' failed!'})
+      when 'Completed' then new Notification('Riffraff', {body: 'Deployment of ' + buildName + ' is finished'})
+    disableCheck()
```
sihil: Since you set data-run-state on NotRunning this will disable the check immediately when a deploy is waiting to run.

riff-raff/app/views/deploy/viewDeploy.scala.html
```diff
@@ -1,12 +1,17 @@
...
-@main(s"Deploy for ${record.buildName}", request, List("auto-refresh", "deploy-verbose", "stop-deploy")) {
+@main(s"Deploy for ${record.buildName}", request, List("auto-refresh", "deploy-verbose", "stop-deploy", "notifications")) {
     <style>
     span.message-verbose {
       display:@if(!verbose) {none} else {list-item};
     }
     </style>
+    <script language="JavaScript">
+        window.riffraff = {
+            buildName: '@record.buildName'
```
sihil: Add the build ID and stage as well??

riff-raff/app/assets/javascripts/notifications.coffee
alexduf: yes I want it to be out of sync from the main refresh loop (1s) otherwise we can have up to 2s of latency. (I admit it's a detail, but hey why not)

riff-raff/app/views/deploy/logSummary.scala.html
alexduf: You're right I'll change the logic

riff-raff/app/assets/javascripts/notifications.coffee
sihil: Rather than set your own timer you might be able to use `postRefresh`. See https://github.com/guardian/riff-raff/blob/master/riff-raff/app/assets/javascripts/stop-deploy.coffee#L11-L14

riff-raff/app/views/deploy/viewDeploy.scala.html
sihil: Rather than set a global variable you could set another data attribute on the element.

alexduf:
@sihil I think I've covered all the above points, except the global state. Do you think the global state is a blocker? I think it's useful to have a way to get the deployment state without relying on the DOM

sihil:
I'm still not sure you're handling state appropriately. The NotRunning state still polls as at some stage it will go into the Running state.

I'm in the process of re-writing all of the client side in ScalaJS and react anyway so I wouldn't worry.

PR #373, opened 2016-10-12
Further improve parameter validation

sihil:
I had been overly enthusiastic about parameter validation and it turned out that there are some cases in which no default is specified and that's still OK (i.e. the parameter has no default but is optional). One of these cases is when there are two parameters and you need to set one and only one of these.

This adds an optional field and also enhances the withDefault functionality to provide the DeployTarget instance - this means that we can shift more default logic into the parameters and make fewer params optional when really they default from stack or some such deploy time data.

### DIFF
--- a/magenta-lib/src/main/scala/magenta/deployment_type/DeploymentType.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/DeploymentType.scala
@@ -55,7 +55,7 @@ trait ParamRegister {
   *                       might be needed if the value is not required to have a default or when only one of two
   *                       different parameters are specified.
   * @param defaultValue The default value for this parameter - used when a value is not found in the map
-  * @param defaultValueFromContext A function that returns a defualt value for this parameter based on the package for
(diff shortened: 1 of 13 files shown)


magenta-lib/src/main/scala/magenta/deployment_type/DeploymentType.scala
```diff
@@ -45,31 +45,61 @@ trait ParamRegister {
...
+  * A parameter for a deployment type
+ *
+  * @param name The name of the parameter that should be extracted from the parameter map (riff-raff.yaml) or data map
+  *             (deploy.json)
+  * @param documentation A documentation string (in markdown) that describes this parameter
+  * @param optionalInYaml This can be set to true to make the parameter optional even when there are no defaults. This
+  *                       might be needed if the value is not required to have a default or when only one of two
+  *                       different parameters are specified.
+  * @param defaultValue The default value for this parameter - used when a value is not found in the map
+  * @param defaultValueFromContext A function that returns a defualt value for this parameter based on the package for
```
TBonnin: `defualt` 😁

PR #375, opened 2016-10-13
Slightly better validation UI for demoing

sihil:
This provides a better validation UI for my riff-raff.yaml demo this Friday.

### DIFF
--- a/riff-raff/app/views/validation/validationErrors.scala.html
+++ b/riff-raff/app/views/validation/validationErrors.scala.html
@@ -10,8 +10,8 @@
             </div>
             <div class="panel-body">
                 <ul>
-                @errors.map(_.message).map { msg =>
-                  <li>@msg</li>
+                @errors.map { error =>
+                  <li>@error.message</li>
                 }
                 </ul>
             </div>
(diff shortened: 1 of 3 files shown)


riff-raff/app/views/validation/validationErrors.scala.html
```diff
@@ -3,8 +3,19 @@
...
+    @for((context, errors) <- errors.errors.toList.groupBy(_.context)) {
+        <div class="panel panel-danger">
+            <div class="panel-heading">
+                <h3 class="panel-title"><span class="glyphicon glyphicon-exclamation-sign" aria-hidden="true"></span><span>Error in @context</span></h3>
+            </div>
+            <div class="panel-body">
+                <ul>
+                @errors.map(_.message).map { msg =>
+                  <li>@msg</li>
+                }
```
TBonnin: would the following be the same
```
@errors.map { error =>
    <li>@error.message</li>
}
```

philwills:
I think @TBonnin is right, but otherwise this looks like sensible improvements.

riff-raff/app/views/validation/validationErrors.scala.html
sihil: Indeed. This is clearly the kind of thing that happens when refactoring code at midnight.

PR #376, opened 2016-10-14
Add some debug and warnings around the S3 upload task

sihil:
This is to help with tracking down why some static files don't seem to be uploading correctly.


TBonnin:
👍

PR #377, opened 2016-10-18
Make the reporting of a default parameter more functional

sihil:
The code was doing a get on the wrong variable. Regardless I've modified the code so that there is no need to do a get by switching the nested `if` to pattern matching.

Will write some tests later today after my meetings and @adamnfish has shipped Janus.


adamnfish:
:+1: well fixed!

PR #378, opened 2016-10-18
Move Param classes into own file and write some tests

sihil:
As promised in #377 here are some tests for the `Param` class.

I've also moved `Param` into a separate file.


philwills:
⛱ 🌤 📚

PR #379, opened 2016-10-21
Eliminate use of DeploymentType.all global

sihil:
Whilst writing the YAML preview pages I realised that eliminating the use of the `DeploymentType.all` global would make the code clearer and (naturally) make it easier to test.

This also ties up a loose end where the code still used `ValidatedNel` instead of `Validated` and the `ConfigErrors` semigroup.

### DIFF
--- a/riff-raff/app/AppComponents.scala
+++ b/riff-raff/app/AppComponents.scala
@@ -13,9 +13,9 @@ import utils.HstsFilter
 import ci.ContinuousDeployment
 import controllers._
 import deployment.{DeploymentEngine, Deployments}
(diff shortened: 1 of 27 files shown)


riff-raff/app/AppComponents.scala
```diff
@@ -13,9 +13,9 @@ import utils.HstsFilter
 import ci.ContinuousDeployment
 import controllers._
 import deployment.{DeploymentEngine, Deployments}
+import magenta.deployment_type.{AmiCloudFormationParameter, CloudFormation, Lambda, SelfDeploy, _}
```
sihil: IntelliJ.

sihil:
It should all become significantly simpler once JSON deploys are decommissioned. Deploy type checking is all done in the resolver in that instance.

PR #380, opened 2016-10-21
Preview functionality for YAML config files

sihil:
This introduces the ability to preview deploys that use the new YAML configuration files.

**Note:** This does not provide any filtering yet.

Previews are still done in the same way from the deploy screen. They will now be sent to the new preview endpoint which will try to get the YAML config file. If it fails then it redirects the user to the old preview endpoint which will then get the JSON config.

Assuming it is a YAML config file then this is what it looks like:
screen shot 2016-10-21 at 18 25 21

Each dependency is an anchor link that jumps straight to the related tasks.


riff-raff/app/deployment/preview/Preview.scala
```diff
@@ -0,0 +1,45 @@
...
+import com.gu.management.Loggable
+import magenta.`package`._
+import magenta.artifact.S3YamlArtifact
+import magenta.deployment_type.DeploymentType
+import magenta.graph.{DeploymentTasks, Graph}
+import magenta.input.resolver._
+import magenta.input.{ConfigErrors, Deployment}
+import magenta.{DeployParameters, DeploymentResources}
+
+object Preview extends Loggable {
```
philwills: Having different classes in the same project with the same name seems likely to cause confusion, even in different packages.

riff-raff/app/deployment/preview/Preview.scala
```diff
@@ -0,0 +1,45 @@
...
+      previewGraph <- sequenceGraph {
+        flattenedGraph.map { deployment =>
+          TaskResolver.resolve(deployment, resources, parameters, deploymentTypes, artifact).map(deployment -> _)
+        }
+      }
+    } yield previewGraph
+    Preview(validatedGraph)
+  }
+
+  private[preview] def sequenceGraph[A, E](graph: Graph[Validated[E, A]])(implicit E: Semigroup[E]): Validated[E, Graph[A]] = {
```
philwills: Might it make sense to implement `Traverse` (http://typelevel.org/cats/typeclasses/traverse.html) for `Graph`?

riff-raff/app/deployment/preview/PreviewCoordinator.scala
```diff
@@ -0,0 +1,53 @@
...
+    val artifact = S3YamlArtifact.apply(parameters.build, conf.Configuration.artifact.aws.bucketName)
+    val maybeConfig = artifact.deployObject.fetchContentAsString()(resources.artifactClient)
+
+    maybeConfig match {
+      case Some(config) =>
+        logger.info(s"Got configuration for $previewId - resolving")
+        val previewFuture = Future(Preview(artifact, config, parameters, resources, deploymentTypes))
+        Await.ready(agent.alter{ _ + (previewId -> PreviewResult(previewFuture)) }, 30.second)
+        Right(previewId)
+      case None =>
```
philwills: There's a `toEither` in cats which I think saves a little noise for this kind of thing.

riff-raff/app/deployment/preview/PreviewCoordinator.scala
```diff
@@ -0,0 +1,53 @@
...
+import magenta.{DeployParameters, DeployReporter, DeploymentResources}
+import resources.PrismLookup
+
+import scala.concurrent.{Await, Future}
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.duration._
+
+class PreviewCoordinator(prismLookup: PrismLookup, deploymentTypes: Seq[DeploymentType]) extends Loggable {
+  implicit lazy val system = ActorSystem("preview")
+  val agent = Agent[Map[UUID, PreviewResult]](Map.empty)
```
philwills: Might it be simpler to just use a `ConcurrentHashMap`?

riff-raff/app/deployment/preview/PreviewCoordinator.scala
sihil: I can do a `toRight`? Or cats has a `fromOption` pimped on to the `Either` object? Do you mean one of those or something I've missed?

riff-raff/app/deployment/preview/PreviewCoordinator.scala
sihil: Yes - that's panned out to be simpler.

riff-raff/app/deployment/preview/Preview.scala
sihil: Yes - I have come to that conclusion whilst working on the filtering functionality! I've renamed the controller to `PreviewController` to make it obvious and have prefixed `Legacy` on the front of all the JSON preview classes.

riff-raff/app/deployment/preview/PreviewCoordinator.scala
philwills: Ah, thought it was the former, but it's the latter:

https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/syntax/either.scala#L283

adamnfish:
Is it missing a "sure, go ahead and deploy this" button?

PR #381, opened 2016-10-25
Add restrictions for preventing undesirable deploys

sihil:
On of the features of Frontend's `goo` wrapper around riff-raff is the ability to block deploys. In addition @adamnfish has expressed a desire to lock down the ability to deploy Janus. 

This goes someway to achieve both of these and hopefully provide just enough flexibility that it will be useful more widely for temporarily blocking deploys.

Further features could be added (or moved to issues for future consideration):
- Nothing currently stops someone creating a new continuous deploy for a restricted project that permits continuous deployments
- Link from a deploy to create a restriction
- Indication of restriction as soon as you enter the project on the deploy screen

### DIFF
--- a/riff-raff/app/ci/ContinuousDeployment.scala
+++ b/riff-raff/app/ci/ContinuousDeployment.scala
@@ -1,14 +1,14 @@
 package ci
 
 import controllers.Logging
-import deployment.Deployments
+import deployment.{ContinuousDeploymentRequestSource, Deployments}
 import lifecycle.Lifecycle
 import magenta.{DeployParameters, Deployer, RecipeName, Stage, Build => MagentaBuild}
 import persistence.ContinuousDeploymentConfigRepository.getContinuousDeploymentList
 import rx.lang.scala.{Observable, Subscription}
 import utils.ChangeFreeze
 
-import scala.util.{Failure, Try}
+import scala.util.Try
 import scala.util.control.NonFatal
 
 class ContinuousDeployment(deployments: Deployments) extends Lifecycle with Logging {
@@ -48,7 +48,10 @@ class ContinuousDeployment(deployments: Deployments) extends Lifecycle with Logg
       if (!ChangeFreeze.frozen(params.stage.name)) {
         log.info(s"Triggering deploy of ${params.toString}")
         try {
-          deployments.deploy(params)
+          deployments.deploy(params, requestSource = ContinuousDeploymentRequestSource) match {
+            case Left(error) => log.error(s"Couldn't continuously deploy $params: $error")
+            case Right(_) => // happy case
+          }
         } catch {
           case NonFatal(e) => log.error(s"Could not deploy $params", e)
         }
--- a/riff-raff/app/ci/ContinuousDeployment.scala
+++ b/riff-raff/app/ci/ContinuousDeployment.scala
@@ -48,9 +48,8 @@ class ContinuousDeployment(deployments: Deployments) extends Lifecycle with Logg
       if (!ChangeFreeze.frozen(params.stage.name)) {
         log.info(s"Triggering deploy of ${params.toString}")
         try {
-          deployments.deploy(params, requestSource = ContinuousDeploymentRequestSource) match {
-            case Left(error) => log.error(s"Couldn't continuously deploy $params: $error")
-            case Right(_) => // happy case
+          deployments.deploy(params, requestSource = ContinuousDeploymentRequestSource).left.foreach { error =>
+            log.error(s"Couldn't continuously deploy $params: ${error.message}")
           }
         } catch {
           case NonFatal(e) => log.error(s"Could not deploy $params", e)
--- a/riff-raff/app/controllers/DeployController.scala
+++ b/riff-raff/app/controllers/DeployController.scala
@@ -4,7 +4,7 @@ import java.util.UUID
 
 import ci.{Builds, S3Tag, TagClassification}
 import conf.Configuration
-import deployment.{Deployments, LegacyPreviewController, LegacyPreviewResult}
+import deployment.{Deployments, LegacyPreviewController, LegacyPreviewResult, UserRequestSource}
 import magenta._
 import magenta.artifact._
 import magenta.deployment_type.DeploymentType
@@ -73,7 +73,10 @@ class DeployController(deployments: Deployments, prismLookup: PrismLookup, deplo
               "previewStacks" -> parameters.stacks.flatMap(_.nameOption).mkString(",")
             )
           case "deploy" =>
-            val uuid = deployments.deploy(parameters)
+            val uuid = deployments.deploy(parameters, requestSource = UserRequestSource(request.user)) match {
+              case Right(uuid) => uuid
+              case Left(error) => throw new IllegalStateException(error)
+            }
             Redirect(routes.DeployController.viewUUID(uuid.toString))
           case _ => throw new RuntimeException("Unknown action")
         }
--- a/riff-raff/app/controllers/DeployController.scala
+++ b/riff-raff/app/controllers/DeployController.scala
@@ -7,6 +7,7 @@ import conf.Configuration
 import deployment.{Deployments, LegacyPreviewController, LegacyPreviewResult, UserRequestSource}
 import magenta._
 import magenta.artifact._
+import cats.syntax.either._
 import magenta.deployment_type.DeploymentType
 import org.joda.time.{DateTime, DateTimeZone}
 import org.joda.time.format.DateTimeFormat
@@ -73,9 +74,8 @@ class DeployController(deployments: Deployments, prismLookup: PrismLookup, deplo
               "previewStacks" -> parameters.stacks.flatMap(_.nameOption).mkString(",")
             )
           case "deploy" =>
-            val uuid = deployments.deploy(parameters, requestSource = UserRequestSource(request.user)) match {
-              case Right(uuid) => uuid
-              case Left(error) => throw new IllegalStateException(error)
+            val uuid = deployments.deploy(parameters, requestSource = UserRequestSource(request.user)).valueOr{ error =>
+              throw new IllegalStateException(error.message)
             }
             Redirect(routes.DeployController.viewUUID(uuid.toString))
           case _ => throw new RuntimeException("Unknown action")
--- a/riff-raff/app/controllers/Restrictions.scala
+++ b/riff-raff/app/controllers/Restrictions.scala
@@ -1,19 +1,20 @@
 package controllers
 
 import java.util.UUID
-import java.util.regex.PatternSyntaxException
 
-import com.gu.googleauth.UserIdentity
-import deployment.{RequestSource, UserRequestSource}
+import conf.Configuration.auth
+import deployment.Error
 import org.joda.time.DateTime
 import persistence.RestrictionConfigDynamoRepository
 import play.api.data.Form
 import play.api.data.Forms._
 import play.api.i18n.{I18nSupport, MessagesApi}
 import play.api.libs.ws.WSClient
-import play.api.mvc.{Controller, Result}
+import play.api.mvc.Controller
 import restrictions.{RestrictionChecker, RestrictionConfig, RestrictionForm}
 import utils.Forms.uuid
+
+import scala.util.Try
 
 class Restrictions()(implicit val messagesApi: MessagesApi, val wsClient: WSClient) extends Controller with LoginActions
   with I18nSupport {
@@ -35,13 +36,13 @@ class Restrictions()(implicit val messagesApi: MessagesApi, val wsClient: WSClie
         f.continuousDeployment,  f.note))
     ).verifying(
       "Stage is invalid - should be a valid regular expression or contain no special values",
-      form => try { form.stage.r; true } catch { case e:PatternSyntaxException => false }
+      form => Try(form.stage.r).isSuccess
     )
   )
 
   def list = AuthAction { implicit request =>
     val configs = RestrictionConfigDynamoRepository.getRestrictionList.toList.sortBy(r => r.projectName + r.stage)
-    Ok(views.html.restrictions.list(configs))
+    Ok(views.html.restrictions.list(configs, auth.superusers))
   }
 
   def form = AuthAction { implicit request =>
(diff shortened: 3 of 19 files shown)


riff-raff/app/deployment/Deployments.scala
```diff
@@ -21,25 +22,29 @@ import scala.util.control.NonFatal
...
-    if (enableQueueingSwitch.isSwitchedOff)
-      throw new IllegalStateException("Unable to queue a new deploy; deploys are currently disabled by the %s switch" format enableQueueingSwitch.name)
-
-    val params = if (requestedParams.build.id != "lastSuccessful")
-      requestedParams
+    val restrictionsPreventingDeploy = RestrictionChecker.configsThatPreventDeployment(RestrictionConfigDynamoRepository,
+      requestedParams.build.projectName, requestedParams.stage.name, requestSource)
+    if (restrictionsPreventingDeploy.nonEmpty) {
+      Left(s"Unable to queue deploy as restrictions are currently in place: ${restrictionsPreventingDeploy.map(r => s"${r.fullName}: ${r.note}").mkString("; ")}")
+    } else if (enableQueueingSwitch.isSwitchedOff)
```
adamnfish: :minor_point: Why omit the braces from this `else if` clause?

riff-raff/app/deployment/model.scala
```diff
@@ -3,11 +3,19 @@ package deployment
...
 
 import ci.{Builds, S3Build}
+import com.gu.googleauth.UserIdentity
+import controllers.ApiKey
 import magenta.{DeployParameters, ReportTree, _}
 import org.joda.time.{DateTime, Duration, Interval}
 import org.joda.time.format.PeriodFormatterBuilder
 import utils.VCSInfo
 
+trait RequestSource
```
adamnfish: Should this be sealed?

riff-raff/app/restrictions/RestrictionChecker.scala
```diff
@@ -0,0 +1,49 @@
...
+  /** Method that runs one function if the supplied config is editable by the supplied source and another if it is not
+    * @param currentConfig The config to check - note that when a config is being checked this must be the original
+    *                      config not the updated config
+    * @param identity The identity of the user trying to make the edit
+    * @param ifEditable The function whose result will be returned if we consider the config to be editable
+    * @param ifNotEditable The function whose result will be returned if the config cannot be edited
+    * @tparam T The function return type
+    * @return The result of either ifEditable or ifNotEditable
+    */
+  def editable[T](currentConfig: Option[RestrictionConfig], identity: UserIdentity)
```
adamnfish: Does this really save typing compared to returning a Boolean and using an if statement? If statements also take two blocks - the one to run if it is true and the one to run if it is not.

riff-raff/app/restrictions/RestrictionConfig.scala
```diff
@@ -0,0 +1,32 @@
...
+
+import java.util.UUID
+
+import org.joda.time.DateTime
+
+/**
+  * A configuration that restricts which deploys are allowed
+  * @param id Unique ID for CRUD storage and retrival
+  * @param projectName The project name that this applies to
+  * @param stage The stage that this applies to (this can be a regex, but if no special regex characters are found then
```
adamnfish: Would it be better to have this be a regex type field? A string literal is a subset of a the regex type and this would move the "if no special character" logic to the edge.

riff-raff/app/views/restrictions/form.scala.html
```diff
@@ -0,0 +1,81 @@
...
+
+@main("Create Restriction", request) {
+
+    <h2>Create Restriction</h2>
+    <hr/>
+
+    <div class="row">
+        <div class="col-md-7">
+            @b3.form(action=routes.Restrictions.save) {
+            @CSRF.formField
```
adamnfish: :-)

riff-raff/app/ci/ContinuousDeployment.scala
```diff
@@ -48,7 +48,10 @@ class ContinuousDeployment(deployments: Deployments) extends Lifecycle with Logg
       if (!ChangeFreeze.frozen(params.stage.name)) {
         log.info(s"Triggering deploy of ${params.toString}")
         try {
-          deployments.deploy(params)
+          deployments.deploy(params, requestSource = ContinuousDeploymentRequestSource) match {
```
philwills: Would a `leftMap` (https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/syntax/either.scala#L120) be more readable here, or would that make you think we cared about the result value?

riff-raff/app/controllers/DeployController.scala
```diff
@@ -73,7 +73,10 @@ class DeployController(deployments: Deployments, prismLookup: PrismLookup, deplo
               "previewStacks" -> parameters.stacks.flatMap(_.nameOption).mkString(",")
             )
           case "deploy" =>
-            val uuid = deployments.deploy(parameters)
+            val uuid = deployments.deploy(parameters, requestSource = UserRequestSource(request.user)) match {
+              case Right(uuid) => uuid
+              case Left(error) => throw new IllegalStateException(error)
```
philwills: I'm assuming this is a case of minimal viable interface.

riff-raff/app/views/restrictions/list.scala.html
```diff
@@ -0,0 +1,62 @@
...
+    } else {
+        <table class="table table-condensed">
+            <thead>
+                <tr>
+                    <th>Last edited</th>
+                    <th>Project Name</th>
+                    <th>Stage</th>
+                    <th>Locked</th>
+                    <th></th>
+                    <th></th>
```
adamnfish: If this `<th>` doesn't mean anything by itself you could colspan it with the previous one as the header for "actions".

:minor_point:

riff-raff/test/restrictions/RestrictionCheckerTest.scala
```diff
@@ -0,0 +1,168 @@
...
+
+import java.util.UUID
+
+import com.gu.googleauth.UserIdentity
+import controllers.ApiKey
+import deployment.{ApiRequestSource, ContinuousDeploymentRequestSource, UserRequestSource}
+import org.joda.time.DateTime
+import org.scalatest.{FlatSpec, Matchers}
+
+class RestrictionCheckerTest extends FlatSpec with Matchers {
```
adamnfish: There are a lot of tests here for "does my version of an if statement work like an if statement"

riff-raff/app/controllers/api.scala
```diff
@@ -294,21 +293,30 @@ class Api(deployments: Deployments, deploymentTypes: Seq[DeploymentType])(implic
...
+                  "build" -> build,
+                  "stage" -> stage,
+                  "recipe" -> recipe.name,
+                  "hosts" -> toJson(hosts)
+                ),
+                "uuid" -> deployId.toString,
+                "logURL" -> routes.DeployController.viewUUID(deployId.toString).absoluteURL()
+              )
+            )
+          case Left(error) =>
```
philwills: This feels like the job of some transformer from an API result type rather than the responsibility of a particular endpoint.

riff-raff/app/deployment/Deployments.scala
```diff
@@ -21,25 +22,29 @@ import scala.util.control.NonFatal
...
     log.info(s"Started deploying $requestedParams")
-    if (enableQueueingSwitch.isSwitchedOff)
-      throw new IllegalStateException("Unable to queue a new deploy; deploys are currently disabled by the %s switch" format enableQueueingSwitch.name)
-
-    val params = if (requestedParams.build.id != "lastSuccessful")
-      requestedParams
+    val restrictionsPreventingDeploy = RestrictionChecker.configsThatPreventDeployment(RestrictionConfigDynamoRepository,
+      requestedParams.build.projectName, requestedParams.stage.name, requestSource)
+    if (restrictionsPreventingDeploy.nonEmpty) {
+      Left(s"Unable to queue deploy as restrictions are currently in place: ${restrictionsPreventingDeploy.map(r => s"${r.fullName}: ${r.note}").mkString("; ")}")
```
philwills: I'd rather we had a type other than `String` for this, but happy for that to come later.

riff-raff/app/persistence/RestrictionConfigDynamoRepository.scala
```diff
@@ -0,0 +1,22 @@
+package persistence
+
+import java.util.UUID
+
+import com.gu.scanamo.Table
+import restrictions.{RestrictionConfig, RestrictionsConfigRepository}
+
+object RestrictionConfigDynamoRepository extends DynamoRepository with RestrictionsConfigRepository {
```
philwills: I hate this kind of redundant interface, but the alternative isn't as simple as it should be without lots of wider changes.

riff-raff/app/restrictions/RestrictionChecker.scala
```diff
@@ -0,0 +1,49 @@
...
+  def sourceMatches(config: RestrictionConfig, source: RequestSource): Boolean = {
+    source match {
+      case ContinuousDeploymentRequestSource => config.continuousDeployment
+      case UserRequestSource(identity) => config.whitelist.contains(identity.email)
+      case _ => false
+    }
+  }
+
+  def stageMatches(configStage: String, stage: String): Boolean =
+    if (configStage.matches(""".*[$^.+*?()\[{|].*""")) {
```
philwills: Yikes!

riff-raff/app/controllers/Restrictions.scala
```diff
@@ -0,0 +1,86 @@
...
+        }
+      }
+    )
+  }
+  def edit(id: String) = AuthAction { implicit request =>
+    RestrictionConfigDynamoRepository.getRestriction(UUID.fromString(id)).map{ rc =>
+      val canSave = RestrictionChecker.editable(Some(rc), request.user)(true){_=>false}
+      Ok(views.html.restrictions.form(restrictionsForm.fill(
+        RestrictionForm(rc.id, rc.projectName, rc.stage, rc.editingLocked, rc.whitelist, rc.continuousDeployment, rc.note)
+      ), saveDisabled = !canSave))
```
adamnfish: This is quite an important piece of logic and it's hidden at the back of the arg list behind a `!`.

riff-raff/app/controllers/Restrictions.scala
```diff
@@ -0,0 +1,86 @@
...
+            request.user.email, f.editingLocked, f.whitelist, f.continuousDeployment, f.note)
+          RestrictionConfigDynamoRepository.setRestriction(newConfig)
+          Redirect(routes.Restrictions.list())
+        }
+      }
+    )
+  }
+  def edit(id: String) = AuthAction { implicit request =>
+    RestrictionConfigDynamoRepository.getRestriction(UUID.fromString(id)).map{ rc =>
+      val canSave = RestrictionChecker.editable(Some(rc), request.user)(true){_=>false}
```
adamnfish: You should assign the boolean in the state you mean to use it. `cannotSave`, in this case.

riff-raff/app/controllers/Restrictions.scala
adamnfish: (see also above)

adamnfish:
@philwills yeah, I mentioned IRL having admins that can override a locked deployment might be handy. That said, I think in the security case of "I intend for the only way to deploy this to be continuous deployment" it makes sense to be able to lock it down to some extent.

TBonnin:
As discussed in person yesterday, it seems to me the use case for blocking is more at deploy level rather than project level, but I understand Riff-Raff doesn't keep track of deploy entities at the moment.
For example it mean that blocking multi-projects deploy like Frontend would requires creating multiple restrictions. Not ideal but fine for now I guess.

riff-raff/app/controllers/DeployController.scala
sihil: Ideally we'd eliminate the throwing of exceptions entirely, but pulling the exception throwing up to the top level is a step in the right direction without making too many sweeping changes.

I've tweaked this to use `valueOr` from the cats either syntax.

riff-raff/app/controllers/api.scala
sihil: Agreed - `ApiJsonEndpoint` should probably take something like `Request => Either[ApiError, ApiResult]`.

Not in this PR though, do you want to raise an issue?

riff-raff/app/deployment/Deployments.scala
sihil: It was easy to do so I've created a small `case class Error(message: String) extends AnyVal`.

riff-raff/app/persistence/RestrictionConfigDynamoRepository.scala
sihil: Testing is non-trivial without it. Having a local dynamodb for tests would also eliminate it.

riff-raff/app/ci/ContinuousDeployment.scala
sihil: Ah yes - exactly what I did after @philwills similar comment.

riff-raff/app/controllers/DeployController.scala
```diff
@@ -73,7 +73,10 @@ class DeployController(deployments: Deployments, prismLookup: PrismLookup, deplo
               "previewStacks" -> parameters.stacks.flatMap(_.nameOption).mkString(",")
             )
           case "deploy" =>
-            val uuid = deployments.deploy(parameters)
+            val uuid = deployments.deploy(parameters, requestSource = UserRequestSource(request.user)) match {
```
sihil: I've used cats' `valueOr` pimp from `cats.syntax.either._`.

riff-raff/app/controllers/Restrictions.scala
```diff
@@ -0,0 +1,86 @@
...
+      "note" -> nonEmptyText
+    )((id, projectName, stage, editingLocked, whitelist, cdPermitted, note) =>
+      RestrictionForm(id, projectName, stage, editingLocked,
+        whitelist.map(_.split('\n').toSeq.filter(_.nonEmpty)).getOrElse(Seq.empty), cdPermitted, note)
+    )(f =>
+      Some((f.id, f.projectName, f.stage, f.editingLocked, Some(f.whitelist.mkString("\n")),
+        f.continuousDeployment,  f.note))
+    ).verifying(
+      "Stage is invalid - should be a valid regular expression or contain no special values",
+      form => try { form.stage.r; true } catch { case e:PatternSyntaxException => false }
```
sihil: I thought that would catch fatal errors - but actually Try has that covered which I hadn't realised. Nice idea.

riff-raff/app/controllers/Restrictions.scala
sihil: Fairly important, although this is for display purposes only rather than enforcing it. I've rearranged the code so it's clearer.

riff-raff/app/controllers/Restrictions.scala
```diff
@@ -0,0 +1,86 @@
...
+    Ok(views.html.restrictions.list(configs))
+  }
+
+  def form = AuthAction { implicit request =>
+    val newForm = restrictionsForm.fill(
+      RestrictionForm(UUID.randomUUID(), "", "", editingLocked = false, Seq.empty, continuousDeployment = false, ""))
+    Ok(views.html.restrictions.form(newForm, saveDisabled = false))
+  }
+
+  val editableOrForbidden: (Option[RestrictionConfig], UserIdentity) => (=> Result) => Result =
```
sihil: I **_am_** trying to be clever.

riff-raff/app/controllers/Restrictions.scala
```diff
@@ -0,0 +1,86 @@
...
+    (maybeConfig, identity) => (ifEditable) =>
+    RestrictionChecker.editable[Result](maybeConfig, identity)(ifEditable){ reason =>
+      Forbidden(s"Not possible to edit this restriction: $reason")
+    }
+
+  def save = AuthAction { implicit request =>
+    restrictionsForm.bindFromRequest().fold(
+      formWithErrors => Ok(views.html.restrictions.form(formWithErrors, saveDisabled = false)),
+      f => {
+        editableOrForbidden(RestrictionConfigDynamoRepository.getRestriction(f.id), request.user) {
```
sihil: Fair comment. I've made various modifications to address this.

riff-raff/app/deployment/Deployments.scala
sihil: Good spot.

riff-raff/app/restrictions/RestrictionChecker.scala
sihil: Removed.

riff-raff/app/restrictions/RestrictionConfig.scala
sihil: Regex doesn't support Scala like equality (i.e. "a".r != "a".r) so I tend not to use it as a field in case classes  - it breaks the case class semantics.

riff-raff/app/views/restrictions/form.scala.html
sihil: 😄  - `b3` actually has a `formCSRF` that does this for you so I've just switched over.

riff-raff/test/restrictions/RestrictionCheckerTest.scala
sihil: With the change of signature and use these are now saner tests.

sihil:
@TBonnin I think that's the right level for a restriction usually. I'm hoping that projects and deploys should match much more closely once teams move to riff-raff.yaml. Potentially adding API endpoints for adding and removing restrictions might be helpful for doing batches in the case where this is not enough - or further improvements (potentially regexes on the project name too?)

riff-raff/app/persistence/RestrictionConfigDynamoRepository.scala
philwills: I'll look to add https://github.com/localytics/sbt-dynamodb

sihil:
OK - I've addressed all of the feedback and I've also added a really simple superuser feature that allows users in a list of e-mail addresses (in the config) to edit any configuration. We can add half a dozen or so users to this list as a safeguard against people abusing the locking feature.

PR #382, opened 2016-10-26
Provide the real deployment type name

sihil:
Sadly broke the deployment type page earlier. I've fixed this and also extracted the rather gnarly logic into a separate object that can be more easily tested. And written a couple of tests to do exactly that.


sihil:
Will extract some of the logic and make a basic test.

PR #383, opened 2016-10-31
Ability to Filter a YAML deploy

sihil:
Hopefully the final piece in the puzzle is the ability to deploy only a subset of components. This PR does that and consists of the following pieces:
 - Introduce the idea of filters
 - Implement a simple `DeploymentId` filter
 - Add the filter to the existing `DeployParameters` class
 - Honour the filter whilst building a graph
 - Persist the new filter field to the database (automagic is wonderful - thanks @cb372)


riff-raff/app/utils/Forms.scala
```diff
@@ -1,21 +1,51 @@
...
+      val output = Map(key -> idToString(value))
+      logger.info(s"deploymentIdList/unbind: $output")
+      output
+    }
   })
+
+  // serialisation and de-serialisation code for deployment IDs
+  val DEPLOYMENT_DELIMITER = '!'
+  val FIELD_DELIMITER = '*'
+  def idToString(d: List[DeploymentId]): String = {
```
philwills: I'd rather have this kind of logic in the companion object (possibly as a type class) than away in this Forms file. Seems eminently testable and a nice candidate for property based testing.

sihil: This conversion logic is peculiar to the frontend whilst the model is in the main library. This was the reason I broke it out as it is of no concern to the library that a web app might want a web safe string representation of this case class. It should however have some tests.

sihil:
I've addressed all your feedback @philwills - think this is pretty much ready to be merged.

PR #384, opened 2016-11-10
Extends deployment type documentation to cover actions

sihil:
This extends the deployment type documentation that currently covers parameters to also cover actions. This has been implemented in exactly the same way as parameters were. Sadly this has not yet stretched to automatic documentation of tasks and I suspect this would be difficult, especially for actions that have parameters without any default.

The overall documentation has also been improved as a result.

screen shot 2016-11-10 at 13 54 40 1

### DIFF
--- a/magenta-lib/src/main/scala/magenta/Project.scala
+++ b/magenta-lib/src/main/scala/magenta/Project.scala
@@ -78,7 +78,7 @@ case class DeployTarget(parameters: DeployParameters, stack: Stack, region: Regi
  An action represents a step within a recipe. It isn't executable
  until it's resolved against a particular host.
  */
-trait Action {
+trait ActionResolver {
   def apps: Seq[App]
   def description: String
   def resolve(resources: DeploymentResources, target: DeployTarget): List[Task]
@@ -88,7 +88,7 @@ case class App (name: String)
 
 case class Recipe(
   name: String,
-  actions: Iterable[Action] = Nil, //executed once per app (before the host actions are executed)
+  actions: Iterable[ActionResolver] = Nil, //executed once per app (before the host actions are executed)
   dependsOn: List[String] = Nil
 )
 
--- a/magenta-lib/src/main/scala/magenta/Project.scala
+++ b/magenta-lib/src/main/scala/magenta/Project.scala
@@ -74,10 +74,7 @@ case class DeploymentResources(reporter: DeployReporter, lookup: Lookup, artifac
 
 case class DeployTarget(parameters: DeployParameters, stack: Stack, region: Region)
 
-/*
- An action represents a step within a recipe. It isn't executable
- until it's resolved against a particular host.
- */
+/* An ActionResolver represents a step within a deployment. */
 trait ActionResolver {
   def apps: Seq[App]
   def description: String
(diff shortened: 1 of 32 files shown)


magenta-lib/src/main/scala/magenta/deployment_type/Action.scala
```diff
@@ -0,0 +1,14 @@
+package magenta.deployment_type
+
+import magenta.{DeployTarget, DeploymentPackage, DeploymentResources}
+import magenta.tasks.Task
+
+trait ActionRegister {
+  def add(action: Action)
```
alexduf: that would be nice to have a return type here. I can guess it's a side effect, but having `Unit` would make it super clear

magenta-lib/src/main/scala/magenta/deployment_type/AutoScaling.scala
```diff
@@ -12,17 +12,7 @@ object AutoScaling  extends DeploymentType {
...
-      | - wait for the new instances to enter service
-      | - terminate previously tagged instances
-      |
-      |The action checks whether the auto-scaling group maxsize is big enough before starting the process.
-      |
-      |It also suspends and resumes cloud watch alarms in order to prevent false alarms.
-      |
-      |This deploy type has two actions, `deploy` and `uploadArtifacts`. `uploadArtifacts` simply uploads the files
-      |in the package directory to the specified bucket. `deploy` carries out the auto-scaling group rotation.
+      | - scale up, wait for the new instances to become healthy and ten scale back down
```
alexduf: s/ten/then/

magenta-lib/src/main/scala/magenta/deployment_type/Lambda.scala
```diff
@@ -125,41 +118,60 @@ object Lambda extends DeploymentType  {
...
+    }.distinct
+  }
+  val updateLambda = Action("updateLambda",
+    """
+      |Updates the lambda to use new code using the UpdateFunctionCode API.
+      |
+      |Typically this copies the new function code form S3, but when `bucket` is not provided this directly uploads the
+      |file directly into the function.
+      |
+      |The fucntion name to update is determined by the `functionName` or `functions` parameters.
```
alexduf: s/fucntion/function/

magenta-lib/src/main/scala/magenta/Project.scala
```diff
@@ -78,7 +78,7 @@ case class DeployTarget(parameters: DeployParameters, stack: Stack, region: Regi
  An action represents a step within a recipe. It isn't executable
```
philwills: Do we need to update or delete the comment associated with this trait?

magenta-lib/src/main/scala/magenta/deployment_type/Action.scala
philwills: +1

magenta-lib/src/main/scala/magenta/Project.scala
```diff
@@ -74,10 +74,7 @@ case class DeploymentResources(reporter: DeployReporter, lookup: Lookup, artifac
 
 case class DeployTarget(parameters: DeployParameters, stack: Stack, region: Region)
 
-/*
- An action represents a step within a recipe. It isn't executable
- until it's resolved against a particular host.
- */
+/* An ActionResolver represents a step within a deployment. */
```
philwills: Why don't we call it `DeploymentStep` then?

PR #385, opened 2016-11-14
Support looking up CloudFormation stack by tags

davidfurey:
Add a param to `UpdateAmiCloudFormationParameterTask` to specify that
stack should be found based on app/stack/stage tags
(based on PR #303)

### DIFF
--- a/magenta-lib/src/main/scala/magenta/deployment_type/AmiCloudFormationParameter.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/AmiCloudFormationParameter.scala
@@ -40,10 +40,12 @@ object AmiCloudFormationParameter extends DeploymentType {
 
       val cloudFormationStackLookupStrategy = {
         if (cloudformationStackByTags(pkg, target, reporter)) {
+          assert(pkg.apps.size == 1, s"The $name package type can only be used when exactly one app is specified - you have [${pkg.apps.map(_.name).mkString(",")}]")
+          assert(target.stack.nameOption.isDefined, s"The $name package type can only be used when a stack is specified")
           LookupByTags(Map(
-            "Stage" -> List(target.parameters.stage.name),
-            "Stack" -> target.stack.nameOption.toList,
-            "App" -> pkg.pkgApps.map(_.name).toList
+            "Stage" -> target.parameters.stage.name,
+            "Stack" -> target.stack.nameOption.get,
+            "App" -> pkg.pkgApps.map(_.name).head
           ))
         } else {
           val stackName = target.stack.nameOption.filter(_ => prependStackToCloudFormationStackName(pkg, target, reporter))
--- a/magenta-lib/src/main/scala/magenta/deployment_type/AmiCloudFormationParameter.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/AmiCloudFormationParameter.scala
@@ -14,7 +14,7 @@ object AmiCloudFormationParameter extends DeploymentType {
 
   val cloudformationStackByTags = Param[Boolean]("cloudFormationStackByTags",
     documentation = "Whether to find the cloudFormationStack by name or by tags"
-  ).default(false)
+  ).defaultFromContext((pkg, _) => Right(!pkg.legacyConfig))
   val cloudFormationStackName = Param[String]("cloudFormationStackName",
     documentation = "The name of the CloudFormation stack to update"
   ).defaultFromContext((pkg, _) => Right(pkg.name))
--- a/magenta-lib/src/main/scala/magenta/deployment_type/AmiCloudFormationParameter.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/AmiCloudFormationParameter.scala
@@ -40,12 +40,19 @@ object AmiCloudFormationParameter extends DeploymentType {
 
       val cloudFormationStackLookupStrategy = {
         if (cloudformationStackByTags(pkg, target, reporter)) {
-          assert(pkg.apps.size == 1, s"The $name package type can only be used when exactly one app is specified - you have [${pkg.apps.map(_.name).mkString(",")}]")
-          assert(target.stack.nameOption.isDefined, s"The $name package type can only be used when a stack is specified")
-          LookupByTags(Map(
-            "Stage" -> target.parameters.stage.name,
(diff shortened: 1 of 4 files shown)


magenta-lib/src/main/scala/magenta/deployment_type/CloudFormation.scala
```diff
@@ -89,7 +90,7 @@ object CloudFormation extends DeploymentType {
           target.stack,
           createStackIfAbsent(pkg, target, reporter)
         ),
-        CheckUpdateEventsTask(fullCloudFormationStackName)
+        CheckUpdateEventsTask(LookupByName(fullCloudFormationStackName))
```
sihil: IMHO both of these deployment types should support tag lookup.

magenta-lib/src/main/scala/magenta/tasks/UpdateCloudFormationTask.scala
```diff
@@ -242,6 +265,20 @@ trait CloudFormation extends AWS {
     client.describeStackEvents(
       new DescribeStackEventsRequest().withStackName(name)
     )
+
+  def findStackByTags(tags: Map[String, List[String]], reporter: DeployReporter)(implicit keyRing:KeyRing) = {
+    val cfnStacks = client.describeStacks(
+      new DescribeStacksRequest()
+    ).getStacks.filter { stack =>
```
sihil: `describeStacks` is a paged API so a single call will potentially only return a subset of stacks. You probably want to make this call recursive depending on the value of `getNextToken`: http://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/cloudformation/model/DescribeStacksResult.html#getNextToken--

magenta-lib/src/main/scala/magenta/tasks/UpdateCloudFormationTask.scala
```diff
@@ -21,6 +22,14 @@ object UpdateCloudFormationTask {
   case object UseExistingValue extends ParameterValue
   case class TemplateParameter(key:String, default:Boolean)
 
+  sealed trait CloudFormationStackLookupStrategy
+  case class LookupByName(cloudFormationStackName: String) extends CloudFormationStackLookupStrategy {
+    override def toString = s"called $cloudFormationStackName"
+  }
+  case class LookupByTags(tags: Map[String, List[String]]) extends CloudFormationStackLookupStrategy {
```
davidfurey: Because with the json representation `pkgApps.size` can be be `>1`.  It would definitely be simpler to change to `Map[String,String]` if YAML enforces the size, and we are happy to just use the first app for json deploys.

magenta-lib/src/main/scala/magenta/deployment_type/CloudFormation.scala
davidfurey: Yes, they should.  Does that block this PR?

sihil: Seems like a relatively small piece of work, but no it doesn't block it.

magenta-lib/src/main/scala/magenta/deployment_type/AmiCloudFormationParameter.scala
```diff
@@ -34,21 +38,34 @@ object AmiCloudFormationParameter extends DeploymentType {
...
-      val cloudFormationStackNameParts = Seq(stackName, Some(cloudFormationStackName(pkg, target, reporter)), stageName).flatten
-      val fullCloudFormationStackName = cloudFormationStackNameParts.mkString("-")
+      val cloudFormationStackLookupStrategy = {
+        if (cloudformationStackByTags(pkg, target, reporter)) {
+          assert(pkg.apps.size == 1, s"The $name package type can only be used when exactly one app is specified - you have [${pkg.apps.map(_.name).mkString(",")}]")
+          assert(target.stack.nameOption.isDefined, s"The $name package type can only be used when a stack is specified")
+          LookupByTags(Map(
+            "Stage" -> target.parameters.stage.name,
+            "Stack" -> target.stack.nameOption.get,
+            "App" -> pkg.pkgApps.map(_.name).head
```
davidfurey: I'm not really happy about the `.head` and `.get`.  I suppose this would disappear if/when yaml becomes mandatory.

sihil: You could zip two options and map over it, but ultimately the assertions have been made above. A TODO comment noting that this could be removed once the legacy format is removed might be of use.

magenta-lib/src/main/scala/magenta/tasks/UpdateCloudFormationTask.scala
```diff
@@ -242,6 +267,35 @@ trait CloudFormation extends AWS {
...
+      else
+        recur(response.getNextToken, stacks)
+    }
+
+    recur() match {
+      case cfnStack :: Nil => Some(cfnStack)
+      case Nil => reporter.fail(s"No matching cloudformation stack match for $tags.")
+      case _ =>
+        reporter.fail(s"More than one cloudformation stack match for $tags. Failing fast since this may be non-deterministic.")
+    }
```
sihil: Part of me wonders if this logic should be outside of this function to make it less dependent on side-effects. But I'm aware that much of riff-raff is like this so I'm not too bothered.

magenta-lib/src/main/scala/magenta/deployment_type/AmiCloudFormationParameter.scala
```diff
@@ -40,12 +40,19 @@ object AmiCloudFormationParameter extends DeploymentType {
...
+            app <- pkg.pkgApps.map(_.name).headOption
+            stage = target.parameters.stage.name
+          } yield LookupByTags(Map(
+            "Stage" -> stage,
+            "Stack" -> stack,
+            "App" -> app
+          ))
+
+          lookupByTags.getOrElse(reporter.fail(
+            s"The $name package type can only be used when the stack and exactly one app is specified - you have stage=${target.stack.nameOption} and apps=${pkg.apps.map(_.name).mkString(",")}"
```
sihil: I think this no longer fails if `apps.size > 1`?

sihil:
You happy to merge and push this out @davidfurey? (It's configured to continuously deploy from master)

PR #386, opened 2016-11-14
Fix relative links in index.md

thomaska:
The relative links in the `index.md` files don't have the suffix `.md` and return a 404.
Eg.:
screen shot 2016-11-14 at 15 43 19

This is an attempt to repair the broken links.
cc @sihil

### DIFF
--- a/magenta-lib/docs/magenta-lib/concepts/index.md
+++ b/magenta-lib/docs/magenta-lib/concepts/index.md
@@ -7,7 +7,7 @@ actually works under the hood and give some details on how it understands the fi
 
 The following concepts are discussed:
 
- - [How Magenta works](how-magenta-works)
- - [The terminology of Magenta](terminology)
- - [The anatomy of the artifacts.zip file](artifacts.zip)
- - [The anatomy of a deploy.json file](deploy.json)
+ - [How Magenta works](how-magenta-works.md)
+ - [The terminology of Magenta](terminology.md)
+ - [The anatomy of the artifacts.zip file](artifacts.zip.md)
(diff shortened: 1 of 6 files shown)


philwills:
These are intended for viewing in the app, e.g. https://host/docs/magenta-lib/concepts/, rather than via Github and by that point, they don't need the `.md` extension. If we wanted to support links in Github we'd need to at least change the rendering app as well to support the `.md` extension.

sihil:
I suspect (but haven't checked) that this will break the documentation links when used inside riff-raff (https://riffraff.gutools.co.uk/docs/). Can you maybe deploy to CODE and confirm how the parser handles this? It might be that adding a new route that chops off the .md automatically would fix this for both github and internal browsing.

thomaska:
Oh ok. This makes sense then! Sorry I wasn't aware of the https://riffraff.gutools.co.uk/docs/ !
@sihil I tried deploying on CODE but I got:
`
[13:35:22] task ChangeSwitch shutdown-when-inactive to ON
[13:37:29] Unhandled exception in task ChangeSwitch shutdown-when-inactive to ON`

Is it ok if I just close this PR?

sihil:
Did you deploy from the CODE instance of Riff-Raff? At the moment you cannot cross deploy from PROD to CODE or CODE to PROD.

You're the second person to try doing this and I think it's a quick fix to make it work in both contexts so I'd advocate doing that.

thomaska:
@sihil Thanks for passing by today! 
I deployed on CODE and checked the github links and both seem ok now.

sihil:
Looks great. 👍

PR #388, opened 2016-11-17
Make actions a NEL for a fully resolved Deployment

philwills:
Actions are necessary for a deployment and this was enforced by the validation, but not previously reflected in the types.


sihil:
Nice approach. Could also introduce another type to use after the deployments are flattened (i.e. only one action, stack and region each). This is currently not compile time checked but is relied on in a couple of places.

PR #390, opened 2016-11-22
Add region and tag handling to Cloudformation

sihil:
This really follows up on #385 which made it possible to find a cloudformation stack by tags instead of by name in the `AmiCloudFormationParameter` deployment type. This does the same for the `CloudFormation` deployment type as well.

Whilst doing this work I noticed that the cloudformation deployment types ignored the region specification in the new riff-raff.yaml file so I've modified the tasks and client creation to facilitate this. Whilst doing this re-structure I moved the AWS Cloudformation object into same file as the rest of the AWS API wrappers.

In addition I've improved some of the documentation and centralised some logic that creates the lookup strategy.

It will help slightly to view this PR with `w=1` to hide whitespace only changes.

### DIFF
--- a/magenta-lib/src/main/scala/magenta/deployment_type/CloudFormation.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/CloudFormation.scala
@@ -2,7 +2,7 @@ package magenta.deployment_type
 
 import magenta.artifact.S3Path
 import magenta.tasks.{CheckUpdateEventsTask, UpdateCloudFormationTask}
-import magenta.tasks.UpdateCloudFormationTask.LookupByName
+import magenta.tasks.UpdateCloudFormationTask.{LookupByName, LookupByTags}
 
 object CloudFormation extends DeploymentType {
   val name = "cloud-formation"
@@ -19,6 +19,12 @@ object CloudFormation extends DeploymentType {
       |current state.
     """.stripMargin
 
+  val cloudformationStackByTags = Param[Boolean]("cloudFormationStackByTags",
+    documentation =
+      """When false we derive the stack using the `cloudFormationStackName`, `prependStackToCloudFormationStackName` and
+        |`appendStageToCloudFormationStackName` parameters. When true we find the stack by looking for one with matching
+        |stack, app and stage tags in the same way that autoscaling groups are discovered.""".stripMargin
+  ).defaultFromContext((pkg, _) => Right(!pkg.legacyConfig))
   val cloudFormationStackName = Param[String]("cloudFormationStackName",
     documentation = "The name of the CloudFormation stack to update"
   ).defaultFromContext((pkg, _) => Right(pkg.name))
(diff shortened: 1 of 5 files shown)


magenta-lib/src/main/scala/magenta/deployment_type/CloudFormation.scala
```diff
@@ -69,18 +75,28 @@ object CloudFormation extends DeploymentType {
...
-      val stackName = target.stack.nameOption.filter(_ => prependStackToCloudFormationStackName(pkg, target, reporter))
-      val stageName = Some(target.parameters.stage.name).filter(_ => appendStageToCloudFormationStackName(pkg, target, reporter))
-      val cloudFormationStackNameParts = Seq(stackName, Some(cloudFormationStackName(pkg, target, reporter)), stageName).flatten
-      val fullCloudFormationStackName = cloudFormationStackNameParts.mkString("-")
+      val cloudFormationStackLookupStrategy = {
+        if (cloudformationStackByTags(pkg, target, reporter)) {
+          LookupByTags(pkg, target, reporter)
+        } else {
+          LookupByName(
+            target.stack,
```
davidfurey: Why not use named params for all here?

magenta-lib/src/main/scala/magenta/tasks/UpdateCloudFormationTask.scala
```diff
@@ -63,45 +95,57 @@ case class UpdateCloudFormationTask(
...
+            reporter.info("Cloudformation update has no changes to template or parameters")
+          case ase:AmazonServiceException if ase.getMessage contains "Template format error: JSON not well-formed" =>
+            reporter.info(s"Cloudformation update failed with the following template content:\n$templateString")
+            throw ase
+        }
+      case None =>
+        if (createStackIfAbsent) {
+          val nameToCallStack = UpdateCloudFormationTask.nameToCallNewStack(cloudFormationStackLookupStrategy)
+          reporter.info(s"Stack $cloudFormationStackLookupStrategy doesn't exist. Creating stack using name $nameToCallStack.")
+          CloudFormation.createStack(reporter, nameToCallStack, templateString, parameters, cfnClient)
```
davidfurey: This needs to tag the new stack if strategy is LookupByTags

magenta-lib/src/main/scala/magenta/deployment_type/CloudFormation.scala
sihil: I was just making intelliJ's default scala analysis happy - it likes boolean params to be named. I think the others are pretty obvious.

magenta-lib/src/main/scala/magenta/tasks/UpdateCloudFormationTask.scala
```diff
@@ -63,45 +95,61 @@ case class UpdateCloudFormationTask(
...
+          case ase:AmazonServiceException if ase.getMessage contains "No updates are to be performed." =>
+            reporter.info("Cloudformation update has no changes to template or parameters")
+          case ase:AmazonServiceException if ase.getMessage contains "Template format error: JSON not well-formed" =>
+            reporter.info(s"Cloudformation update failed with the following template content:\n$templateString")
+            throw ase
+        }
+      case None =>
+        if (createStackIfAbsent) {
+          val nameToCallStack = UpdateCloudFormationTask.nameToCallNewStack(cloudFormationStackLookupStrategy)
+          val stackTags = cloudFormationStackLookupStrategy match {
```
davidfurey: Could use PartialFunction.condOpt here instead

davidfurey:
Looks good to me, thanks Simon.
👍

PR #391, opened 2016-11-22
Fixes and improved logging for ASG deployments

sihil:
@johnduffell had some issues with an ASG deploy the other day. Understanding what happened was difficult due to the lack of information surfaced by Riff-Raff. This PR adds a bunch of extra logging but also fixes a potentially dangerous bug in `CheckForStabilization`.

`CheckForStabilization` previously checked whether the ASG was stable and then allowed the deployment to proceed regardless of the outcome of the check. The deployment will now fail with a helpful error.

### DIFF
--- a/magenta-lib/src/main/scala/magenta/tasks/ASGTasks.scala
+++ b/magenta-lib/src/main/scala/magenta/tasks/ASGTasks.scala
@@ -10,6 +10,7 @@ import scala.collection.JavaConversions._
 case class CheckGroupSize(pkg: DeploymentPackage, stage: Stage, stack: Stack, region: Region)(implicit val keyRing: KeyRing) extends ASGTask {
   override def execute(asg: AutoScalingGroup, reporter: DeployReporter, stopFlag: => Boolean, asgClient: AmazonAutoScalingClient) {
     val doubleCapacity = asg.getDesiredCapacity * 2
+    reporter.verbose(s"ASG desired = ${asg.getDesiredCapacity}; ASG max = ${asg.getMaxSize}; Target = $doubleCapacity")
     if (asg.getMaxSize < doubleCapacity || doubleCapacity == 0) {
       reporter.fail(
         s"Autoscaling group does not have the capacity to deploy current max = ${asg.getMaxSize} - desired max = $doubleCapacity"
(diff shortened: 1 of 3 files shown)


magenta-lib/src/main/scala/magenta/tasks/ASGTasks.scala
```diff
@@ -58,7 +62,10 @@ case class WarmupGrace(durationMillis: Long)(implicit keyRing: KeyRing) extends
 case class CheckForStabilization(pkg: DeploymentPackage, stage: Stage, stack: Stack, region: Region)(implicit val keyRing: KeyRing) extends ASGTask {
   override def execute(asg: AutoScalingGroup, reporter: DeployReporter, stopFlag: => Boolean, asgClient: AmazonAutoScalingClient) {
     implicit val elbClient = ELB.makeElbClient(keyRing, region)
-    ASG.isStabilized(asg, asgClient, elbClient)
+    ASG.isStabilized(asg, asgClient, elbClient) match {
+      case Left(reason) => reporter.fail(s"ASG not stable: $reason")
+      case Right(true) =>
```
johnduffell: how about Right(false) ?

magenta-lib/src/main/scala/magenta/tasks/AWS.scala
```diff
@@ -56,16 +56,25 @@ object ASG extends AWS {
...
     elbName(asg) match {
       case Some(name) => {
         val elbHealth = ELB.instanceHealth(name, elbClient)
-        elbHealth.size == asg.getDesiredCapacity && elbHealth.forall( instance => instance.getState == "InService")
+        if (elbHealth.size != asg.getDesiredCapacity)
+          Left(s"Number of ELB instances (${elbHealth.size}) and ASG desired capacity (${asg.getDesiredCapacity}) don't match")
+        else if (!elbHealth.forall( instance => instance.getState == "InService"))
+          Left(s"Only ${elbHealth.count(_.getState == "InService")} of ${elbHealth.size} ELB instances InService")
+        else
+          Right(true)
```
johnduffell: if it's only ever Right(true) then why not make it Right(())  i.e. unit? otherwise it's confusing.

The other option is to make it an Option[String] where it's an optional error.  But that's going to seriously mess with people's minds!

johnduffell:
great! I think that will really help us, thanks 👍 🚀 
just a few stylistic comments you can see above

PR #392, opened 2016-11-24
Support Github HTTPS URLs

philwills:
Also got rid of support for the ancient realm of `git.gudev.gnl` and moved to URI instead of URL because URL does blocking DNS lookups.


AWare:
👍

sihil:
Lovely - we should actually expose this in the UI somewhere...

philwills:
@sihil YES!!!!!!!!

PR #394, opened 2016-11-25
Link from deploy to VCS commit

philwills:
I've been trying to think of a good way to do this, but this works
and doesn't have any significant negative impacts.

image


philwills:
cc @guardian/deploy-infrastructure

PR #395, opened 2016-11-25
Who doesn't like context menus?

sihil:
Great question right? Inspired by #394 I thought I'd implement a context menu on the history screen that allows you to directly access a bunch of features.

It looks a bit like this:
context

In addition, I've converted a bunch of the JS dependencies over to use webjars to make upgrading a little easier (and upgraded them in the process - hopefully haven't broken anything...).


adamnfish:
"People using web browsers to interact with your application"

They aren't discoverable, aren't helpful on touchscreens and override native behaviours. Maybe a button or drop down on each row to expose these interactions?

:opinion_warning:

riff-raff/app/assets/javascripts/history-context-menu.coffee
```diff
@@ -0,0 +1,71 @@
...
+      url+separator+param+'='+encodedValue
+
+  menu = null
+  if (window.autoRefresh)
+    window.autoRefresh.postRefresh ->
+      if menu?
+        menu.destroy()
+      menu = new BootstrapMenu('.history-context-menu', {
+        fetchElementData: (rowElem) ->
+          console.log 'doing this'
```
adamnfish: Did you leave this here by mistake?

philwills: Doin' it cause it feels good!

riff-raff/app/views/snippets/recordTable.scala.html
```diff
@@ -22,42 +22,46 @@
...
+                    @record.metaData.get(deployment.Record.RIFFRAFF_HOSTNAME).getOrElse(Html(""))
+                }{
+                @defining(record.state) { state =>
+                @state match {
+                case RunState.Completed => { <span class="label label-success">Completed</span> }
+                case RunState.Failed => { <span class="label label-danger">Failed</span> }
+                case RunState.NotRunning => { <span class="label label-default">Waiting</span> }
+                case _ => {
+                    <span class="label label-info">Running (@record.completedPercentage%)</span>
+                }
```
adamnfish: Oh god the braces

philwills:
I kind of agree with @adamnfish, but all this functionality is available by other means, so I don't think the discoverability is as much of a problem as it might be.

sihil:
I've made a few more tweaks and squeezed on a menu to the right hand side. Now looks like this:
screen shot 2016-11-28 at 11 19 44

riff-raff/app/views/snippets/recordTable.scala.html
sihil: No option in twirl templates - braces are required.

PR #397, opened 2016-11-28
Change Riff-Raff to be able to deal with templates over 50kB in size

sihil:
If a template body is larger than 50kB in size (51200 bytes) then we cannot use the CloudFormation SDK's `templateBody`. Instead we must upload the template to S3 and use `templateUrl` instead.

In order to avoid dealing with more bucket configuration my suggestion is that we create a unique bucket for the purpose of temporarily storing these templates. This PR creates buckets with the name: `riff-raff-cfn-templates-<accountNumber>-<region>` (with a lifecycle expiration rule of one day). This can be created by hand if preferred. If it exists (or can be created) then the template will be uploaded to an object key that is named using the stack and current time in milliseconds.

This doesn't alter the SDK path when the template is smaller than 50kB.

/cc @TBonnin

### DIFF
--- a/magenta-lib/src/main/scala/magenta/tasks/AWS.scala
+++ b/magenta-lib/src/main/scala/magenta/tasks/AWS.scala
@@ -16,16 +16,49 @@ import com.amazonaws.services.elasticloadbalancing.AmazonElasticLoadBalancingCli
 import com.amazonaws.services.elasticloadbalancing.model.{Instance => ELBInstance, _}
 import com.amazonaws.services.lambda.AWSLambdaClient
 import com.amazonaws.services.lambda.model.UpdateFunctionCodeRequest
-import com.amazonaws.services.s3.AmazonS3Client
+import com.amazonaws.services.s3.model.BucketLifecycleConfiguration
+import com.amazonaws.services.s3.model.BucketLifecycleConfiguration.Rule
+import com.amazonaws.services.s3.{AmazonS3, AmazonS3Client}
+import com.amazonaws.services.securitytoken.AWSSecurityTokenServiceClient
+import com.amazonaws.services.securitytoken.model.GetCallerIdentityRequest
 import magenta.{App, DeployReporter, DeploymentPackage, KeyRing, NamedStack, Region, Stack, Stage, UnnamedStack}
 
 import scala.annotation.tailrec
 import scala.collection.JavaConverters._
-import scala.collection.mutable
 
 object S3 extends AWS {
   def makeS3client(keyRing: KeyRing, region: Region, config: ClientConfiguration = clientConfiguration): AmazonS3Client =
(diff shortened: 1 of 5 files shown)


magenta-lib/src/main/scala/magenta/tasks/UpdateCloudFormationTask.scala
```diff
@@ -78,6 +79,19 @@ object UpdateCloudFormationTask {
...
+    val templateTooBigForSdkUpload = templateBody.length > 51200
+
+    if (templateTooBigForSdkUpload) {
+      val bucketName = S3.accountSpecificBucket("riff-raff-cfn-templates", s3Client, stsClient, region, Some(1))
+      val keyName = s"$stackName-${new DateTime().getMillis}"
+      s3Client.putObject(bucketName, keyName, templateBody)
+      TemplateUrl(s"s3://$bucketName/$keyName")
+    } else {
+      TemplateBody(templateBody)
+    }
```
TBonnin: why not always upload it to S3? won't be slower and we wouldn't rely on this limit of 51200 bytes that might change eventually

magenta-lib/src/main/scala/magenta/tasks/AWS.scala
```diff
@@ -16,16 +16,49 @@ import com.amazonaws.services.elasticloadbalancing.AmazonElasticLoadBalancingCli
...
 
 object S3 extends AWS {
   def makeS3client(keyRing: KeyRing, region: Region, config: ClientConfiguration = clientConfiguration): AmazonS3Client =
     new AmazonS3Client(provider(keyRing), config).withRegion(awsRegion(region))
+
+  /**
+    * Check (and if necessary create) that an S3 bucket exists for the account and region. Note that it is assumed that
+    * the clients and region provided all match up - i.e. they are all configured to the same region and use the same
+    * credentials.
+    * @param prefix The prefix for the bucket. The resulting bucket will be <prefix>-<accountNumber>-<region>.
```
TBonnin: Do we need a bucket per account?

sihil: I think so, otherwise we need to set up crazy permissions. This follows a similar approach to Amazon's CFN console that creates a bucket per account and per region.

magenta-lib/src/main/scala/magenta/tasks/UpdateCloudFormationTask.scala
sihil: My aim was to not change the existing path (it potentially needs more permissions - STS and S3). But it does mean more maintenance pain.

sihil: On second thoughts I've changed it so that if you are using the new riff-raff.yaml format then we'll upload by default. This seems like a good compromise - as people switch over they'll pick up the new behaviour without it breaking existing stuff.

PR #401, opened 2016-12-01
Add pages to view and download artifact files

sihil:
This adds a super simple artifact listing that shows the list of files that are available to riff-raff. This is helpful for debugging why things are not working as expected. It is also possible to download the files directly via Riff-Raff. I don't think this should be a security issue, but it's possible teams are putting secret data in their deployable artifacts.

I've added links to access this from the preview, history and deploy log pages.

Looks like this:

screen shot 2016-12-01 at 18 08 37

 /cc @adamnfish

### DIFF
--- a/riff-raff/app/controllers/DeployController.scala
+++ b/riff-raff/app/controllers/DeployController.scala
@@ -3,6 +3,8 @@ package controllers
 import java.net.URLEncoder
 import java.util.UUID
 
+import akka.stream.scaladsl.{Source, StreamConverters}
+import akka.util.ByteString
 import ci.{Builds, S3Tag, TagClassification}
 import conf.Configuration
 import controllers.forms.{DeployParameterForm, UuidForm}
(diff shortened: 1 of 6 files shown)


riff-raff/app/views/artifact/listFiles.scala.html
```diff
@@ -0,0 +1,21 @@
...
+    <hr/>
+    <div>
+        @if(objects.isEmpty) {
+            No files
+        } else {
+            <table class="table table-condensed">
+                <thead><tr><td>Key</td><td class="text-right">Object size (bytes)</td></tr></thead>
+                <tbody>
+                @for((path, obj) <- objects) {
+                    <tr><td><a href="@routes.DeployController.getArtifactFile(obj.key)">@path</a></td><td class="text-right">@obj.size</td></tr>
```
adamnfish: `!! -h`

philmcmahon:
👍

riff-raff/app/controllers/DeployController.scala
```diff
@@ -281,4 +285,31 @@ class DeployController(deployments: Deployments, prismLookup: PrismLookup, deplo
     }
   }
 
+  def deployFiles(projectName: String, id: String) = AuthAction { implicit request =>
+    def pathAndContent(artifact: S3Artifact): Option[(S3Artifact, S3Path, String)] = {
```
adamnfish: If this is worth extracting, is it worth testing? If it's useful "logic" then it might be better placed out of the controller with a unit test on it.

adamnfish: Ah, I see it is impure :-/

This is ok so these are minor points, but this controller method feels alarming to me. It looks like the main goal of this first section is to try both deploy object paths, but that's hidden away in a function that transforms a few bits of data into a format you use.
Strange as well, that they get transformed into a special format only to be immediately matched on. Would it not be better to just have these be in scope? A for-comp seems like it might be a good fit for doing this.

Related, `deployObject` seems like a redundant name and I suspect it only exists to try and help explain why we even have this (potentially unnecessary) intermediate structure.

It feels like with a bit of thought this could be significantly simpler.

PR #402, opened 2016-12-02
Correctly construct the path to the legacy artifacts.zip

sihil:
In #400, I missed a location where a `/` is being explicitly added on an S3 path. This prevents initial deploys that are using the legacy `artifacts.zip` format (prior to them being converted to the new format).


philmcmahon:
👍 👍 👍

PR #403, opened 2016-12-05
Wait again for stabilisation after instance termination

sihil:
Fixing `CheckForStabilization` in #391 has revealed that an ASG can continue to be unstable after a deploy has completed. This happens because the `CullInstancesWithTerminationTag` task is asynchronous and causes an ASG to become unstable again (the ASG desired number < actual number) whilst the ASG kills off the excess instances. 

This state doesn't last for long, but long enough that the `CheckForStabilization` task of a subsequent deploy can fail if it runs immediately after the initial deploy.

This PR introduces a small pause and then a further execution of the `WaitForStabilization` task to wait for the ASG to become stable again. I believe that this should probably only add the pause (default 10s) to most deploys, but might add another 30s if the check doesn't immediately pass.


philwills:
Making deploys slower (even if it's only apparent completion) makes me sad, but trading for consistency is worth a 👍

jfsoul:
Looks good to me, I will keep an eye out for close-together deploys that may have previously caused some failures.

PR #407, opened 2016-12-07
Add basic mixpanel tracking

sihil:
There is no data at the moment on how much Riff-Raff is used, in what way, or by whom. This adds some basic tracking which should help inform how features are further developed.

### DIFF
--- a/riff-raff/app/conf/context.scala
+++ b/riff-raff/app/conf/context.scala
@@ -177,6 +177,10 @@ class Configuration(val application: String, val webappConfDirectory: String = "
     }
   }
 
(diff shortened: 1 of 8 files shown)


riff-raff/app/conf/context.scala
```diff
@@ -177,6 +177,10 @@ class Configuration(val application: String, val webappConfDirectory: String = "
     }
   }
 
+  object mixpanel {
+    val token = configuration.getStringProperty("mixpanel.token")
```
sihil: I've already set this token in the PROD configuration file.

PR #408, opened 2016-12-12
Add buttons to select / deselect all of the tasks

sihil:
Deploying only one or tow tasks can be quite painful at the moment. This adds buttons that deselect and select all of the tasks in a YAML preview.

screen shot 2016-12-12 at 12 40 23

### DIFF
--- a/riff-raff/app/views/preview/yaml/preview.scala.html
+++ b/riff-raff/app/views/preview/yaml/preview.scala.html
@@ -4,7 +4,7 @@
         parameters: magenta.DeployParameters,
         previewId: String)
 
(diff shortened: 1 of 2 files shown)


riff-raff/app/views/preview/yaml/preview.scala.html
```diff
@@ -24,6 +24,11 @@
         </div>
 
         <div class="actions pull-right"><p>
+        <div class="btn-group" role="group" aria-label="...">
```
jfsoul: subjective nitpick: indent one more level here?

PR #412, opened 2016-12-21
Better logs

sihil:
Large logs are hard to grok at the moment - especially to find the errors. This change makes the logs collapsible and will automatically collapse successfully completed branches. However if a branch fails then it will by default be unfurled making it easy to identify the problems.

In order to do this the `DeployReport` code has been overhauled so as to expose the UUID of each log entry in the tree (in the process a superfluous top level of the tree has been removed and an `EmptyDeployReport` has been introduced). This provides an easy to use ID for targeting in the DOM using the standard bootstrap collapse feature. On top of this there is a little coffeescript that switches out the chevrons when you furl and unfurl part of the tree.

The dotcom deploy is most improved by this. A successful deploy now looks like this (two pages instead of a dozen):
screen shot 2016-12-21 at 14 18 32

In much the same way that selecting text can't be done whilst a deploy is running, it is not possible to unfurl parts of a still running deploy. Fixing this will require some smarter update code.

### DIFF
--- a/magenta-lib/src/main/scala/magenta/reporting.scala
+++ b/magenta-lib/src/main/scala/magenta/reporting.scala
@@ -2,7 +2,7 @@ package magenta
 
 import org.joda.time.{DateTime, DateTimeZone}
 import org.joda.time.format.DateTimeFormat
-import java.util.{UUID, Locale}
+import java.util.{Locale, UUID}
 
 object RunState extends Enumeration {
   type State = Value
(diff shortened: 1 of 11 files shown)


magenta-lib/src/main/scala/magenta/reporting.scala
```diff
@@ -42,134 +42,118 @@ trait MessageState {
...
-    }.getOrElse(MessageState(node.stack.top, node.stack.time))
-
-    ReportTree(messageState, children.map(wrapperToTree(_,all)))
-  }
-
-  def apply(list: List[MessageWrapper], title: String = "", titleTime: Option[DateTime] = None): ReportTree = {
-    val time = titleTime.getOrElse( list.headOption.map(_.stack.time).getOrElse( new DateTime() ))
-    ReportTree(Report(title, time), list.headOption.map(root => List(wrapperToTree(root,list))).getOrElse(Nil))
+  def failureMessage: Option[Fail] = {
+    allMessages.filter(_.message.getClass == classOf[Fail]).map(_.message).headOption.asInstanceOf[Option[Fail]]
```
TBonnin: Not related to this PR but I wonder how could one avoid relying on `asInstanceOf` in this case?

magenta-lib/src/main/scala/magenta/reporting.scala
```diff
@@ -42,134 +42,118 @@ trait MessageState {
...
+  val message = messageState.message
+  val timeString = Some(messageState.timeOfDay)
+  val state = messageState.state
+
+  private val childRunning: Boolean = children.foldLeft(false){_ || _.isRunning}
+  val isRunning: Boolean = messageState.isRunning || childRunning
 
-  def map(block: MessageState => MessageState): ReportTree = {
-    ReportTree(block(messageState), children.map(_.map(block)))
+  val allMessages: Seq[MessageState] = (messageState :: children.flatMap(_.allMessages)).sortWith{ (left, right) =>
```
philwills: `sortBy` might be neater here.

Issue #413, opened 2016-12-21
ParseError exceptions don't retry


sihil:
We've seen a few occasions where the XML returned from a `describeAutoScalingGroups` API call fails due to a `ParseError`. We've now got the exception (below) that causes this so we should be able to write some better retry handling. See #409 for further details.

We originally thought that it was an `IOException` that was causing this, but that's not true - it's an `javax.xml.stream.XMLStreamException`.

```
magenta.FailException: Unhandled exception in task SuspendAlarmNotifications Suspending Alarm Notifications - group will no longer scale on any configured alarms
        at magenta.DeployReporter$.magenta$DeployReporter$$failException(logging.scala:128)
        at magenta.DeployReporter$.failException(logging.scala:131)
        at magenta.DeployReporter$.withFailureHandling(logging.scala:106)
        at magenta.DeployReporter$.magenta$DeployReporter$$sendContext(logging.scala:118)
        at magenta.DeployReporter.taskContext(logging.scala:33)
        at deployment.actors.TasksRunner$$anonfun$receive$1$$anonfun$applyOrElse$1$$anonfun$apply$2.apply(TasksRunner.scala:36)
        at deployment.actors.TasksRunner$$anonfun$receive$1$$anonfun$applyOrElse$1$$anonfun$apply$2.apply(TasksRunner.scala:27)
        at scala.collection.immutable.List.foreach(List.scala:381)
        at deployment.actors.TasksRunner$$anonfun$receive$1$$anonfun$applyOrElse$1.apply(TasksRunner.scala:27)
        at deployment.actors.TasksRunner$$anonfun$receive$1$$anonfun$applyOrElse$1.apply(TasksRunner.scala:25)
        at magenta.DeployReporter$$anonfun$magenta$DeployReporter$$sendContext$1.apply(logging.scala:119)
        at magenta.DeployReporter$$anonfun$magenta$DeployReporter$$sendContext$1.apply(logging.scala:118)
        at magenta.DeployReporter$.withFailureHandling(logging.scala:98)
        at magenta.DeployReporter$.magenta$DeployReporter$$sendContext(logging.scala:118)
        at magenta.DeployReporter.infoContext(logging.scala:38)
        at deployment.actors.TasksRunner$$anonfun$receive$1.applyOrElse(TasksRunner.scala:25)
        at akka.actor.Actor$class.aroundReceive(Actor.scala:484)
        at deployment.actors.TasksRunner.aroundReceive(TasksRunner.scala:12)
[...]

PR #415, opened 2016-12-22
Improved exception handling

sihil:
This addresses #413 and should also eliminate the *DeploymentRunner unexpectedly terminated* messages.

The first of these is achieved by modifying the custom retry policy to retry on this one specific `XMLStreamException` (when the message starts with `ParseError at`).

The second is caused by `TasksRunner` actors throwing exceptions. The `DeployGroupRunner` actor watches the child `TasksRunner` actors that it creates so it is notified of unexpected terminations. In this case though the exception is expected and it is communicated back via a message. Unfortunately the exception is re-thrown (for the side effect of sending the appropriate error message to the `DeployReporter`). In order to prevent the Actor terminating we re-catch and blackhole NonFatal exceptions. The Actor will then be cleaned up when the `DeployGroupRunner` actor is cleaned up.

### DIFF
--- a/magenta-lib/src/main/scala/magenta/tasks/AWS.scala
+++ b/magenta-lib/src/main/scala/magenta/tasks/AWS.scala
@@ -1,6 +1,7 @@
 package magenta.tasks
 
 import java.nio.ByteBuffer
+import javax.xml.stream.XMLStreamException
 
 import com.amazonaws.{AmazonClientException, AmazonWebServiceRequest, ClientConfiguration}
 import com.amazonaws.auth.{AWSCredentialsProvider, AWSCredentialsProviderChain, BasicAWSCredentials}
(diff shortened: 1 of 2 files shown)


magenta-lib/src/main/scala/magenta/tasks/AWS.scala
```diff
@@ -431,20 +432,30 @@ trait AWS extends Loggable {
...
+   * not knowing whether the query is idempotent or not. There is more discussion about this at
+   * https://github.com/aws/aws-sdk-java/issues/892
+   * */
+  class RiffRaffRetryCondition extends SDKDefaultRetryCondition {
+    private def exceptionInfo(e: Throwable): String = {
       s"${e.getClass.getName} ${e.getMessage} Cause: ${Option(e.getCause).map(e => exceptionInfo(e))}"
     }
+    private def isParseException(exception: AmazonClientException): Boolean = {
+      exception.getCause match {
+        case xse:XMLStreamException if xse.getMessage.startsWith("ParseError at") => true
```
jfsoul: Bit sad that this is how we have to identify a ParseError.  Looks like non-null `xse.getLocation` might be equivalent - but probably no better than this.

riff-raff/app/deployment/actors/TasksRunner.scala
```diff
@@ -22,33 +24,38 @@ class TasksRunner(stopFlagAgent: Agent[Map[UUID, String]]) extends Actor with Lo
...
-          log.debug("Sending completed message")
-          sender ! DeployGroupRunner.DeploymentCompleted(tasks)
-        } catch {
-          case t:Throwable =>
-            log.debug("Sending failed message")
-            sender ! DeployGroupRunner.DeploymentFailed(tasks, t)
-            throw t
         }
+      } catch {
+        // catch non fatal exceptions and swallow them to avoid the actor being terminated at this point
```
philwills: 😢

sihil: Be great to walk through how it is preferable to handle this.

PR #418, opened 2017-01-25
The bucket should be optional for lambda on both deploy.json and riff-raff.yaml

mchv:
I think a `bucket` is not required as mentioned in the documentation
@sihil what do you think?


sihil:
I disagree. If you don't upload via an S3 bucket then you can't safely create the lambda using CloudFormation. I want to move people away from uploading directly to a Lambda and thus the defaults in YAML don't let you do this.

mchv:
@sihil Ok I understand your point.

PR #419, opened 2017-01-25
Allow updates to multiple AMI parameters

DiegoVazquezNanini:
- Add a new parameter call amiParametersToTags to handle multiples amiParameters/tags combinations.
- Add tests to ensure that the old amiTags and amiParameter keep having the same behaviour.

**Note**: changes to AmiCloudFormationParameter are really hard to test without a fundamental refactor therefore I plan to test the change in the CODE env.

@sihil @adamnfish @TBonnin @thomaska

### DIFF
--- a/magenta-lib/src/main/scala/magenta/deployment_type/AmiCloudFormationParameter.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/AmiCloudFormationParameter.scala
@@ -1,5 +1,6 @@
 package magenta.deployment_type
 
+import magenta.deployment_type.CloudFormation.{CfnParam, TagCriteria}
 import magenta.tasks.UpdateCloudFormationTask.{LookupByName, LookupByTags}
 import magenta.tasks.{CheckUpdateEventsTask, UpdateAmiCloudFormationParameterTask}
 
@@ -33,7 +34,7 @@ object AmiCloudFormationParameter extends DeploymentType {
   val amiParameter = Param[String]("amiParameter",
     documentation = "The CloudFormation parameter name for the AMI"
   ).default("AMI")
-  val amiParametersToTags = Param[Map[String, Map[String, String]]]("amiParametersToTags",
+  val amiParametersToTags = Param[Map[CfnParam, TagCriteria]]("amiParametersToTags",
     documentation =
       """AMI cloudformation parameter names mapped to the set of tags that should be used to look up an AMI.
       """.stripMargin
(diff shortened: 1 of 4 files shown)


magenta-lib/src/main/scala/magenta/deployment_type/AmiCloudFormationParameter.scala
```diff
@@ -33,6 +33,11 @@ object AmiCloudFormationParameter extends DeploymentType {
   val amiParameter = Param[String]("amiParameter",
     documentation = "The CloudFormation parameter name for the AMI"
   ).default("AMI")
+  val amiParametersToTags = Param[Map[String, Map[String, String]]]("amiParametersToTags",
```
TBonnin: I didn't find it very easy at first to understand what all those strings represent. Would it make sense to create types to help readability and, more importantly, to restrict what it can contain?

magenta-lib/src/main/scala/magenta/deployment_type/CloudFormation.scala
```diff
@@ -82,6 +89,15 @@ object CloudFormation extends DeploymentType {
...
 
+      val amiParameterMap: Map[String, Map[String, String]] = (amiParametersToTags.get(pkg), amiTags.get(pkg)) match {
+        case (Some(parametersToTags), Some(tags)) =>
+          reporter.warning("Both amiParametersToTags and amiTags supplied. Ignoring amiTags.")
+          parametersToTags
+        case (Some(parametersToTags), _) => parametersToTags
+        case (None, Some(tags)) => Map(amiParameter(pkg, target, reporter) -> tags)
+        case _ => Map.empty
+      }
+
```
TBonnin: Would be nice to avoid duplication, even though @DiegoVazquezNanini had explained to me this is quite hard to do in the current state.

magenta-lib/src/test/scala/magenta/deployment_type/CloudFormationTest.scala
```diff
@@ -63,10 +126,23 @@ class CloudFormationTest extends FlatSpec with Matchers with Inside {
...
+    val combined = UpdateCloudFormationTask.combineParameters(NamedStack("cfn"), PROD, templateParameters, Map("AMI" -> "value1", "routerAMI" -> "value2"))
+
+    combined should be(Map(
+      "AMI" -> SpecifiedValue("value1"),
+      "routerAMI" -> SpecifiedValue("value2"),
+      "Stack" -> SpecifiedValue("cfn"),
+      "Stage" -> SpecifiedValue("PROD")
+    ))
+  }
+
```
TBonnin: Isn't this test the same as `use all values on amiParametersToTags` above?

magenta-lib/src/main/scala/magenta/deployment_type/AmiCloudFormationParameter.scala
sihil: Yes - actually we did start doing that, but were struggling to get the types aligning (for another reason). Perhaps we can now introduce `CfnParam` and `TagCriteria` (or similar) types.

magenta-lib/src/main/scala/magenta/deployment_type/CloudFormation.scala
sihil: Happy for an attempt to be made. In fact it might help keep these two deployment types in sync, which wouldn't be a bad thing.

DiegoVazquezNanini:
Thanks for the review @TBonnin. I added the type aliases and removed the duplicated test.

PR #421, opened 2017-01-27
Revert "Allow updates to multiple AMI parameters"

DiegoVazquezNanini:
Reverts guardian/riff-raff#419


DiegoVazquezNanini:
This change brakes the deployments without the new amiParametersToTags param

PR #422, opened 2017-01-27
Multiple AMI parameter updates - take2

DiegoVazquezNanini:
- We need sensible defaults for amiParametersToTags for backwards compatibility reasons.
@sihil @adamnfish @TBonnin


TBonnin:
Change in 92edecfff29329097abccb3a5c5ecc533fa2abfd looks fine 👍

sihil:
As discussed, you can make parameters optional using the `optionalInYaml` field in `Param`. This retains the simpler pattern matching used to make the AMI parameter map.

DiegoVazquezNanini:
I tested this change in CODE and I can confirm that the change solved the params validation errors.

PR #424, opened 2017-02-09
Add logging in the won't retry case

sihil:
In the continued quest to vanquish #413 this adds further logging of exceptions that we won't both retrying on (clearly not catching the right thing).

### DIFF
--- a/magenta-lib/src/main/scala/magenta/tasks/AWS.scala
+++ b/magenta-lib/src/main/scala/magenta/tasks/AWS.scala
@@ -449,7 +449,12 @@ trait AWS extends Loggable {
     }
     override def shouldRetry(originalRequest: AmazonWebServiceRequest, exception: AmazonClientException, retriesAttempted: Int): Boolean = {
       val willRetry = super.shouldRetry(originalRequest, exception, retriesAttempted) || isParseException(exception)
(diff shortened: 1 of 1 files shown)


jfsoul:
@sihil Anything interesting in the logs for [this failure](https://riffraff.gutools.co.uk/deployment/view/c4f4596c-463e-4483-9926-4252387529f3) from just now?

We've seen a couple more of these in the last week or so.

sihil:
Thanks for highlighting. There is nothing in the logs at all for that exception, which makes me wonder how it isn't triggering the code path.

A deeper look at the underlying source indicates that it won't retry. The exception is thrown [here](https://github.com/aws/aws-sdk-java/blob/1.11.67/aws-java-sdk-core/src/main/java/com/amazonaws/http/AmazonHttpClient.java#L1468) and the retry site only catches IOExceptions (see [here](https://github.com/aws/aws-sdk-java/blob/1.11.67/aws-java-sdk-core/src/main/java/com/amazonaws/http/AmazonHttpClient.java#L968-L983) -
SDKClientException doesn't extend this).

So it looks suspiciously like we will need to write our own retry handling for this as it never reaches the built in evaluator. 😢

PR #425, opened 2017-02-16
Move to amigo: riff-raff.yaml and bootstrap script

TBonnin:
This patch adds a bootstrap script so riff-raff can be downloaded, installed and configured.
This was previously handled by scripts baked in the AMI by machines-images.

I think we need in the future to come up with a solution to install and configure a package we can bake in with amigo. However riff-raff, because of the self-deployment, is definitely a specific case.

I gave up on trying to package riff-raff in a `.deb`. Installing the deb package during the `ExecStartPre` step of the very same service doesn't seem to work. It hangs and finally timeouts.

You can find the related cloudformation changes at: https://github.com/guardian/deploy-tools-platform/pull/53


riff-raff/bootstrap.sh
```diff
@@ -0,0 +1,89 @@
...
+Description=${APP}
+
+[Service]
+WorkingDirectory=${HOME}
+ExecStart=${HOME}/${APP}/bin/${APP}
+ExecReload=/bin/kill -HUP $MAINPID
+Restart=on-failure
+SuccessExitStatus=143
+User=${USER}
+ExecStartPre=/etc/bootstrap.sh -s ${STAGE}
```
sihil: Not convinced by `/etc` as per comment on https://github.com/guardian/deploy-tools-platform/pull/53.

riff-raff/riff-raff.yaml
```diff
@@ -0,0 +1,17 @@
...
+    dependencies:
+      - update-ami
+  update-ami:
+    type: ami-cloudformation-parameter
+    app: riff-raff
+    parameters:
+      amiTags:
+        Recipe: xenial-java8-deploy-infrastructure
+        AmigoStage: PROD
+        BuiltBy: amigo
```
sihil: Nice - Riff-Raff itself moves onto the YAML config!

PR #429, opened 2017-03-15
Ensure region is used to filter images from prism

sihil:
Whilst looking at something with @thomaska we noticed that the query to Prism did not use the region that is provided to the `getLatestAmi` call. This has worked as we are only in one region, but will cause confusion when that is no longer the case.

This provides a fix along with a few tests to describe the behaviour.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -68,7 +68,8 @@ object Dependencies {
     filters,
     ws,
     "com.typesafe.akka" %% "akka-testkit" % "2.4.10" % Test,
-    "org.gnieh" %% "diffson" % "2.0.2" % Test
+    "org.gnieh" %% "diffson" % "2.0.2" % Test,
+    "de.leanovate.play-mockws" %% "play-mockws" % "2.5.1" % Test
   ).map((m: ModuleID) =>
     // don't even ask why I need to do this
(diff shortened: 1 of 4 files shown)


sihil:
@adamnfish does amiable respect regions?

project/Dependencies.scala
```diff
@@ -68,7 +68,8 @@ object Dependencies {
     filters,
     ws,
     "com.typesafe.akka" %% "akka-testkit" % "2.4.10" % Test,
-    "org.gnieh" %% "diffson" % "2.0.2" % Test
+    "org.gnieh" %% "diffson" % "2.0.2" % Test,
+    "de.leanovate.play-mockws" %% "play-mockws" % "2.5.1" % Test
```
TBonnin: Could you achieve the same using the built-in `play.core.server.Server` https://www.playframework.com/documentation/2.5.x/ScalaTestingWebServiceClients#extracting-setup-code

adamnfish:
@sihil Amiable is aware of the regions of the AMIs it finds and ensures that "upgrades" come from the same region. So "yes", unless you mean something else?

project/Dependencies.scala
sihil: Yes - I think I could. Didn't spot it when I read the same page earlier :)

sihil:
That's exactly what I mean @adamnfish. 💯

PR #430, opened 2017-03-22
Update AWS SDK to 1.11.106

TBonnin:
Also fixes all deprecation warnings. (Mostly how the different aws clients are being constructed)

`1.11.106` contains the fix to the `XMLStreamExceptions` issue we have been trying to fixed for a while now. See https://github.com/guardian/riff-raff/issues/413 and https://github.com/guardian/riff-raff/pull/415

https://github.com/aws/aws-sdk-java/blob/799319f670344153d588e7e42095653fabf01673/CHANGELOG.md#bugfixes

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -4,7 +4,10 @@ import Helpers._
 val commonSettings = Seq(
   organization := "com.gu",
   scalaVersion := "2.11.8",
-  scalacOptions ++= Seq("-deprecation", "-feature", "-language:postfixOps,reflectiveCalls,implicitConversions"),
+  scalacOptions ++= Seq("-deprecation", "-feature","-language:postfixOps,reflectiveCalls,implicitConversions", "-Xfatal-warnings"),
+  scalacOptions in(Compile, doc) ++= Seq(
+    "-no-link-warnings" // Suppresses problems with Scaladoc @throws links
+  ),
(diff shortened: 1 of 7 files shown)


jfsoul:
Thanks for looking at this!

How does `1.11.106` address the `XMLStreamExceptions`?  Do we still need the custom retry policy introduced in #415?

TBonnin:
@jfsoul According to `1.11.106` changelog

```
XMLStreamExceptions encountered during XML response parsing that are caused by IOException are now wrapped in IOException before being thrown so that they are passed to the configured retry policies.
```

Regarding #415 I am gonna look right now if we still need it. I didn't want to mix both changes

build.sbt
```diff
@@ -4,7 +4,10 @@ import Helpers._
 val commonSettings = Seq(
   organization := "com.gu",
   scalaVersion := "2.11.8",
-  scalacOptions ++= Seq("-deprecation", "-feature", "-language:postfixOps,reflectiveCalls,implicitConversions"),
+  scalacOptions ++= Seq("-deprecation", "-feature","-language:postfixOps,reflectiveCalls,implicitConversions", "-Xfatal-warnings"),
```
philwills: I do love to see a `-Xfatal-warnings` 

👏

PR #432, opened 2017-03-22
Log running builds when riff-raff shutdown is deferred

TBonnin:
Would log something like this:
`RiffRaff not yet inactive ('ophan::Data Lake::ophan-event-stream-etl 1413' is still running), deferring shutdown request`


TBonnin:
You are right. I will log the UDID instead. Thanks

PR #434, opened 2017-04-24
scalafmt

philwills:
I ran this manually once and then did some fix-up of the `Param` definitions as, ironically, it made these less consistent.

I'm still not sure how I feel about this in general, but it does get rid of a decent chunk of noise and I think the enforcement of a newline in longer map/filter/groupBy type chains works well.

This doesn't integrate it into any build process, just does it as a one-off.

### DIFF
--- a/magenta-lib/src/main/scala/magenta/DeploymentPackage.scala
+++ b/magenta-lib/src/main/scala/magenta/DeploymentPackage.scala
@@ -6,13 +6,12 @@ import play.api.libs.json.JsValue
 
 object DeploymentPackage {
   def apply(name: String,
-    pkgApps: Seq[App],
-    pkgSpecificData: Map[String, JsValue],
-    deploymentTypeName: String,
-    s3Package: S3Path,
-    legacyConfig: Boolean,
-    deploymentTypes: Seq[DeploymentType]
-  ): DeploymentPackage = {
+            pkgApps: Seq[App],
+            pkgSpecificData: Map[String, JsValue],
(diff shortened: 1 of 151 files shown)


magenta-lib/src/main/scala/magenta/DeploymentPackage.scala
```diff
@@ -6,28 +6,26 @@ import play.api.libs.json.JsValue
...
 }
 
-case class DeploymentPackage(
-  name: String,
-  pkgApps: Seq[App],
-  pkgSpecificData: Map[String, JsValue],
-  deploymentType: DeploymentType,
-  s3Package: S3Path,
-  legacyConfig: Boolean) {
+case class DeploymentPackage(name: String,
```
adamnfish: This syntax is an abomination

adamnfish:
I realise this is super out of date but I stumbled across it.

Lots of little fixes, some meaningful improvements and a few things worsened. I'm broadly positive about the change in general, but I'm not convinced it's worth the impact on the history.

magenta-lib/src/main/scala/magenta/DeploymentPackage.scala
philwills: Yea, I think that one might be fixable.

philwills:
Looks like HTTP4S shared your concern, but decided it was worth it (https://github.com/http4s/http4s/commit/958753a3acdff6160ff65b3e74e62a1b585af568)

PR #435, opened 2017-05-08
Deprecate deploy.json and add pauses as a forcing function for migration

sihil:
It's time to officially deprecate `deploy.json`!

A lot of people have already migrated and it is well testing. I've already created plenty of carrot with entertaining tech times, documentation and the validation tool - so now I'm suggesting a light touch stick. 

To that end I've:
 - added highlighting of deploys using deprecated features to the history screen (see below)
 - built functionality that will pause for an increasing number of seconds when deploys use `artifact.zip` or `deploy.json` (this will be 0 seconds until the 23rd of May and form then on will increase by 1 second per calendar day)

screen shot 2017-05-08 at 17 02 49

I've set an explicit end of Sept 2017 date at which point we can start massively simplifying the codebase and mention this in the deprecation warnings.

I'll send some comms out around this explaining what is going to happen and when.

I've also (@adamnfish should like this) deprecated the ability to default a bucket name in an autoscaling deploy.

### DIFF
--- a/riff-raff/app/assets/stylesheets/magenta.less
+++ b/riff-raff/app/assets/stylesheets/magenta.less
@@ -209,3 +209,7 @@ div.deployment-target:target {
 .display-inline {
   display: inline;
 }
+
+tr.has-warnings {
+  background: repeating-linear-gradient( 45deg, #fcd7d7, #fcd7d7 10px, #dddddd 10px, #dddddd 20px );
+}
--- a/riff-raff/app/conf/context.scala
+++ b/riff-raff/app/conf/context.scala
@@ -90,17 +90,22 @@ class Configuration(val application: String, val webappConfDirectory: String = "
 
   object continuousDeployment {
     lazy val enabled = configuration.getStringProperty("continuousDeployment.enabled", "false") == "true"
-    val dynamoClient = AmazonDynamoDBAsyncClientBuilder.standard()
-      .withCredentials(credentialsProviderChain(None, None))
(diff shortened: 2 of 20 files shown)


jennysivapalan:
:carrot:  -> 🏑 
👍

riff-raff/app/assets/stylesheets/magenta.less
```diff
@@ -208,4 +208,8 @@ div.deployment-target:target {
 
 .display-inline {
   display: inline;
+}
+
+tr.has-warnings {
+  background: repeating-linear-gradient( 45deg, #fcd7d7, #fcd7d7 10px, #dddddd 10px, #dddddd 20px );
```
philwills: Fancy!

riff-raff/app/conf/context.scala
```diff
@@ -190,6 +195,13 @@ class Configuration(val application: String, val webappConfDirectory: String = "
     val token = configuration.getStringProperty("mixpanel.token")
   }
 
+  object deprecation {
+    def pauseSeconds:Option[Int] = {
+      val days = new Period(new DateTime(2017,5,22,0,0,0), new DateTime()).toStandardDays.getDays
+      if (days > 0) Some(days) else None
```
philwills: I would be tempted to cap this at say 30 seconds. That seems irritating enough to cause an itch without causing screen smashing in an emergency.

sihil: Yep, that seems reasonable. Perhaps a minute though.

Issue #438, opened 2017-05-12
Deploys sometimes fail without properly cleaning up


sihil:
There have been a couple of cases where a deploy has failed but not cleaned up properly causing subsequent deploys of the same stack/stage to wait forever and the inability to deploy riff-raff itself due to it thinking there are on-going deploys.

A couple of examples:
94d5f1fe-7b44-4d8b-b11c-5be4a8716045
95f35c99-c7ce-47c5-add5-f3a6eccc7670

For the record here are some log snippets.

For `95f35c99-c7ce-47c5-add5-f3a6eccc7670`:
```text
2017-05-11 14:31:53,932 [application-akka.actor.default-dispatcher-3620] INFO  d.Deployments - Started deploying DeployParameters(Deployer(Philip McMahon),Build(Editorial Tools::Workflow Frontend,289),Stage(CODE),RecipeName(default),List(),List(),All)
2017-05-11 14:31:53,937 [application-akka.actor.default-dispatcher-3620] INFO  d.Deployments - Creating deploy record for DeployParameters(Deployer(Philip McMahon),Build(Editorial Tools::Workflow Frontend,289),Stage(CODE),RecipeName(default),List(),List(),All)
2017-05-11 14:31:53,969 [application-akka.actor.default-dispatcher-3620] DEBUG d.DeploymentEngine - Sending start deploy message to co-ordinator
2017-05-11 14:31:53,969 [deploy-akka.actor.default-dispatcher-4] DEBUG d.a.DeployCoordinator - Scheduling deploy
2017-05-11 14:31:54,037 [deploy-akka.deploy-dispatcher-1470] DEBUG d.a.DeployGroupRunner - Running next set of tasks (workflow-frontend [uploadArtifacts] => eu-west-1/workflow/0) on actor 95f35c99-c7ce-47c5-add5-f3a6eccc7670-1
2017-05-11 14:31:54,037 [deploy-akka.deploy-dispatcher-1470] DEBUG d.a.DeployGroupRunner - Running next set of tasks (workflow-frontend-ami-update [update] => eu-west-1/workflow/1) on actor 95f35c99-c7ce-47c5-add5-f3a6eccc7670-2
[...]

PR #440, opened 2017-05-24
Add more debug and ensure UUID of deploy is always logged

sihil:
more logging for tracking down #438 

/cc @mbarton @philwills


mbarton:
Thanks! 👍

PR #444, opened 2017-06-22
Don't blow up if more than a month since deprecation date

philwills:
Attempting to call `Period.toStandardDays` on a period longer than month causes `UnsupportedOperationException: Cannot convert to Days as this period contains months and months vary in length`.

`Days.daysBetween` doesn't have this limitation. 

This is already deployed as everything was blowing up.


sihil:
😳 👍

PR #447, opened 2017-07-11
Upgrade to Play 2.6

philwills:
I really tried to keep this small, but the Play 2.6 upgrade has necessitated some fairly major surgery.

 * I've kept a dependency on guardian-management, but removed the dependency on the Play specific parts of it, as they were minimal and would have required changes to upgrade
 * play-googleauth has been upgraded, which now means controllers require an `AuthAction` object as  parameter, rather than being something they extend. Some changes there mean there has been some duplication between RiffRaff and the library that I've been able to cull
 * Akka Agents are deprecated, but with no real migration path, so I'm punting that to a subsequent PR (Yay!)
 * Nice thing about these changes is that no `CIBuildPoller`/`Builds` are classes used as a dependency, we can get rid of lots of lazy val gunk and move `init` logic to constructors
 * I couldn't resist changing `Hooks` to `HooksController` as having two classes in the same code-base with the same name is way too confusing

This is currently running successfully on CODE where it has deployed itself.


riff-raff/app/controllers/Application.scala
```diff
@@ -84,19 +86,19 @@ class Application(prismLookup: PrismLookup, deploymentTypes: Seq[DeploymentType]
     Ok(views.html.index(request, documentation))
   }
 
-  def deployInfoData = AuthAction { implicit request =>
+  def deployInfoData = authAction { implicit request =>
```
sihil: Would this perhaps make a little more sense if we were to pass round an `auth` instance which is invoked with `auth.Action`?

riff-raff/app/controllers/ContinuousDeployController.scala
```diff
@@ -1,20 +1,20 @@
...
 import org.joda.time.DateTime
-
 import ci.{ContinuousDeploymentConfig, Trigger}
+import com.gu.googleauth.AuthAction
 import persistence.ContinuousDeploymentConfigRepository
 import resources.PrismLookup
 
-class ContinuousDeployController(prismLookup: PrismLookup)(implicit val messagesApi: MessagesApi, val wsClient: WSClient) extends Controller with Logging with LoginActions with I18nSupport {
+class ContinuousDeployController(prismLookup: PrismLookup, authAction: AuthAction[AnyContent], val controllerComponents: ControllerComponents)(implicit val wsClient: WSClient)
+  extends BaseController with Logging with I18nSupport {
```
sihil: We found that extending `AbstractController` was generally easier to work with when switching to 2.6, although perhaps that isn't the case if not using `Action`.

philwills: @DiegoVazquezNanini asked about this in the play-googleauth PR. My conclusion was that the difference was minimal (https://github.com/playframework/playframework/blob/master/framework/src/play/src/main/scala/play/api/mvc/Controller.scala#L168) and that this avoided a tiny bit of repetition, so was marginally preferable.

riff-raff/app/controllers/Application.scala
philwills: I don't feel strongly about it, but followed the convention I saw in the docs here: https://www.playframework.com/documentation/2.6.x/ScalaActionsComposition

riff-raff/app/controllers/ContinuousDeployController.scala
sihil: Sounds good.

riff-raff/app/controllers/Application.scala
sihil: 👍

riff-raff/app/controllers/ContinuousDeployController.scala
philwills: Yea, I do find that a bit surprising and magical.

PR #450, opened 2017-07-17
Fix a failed deploy

AWare:
This adds a link to the fix a failed deploy documentation to the Documentation menu and the deployment page when it fails. And also only has one method to shrink an ASG. 

This is so (hopefully) when your deploy fails (hopefully not) the remediation pathway is just _there_ as a helpful reminder. Useful if you're not quite sure what the really long aws command needed is. (Especially if it's 3am on 247)

note: needs testing on CODE


alexduf:
woups sorry I miss-clicked, disregard the notification Simon

sihil:
I think this is a great idea. My only question is whether we should clarify that this is aimed solely at autoscaling deploys. Whilst this is the majority, it probably helps to clarify.

AWare:
This looks a bit cluttered. 
image
So: 
image

PR #456, opened 2017-08-14
Agent => Box

philwills:
This updates to use Box (https://github.com/guardian/box). 

It's deployed to CODE and seems to work OK.

### DIFF
--- a/riff-raff/app/deployment/Deployments.scala
+++ b/riff-raff/app/deployment/Deployments.scala
@@ -3,7 +3,7 @@ package deployment
 import java.util.UUID
 
 import akka.actor.ActorSystem
-import akka.agent.Agent
+import com.gu.Box
 import akka.util.Switch
 import ci._
 import com.gu.management.DefaultSwitch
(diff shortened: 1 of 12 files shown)


riff-raff/app/deployment/Deployments.scala
```diff
@@ -157,18 +157,16 @@ class Deployments(deploymentEngine: DeploymentEngine, builds: Builds)(implicit v
 
   def cleanup(uuid: UUID) {
     log.debug(s"Queuing removal of deploy record $uuid from internal caches")
-    library sendOff { allDeploys =>
```
sihil: Why was this changed to `send` instead of `sendOff`? The later had non-blocking semantics IIRC?

philwills: `send` and `sendOff` were both non-blocking, but `sendOff` put the work on a different thread (iirc). That mean adding a lot of complexity to support it in Box, so it isn't there at the moment. This was basically seeing if the simplest thing that might possibly work did work.

PR #457, opened 2017-08-17
Some documentation for when riffraff needs to be moved between hosts.

AWare:
@philwills  Some docs about this mornings host swap for you.

Also: what do we think about changing output of `management/switchboard` to plain text to make it easier to parse curl output?

### DIFF
--- a/riff-raff/public/docs/riffraff/administration/auth.md
+++ b/riff-raff/public/docs/riffraff/administration/auth.md
@@ -1,4 +1,4 @@
-<!--- prev:properties -->
+<!--- prev:properties next:move -->
 Authentication and Authorisation
 ================================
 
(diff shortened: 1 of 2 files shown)


philwills:
Looks good to me. I'm sure @sihil will have some input when he gets back, but I think merging this is an improvement over where we are now.

sihil:
Thanks @AWare.

My normal (aka lazy) approach is to check there are no deploys running and then terminate the existing instance (then check it recovers). This approach is probably safer, but it will act weirdly (if your deploy request hits one backend then it will be able to start but if it hits the other it will not).

Happy for this to be merged though as it's not dangerous, just potentially confusing for users.

AWare:
@sihil It shouldn't, the new box isn't joined to the ELB until it's had deploys blocked. And then they're not unblocked until the old one is gone.

sihil:
You're quite right, my bad.

PR #459, opened 2017-09-05
Fix to add check for ASG Doubling

annebyrne:
Recently identified an issue when deploying services that have a `desiredSize` for ASG set to 1. If the call to double the ASG desired size fails in any way, the deployment continues as there was no previous check. This results in the ELB and ASG being compared with the not-doubled desired capacity, and no new instance being added to the ELB. Riff-Raff then attempts to cull the single instance on the ELB and fails. 

This change adds a check that the doubling has successfully completed, logs out if there has been an error and re-attempts the doubling task.

cc @mchv


sihil:
@annebyrne Have you got an example of when this has occurred (e.g. link to deploy exhibiting this behaviour)? Without knowing more detail my instinct is that this should fail fast rather than try to fix the problem but happy to be convinced otherwise.

annebyrne:
Yes! This was the most recent one: https://riffraff.gutools.co.uk/deployment/view/684e0b13-3588-418e-8189-731fc276b381

Typically re-running the deploy solved the issue.

annebyrne:
I've closed this as the issue is actually due to an underlying race condition caused by a missing dependancy when the templates are applied.

PR #460, opened 2017-09-12
Add deploy mechanism when you know region, stack, stage and app

sihil:
If you are armed with the region, stack, stage and app (RSSA) of an instance (say from AMIable telling you an AMI is out of date) then this provides a mechanism for locating the appropriate deploy.

There are two parts. 

1. A target resolver that
 - watches for new builds
 - resolves the deployment graph for each new build to figure out the SSA that it targets and stores the results in a dynamo table
2. a little frontend that
 - takes a RSSA
 - looks it up in the dynamo table
 - guides the user in selecting the best fit

The RSSA might be ambiguous, if so the user will be shown the options (this is ambiguous because of an error in a file):
screen shot 2017-09-12 at 13 02 56

Next, the most recent deploys are shown and the user is asked which should be redeployed:
screen shot 2017-09-12 at 13 03 04

Finally, the confirmation screen that we all know and love:
screen shot 2017-09-12 at 13 03 15

## Considerations
This will not work for projects that do not have recent builds. I did look at resolving the targets during deployment and storing them only for successful deploys but this made the code far more complicated.

## Todo
 - [x] At the moment the poop emoji 💩 appears in the URL for the second of these stages. Whist it was entertaining to use it as a separator in an internal key, this should probably be changed to another character

### DIFF
--- a/riff-raff/app/ci/TargetResolver.scala
+++ b/riff-raff/app/ci/TargetResolver.scala
@@ -1,0 +1,52 @@
+package ci
+
+import cats.data.Validated
+import conf.Configuration
+import controllers.Logging
+import lifecycle.Lifecycle
+import magenta.Build
+import magenta.artifact._
+import magenta.deployment_type.DeploymentType
+import magenta.graph.Graph
+import magenta.input.{All, Deployment}
+import magenta.input.resolver.Resolver
+import persistence.TargetDynamoRepository
+
+case class Target(region: String, stack: String, app: String)
+
+class TargetResolver(ciBuildPoller: CIBuildPoller, deploymentTypes: Seq[DeploymentType]) extends Lifecycle with Logging {
+  val poller = ciBuildPoller.newBuilds.subscribe { build =>
+    for {
+      yaml <- Validated.fromEither(fetchYaml(build.jobName, build.jobId))
+      deployGraph <- Resolver.resolveDeploymentGraph(yaml._2, deploymentTypes, All)
+      targets = extractTargets(deployGraph)
+    } {
+      targets.foreach{ t =>
+        TargetDynamoRepository.set(t, build.jobName)
+      }
+    }
+  }
+
+  def fetchYaml(name: String, id: String): Either[S3Error, (S3Path, String)] = {
+    val build = Build(name, id)
+    val artifact = S3YamlArtifact(build, Configuration.artifact.aws.bucketName)
+    val deployObjectPath = artifact.deployObject
+    val deployObjectContent = S3Location.fetchContentAsString(deployObjectPath)(Configuration.artifact.aws.client)
+    deployObjectContent.map(deployObjectPath -> _)
+  }
+
+  def extractTargets(graph: Graph[Deployment]): Set[Target] = {
+    graph.nodes.values.flatMap { deployment =>
+      for {
+        region <- deployment.regions.toList
+        stack <- deployment.stacks.toList
+      } yield Target(region, stack, deployment.app)
+    }
+  }
+
+  override def init() = {}
+
+  override def shutdown() = {
+    poller.unsubscribe()
+  }
+}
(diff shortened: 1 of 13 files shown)


sihil:
If anyone has any insight as to why my local build works, but not the one on TC then I'd be interested in hearing.

riff-raff/app/persistence/TargetDynamoRepository.scala
```diff
@@ -0,0 +1,35 @@
...
+import com.gu.scanamo.Table
+import org.joda.time.DateTime
+
+case class TargetId(id: String, targetKey: String, region: String, stack: String, app: String, projectName: String, lastSeen: DateTime)
+object TargetId {
+  def apply(tgt: Target, projectName: String, lastSeen: DateTime): TargetId =
+    TargetId(id(tgt, projectName), targetKey(tgt), tgt.region, tgt.stack, tgt.app, projectName, lastSeen)
+  /* Concatenating to make a composite key is pretty 💩, so using 💩 as a separator. #UTF8FTW */
+  def targetKey(tgt: Target) = Seq(tgt.region,tgt.stack,tgt.app).mkString("💩")
+  def id(tgt: Target, projectName: String) = Seq(tgt.region,tgt.stack,tgt.app,projectName).mkString("💩")
```
philwills: Why does the `projectName` need to be part of the key? 

It at least feels separate enough that it could be a range key, rather than being part of the hash key, so that you could not have the separate `targetKey` field.

sihil: It has the benefit that if the same RSSA is seen again then it will effectively be upserted and the `lastSeen` will be updated due to the uniqueness constraint.

Happy to have that Dynamo conversation if you're around.

riff-raff/app/ci/TargetResolver.scala
```diff
@@ -0,0 +1,68 @@
...
+import magenta.input.resolver.Resolver
+import persistence.TargetDynamoRepository
+
+case class Target(region: String, stack: String, app: String)
+
+class TargetResolver(ciBuildPoller: CIBuildPoller, deploymentTypes: Seq[DeploymentType]) extends Lifecycle with Logging {
+  val poller = ciBuildPoller.newBuilds.subscribe { build =>
+    val result = for {
+      yaml <- fetchYaml(build)
+      deployGraph <- Resolver.resolveDeploymentGraph(yaml._2, deploymentTypes, All).toEither
```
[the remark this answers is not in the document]
sihil: Yeah. Unfortunately Either doesn't implement `withFilter`.

sihil:
@philwills Think I've addressed everything.

Note that https://github.com/guardian/deploy-tools-platform/pull/68 needs to be applied to the PROD CFN stack before merging.

riff-raff/app/persistence/TargetDynamoRepository.scala
```diff
@@ -0,0 +1,38 @@
...
+  import com.gu.scanamo.syntax._
+
+  def set(target: Target, projectName: String, lastSeen: DateTime): PutItemResult =
+    exec(table.put(TargetId(target, projectName, lastSeen)))
+
+  def get(targetKey: String, projectName: String): Option[TargetId] = {
+    exec(table.get('targetKey -> targetKey and 'projectName -> projectName)).flatMap(_.toOption)
+  }
+
+  def get(target: Target): List[TargetId] = {
```
philwills: (nit) I'd prefer if this had a different name, rather than being overloaded

sihil: `get`, `Target` or `TargetId`?

sihil: Oh - `get` is overloaded now isn't it. OK, I'll have a look.

PR #462, opened 2017-09-13
Add some better JVM options

sihil:
Riff-Raff was using the default JVM options - this improves the situation significantly by adding some sensible memory allocation as well as fixing the old RUNNING_PID file issue.


mchv:
:+1:

philwills:
😱 🙀 😱

PR #463, opened 2017-09-14
Add more troubleshooting detail for when deploys fail due to dependancy clash

annebyrne:
If dependancies for certain deployment types are not stated in `riff-raff.yaml`, a race condition will be introduced that writes over the desiredSize for the ASG - causing the deploy to fail. This PR adds more detail to the docs and the error to give more clarity.

[Example failed deploy](https://riffraff.gutools.co.uk/deployment/view/684e0b13-3588-418e-8189-731fc276b381)

### DIFF
--- a/magenta-lib/src/main/scala/magenta/deployment_type/Lambda.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/Lambda.scala
@@ -16,6 +16,11 @@ object Lambda extends DeploymentType  {
       |
       |It is recommended to use the `bucket` parameter as storing the function code in S3 works much better when using
       |cloudformation.
+      |
+      |Ensure to add any relevant dependencies to your deploy in your riff-raff.yaml to guard against race conditions,
+      |such as a Riff-Raff Cloudformation update modifying your lambda at the same time this deployment type is running.
+      |
+      |
       """.stripMargin
 
   val regionsParam = Param[List[String]]("regions",
--- a/magenta-lib/src/main/scala/magenta/tasks/ASGTasks.scala
+++ b/magenta-lib/src/main/scala/magenta/tasks/ASGTasks.scala
@@ -112,8 +112,17 @@ case class CullInstancesWithTerminationTag(pkg: DeploymentPackage, stage: Stage,
     implicit val elbClient = ELB.client(keyRing, region)
     val instancesToKill = asg.getInstances.filter(instance => EC2.hasTag(instance, "Magenta", "Terminate", ec2Client))
     val orderedInstancesToKill = instancesToKill.transposeBy(_.getAvailabilityZone)
-    reporter.verbose(s"Culling instances: ${orderedInstancesToKill.map(_.getInstanceId).mkString(", ")}")
-    orderedInstancesToKill.foreach(instance => ASG.cull(asg, instance, asgClient, elbClient))
+    try {
+      reporter.verbose(s"Culling instances: ${orderedInstancesToKill.map(_.getInstanceId).mkString(", ")}")
+      orderedInstancesToKill.foreach(instance => ASG.cull(asg, instance, asgClient, elbClient))
+    } catch {
+      case e: AmazonServiceException if desiredSizeReset(e) => {
+        reporter.warning("Your ASG desired size may have been reset. This may be because two parts of the deploy are attempting to modify a cloudformation stack simultaneously. Please check that appropriate dependencies are included in riff-raff.yaml, or ensure desiredSize isn't set in the Cloudformation.")
+        throw new ASGResetException(e)
+      }
+    }
+
+    def desiredSizeReset(e: AmazonServiceException) = e.getStatusCode == 400 && e.getErrorCode == "ValidationError"
   }
 
   lazy val description = "Terminate instances with the termination tag for this deploy"
@@ -136,6 +145,8 @@ case class ResumeAlarmNotifications(pkg: DeploymentPackage, stage: Stage, stack:
 
   lazy val description = "Resuming Alarm Notifications - group will scale on any configured alarms"
 }
+
+class ASGResetException(val throwable: Throwable) extends Throwable("", throwable)
 
 trait ASGTask extends Task {
   def region: Region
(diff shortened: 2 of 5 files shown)


magenta-lib/src/main/scala/magenta/deployment_type/Lambda.scala
```diff
@@ -16,6 +16,9 @@ object Lambda extends DeploymentType  {
       |
       |It is recommended to use the `bucket` parameter as storing the function code in S3 works much better when using
       |cloudformation.
+      |
+      |You will need to add this as a dependency to your autoscaling deploy in your riff-raff.yaml to guard against race conditions.
```
philwills: I'm unclear about how a `Lambda` deploy could cause problems for an `autoscaling` deploy. Do you have an example?

sihil: Ah - lambda can conflict with cloudformation in the same way as autoscaling can conflict with cloudformation. lambda shouldn't be able to conflict with autoscaling.

annebyrne: Thanks! Will fix the typo

annebyrne:
Is this good to go @sihil ?

magenta-lib/src/main/scala/magenta/tasks/ASGTasks.scala
```diff
@@ -112,8 +112,16 @@ case class CullInstancesWithTerminationTag(pkg: DeploymentPackage, stage: Stage,
...
+    try {
+      reporter.verbose(s"Culling instances: ${orderedInstancesToKill.map(_.getInstanceId).mkString(", ")}")
+      orderedInstancesToKill.foreach(instance => ASG.cull(asg, instance, asgClient, elbClient))
+    } catch {
+      case e: AmazonServiceException if desiredSizeReset(e) => {
+        reporter.info(s"${e.getMessage}. Your ASG desired size may have been reset. Please check that appropriate dependencies are included in riff-raff.yaml, or ensure desiredSize isn't set in the Cloudformation.")
+      }
+    }
+
+    def desiredSizeReset(e: AmazonServiceException) = e.getStatusCode == 400 && e.getErrorCode == "ValidationError"
```
sihil: This seems pretty vague (ASE tend to be 400 ValidationError). I tend to match on the ErrorMessage, but that's brittle. Given that we are in a given task this vagueness seems reasonable.

magenta-lib/src/main/scala/magenta/tasks/ASGTasks.scala
```diff
@@ -112,8 +112,16 @@ case class CullInstancesWithTerminationTag(pkg: DeploymentPackage, stage: Stage,
...
     val instancesToKill = asg.getInstances.filter(instance => EC2.hasTag(instance, "Magenta", "Terminate", ec2Client))
     val orderedInstancesToKill = instancesToKill.transposeBy(_.getAvailabilityZone)
-    reporter.verbose(s"Culling instances: ${orderedInstancesToKill.map(_.getInstanceId).mkString(", ")}")
-    orderedInstancesToKill.foreach(instance => ASG.cull(asg, instance, asgClient, elbClient))
+    try {
+      reporter.verbose(s"Culling instances: ${orderedInstancesToKill.map(_.getInstanceId).mkString(", ")}")
+      orderedInstancesToKill.foreach(instance => ASG.cull(asg, instance, asgClient, elbClient))
+    } catch {
+      case e: AmazonServiceException if desiredSizeReset(e) => {
+        reporter.info(s"${e.getMessage}. Your ASG desired size may have been reset. Please check that appropriate dependencies are included in riff-raff.yaml, or ensure desiredSize isn't set in the Cloudformation.")
```
sihil: The message could more explicitly talk about the potential root cause being that two parts of the deploy are both trying to modify a cloudformation stack simultaneously. At the moment a solution is suggested without the context of what the problem is.

Also: note that `e` is being swallowed here so the deploy will no longer fail; I think it makes sense to either:
 - report a message (maybe warn instead of info) and then re-throw the exception
 - throw a new exception with this message (nesting the original exception for completeness)

magenta-lib/src/main/scala/magenta/deployment_type/Lambda.scala
```diff
@@ -16,6 +16,9 @@ object Lambda extends DeploymentType  {
       |
       |It is recommended to use the `bucket` parameter as storing the function code in S3 works much better when using
       |cloudformation.
+      |
+      |Ensure to add any relevant dependencies to your deploy in your riff-raff.yaml to guard against race conditions.
```
sihil: Maybe add something like `(such as a Riff-Raff Cloudformation update modifying your lambda at the same time this deployment type is running)` to more explicitly outline the potential pitfall.

annebyrne:
Hi @sihil I've made changes you suggested, I've left the check as it is as its contained within the task so should be OK to be pretty vague. LMK what you think! :+1:

annebyrne:
@sihil Is this good to go following changes requested?

magenta-lib/src/main/scala/magenta/tasks/ASGTasks.scala
```diff
@@ -137,6 +146,8 @@ case class ResumeAlarmNotifications(pkg: DeploymentPackage, stage: Stage, stack:
   lazy val description = "Resuming Alarm Notifications - group will scale on any configured alarms"
 }
 
+class ASGResetException(val throwable: Throwable) extends Throwable("", throwable)
```
sihil: A better signature for an exception would be (I don't think you need the val and a message is important):
```
class ASGResetException(msg: String, cause: Throwable) extends Throwable(msg, cause)
```

magenta-lib/src/main/scala/magenta/tasks/ASGTasks.scala
```diff
@@ -112,8 +112,17 @@ case class CullInstancesWithTerminationTag(pkg: DeploymentPackage, stage: Stage,
...
     val orderedInstancesToKill = instancesToKill.transposeBy(_.getAvailabilityZone)
-    reporter.verbose(s"Culling instances: ${orderedInstancesToKill.map(_.getInstanceId).mkString(", ")}")
-    orderedInstancesToKill.foreach(instance => ASG.cull(asg, instance, asgClient, elbClient))
+    try {
+      reporter.verbose(s"Culling instances: ${orderedInstancesToKill.map(_.getInstanceId).mkString(", ")}")
+      orderedInstancesToKill.foreach(instance => ASG.cull(asg, instance, asgClient, elbClient))
+    } catch {
+      case e: AmazonServiceException if desiredSizeReset(e) => {
+        reporter.warning("Your ASG desired size may have been reset. This may be because two parts of the deploy are attempting to modify a cloudformation stack simultaneously. Please check that appropriate dependencies are included in riff-raff.yaml, or ensure desiredSize isn't set in the Cloudformation.")
+        throw new ASGResetException(e)
```
sihil: Giving a message to this exception is of value.

philwills:
I've addressed @sihil's concerns, so unless I hear any other objection, I'll get this merged this afternoon.

sihil:
👍

annebyrne:
Thanks! Sorry for dropping the ball on merging this.

PR #464, opened 2017-09-15
Cloudformation update improvements

sihil:
A couple of improvement to CFN updates I noticed whilst working on some AMI updates.

1. If the update fails and rolls back is doesn't always report a FAILED status. Think that it should also fail if any status is seen with ROLLBACK in.
2. If something else is currently updating the stack we should wait and try again shortly. This uses similar logic to the check, but focused around retrying.

We usually don't retry during deploys and instead fail fast. I think it makes sense to retry here as the locking is being dealt with elsewhere so we don't need to concern ourselves with it.

The use case is where a project has individual builds/deploys going to multiple ASGs in a single CFN stack. Each deployable is configured to update a parameter that controls the AMI for the instances it controls. It makes sense that you might deploy multiple parts of the same stack at once and it would be safe to do so.

I've only added the logic to the parameter update rather than where we actually update a template which would have more serious race condition concerns.

Thoughts welcome.

### DIFF
--- a/magenta-lib/src/main/scala/magenta/tasks/UpdateCloudFormationTask.scala
+++ b/magenta-lib/src/main/scala/magenta/tasks/UpdateCloudFormationTask.scala
@@ -290,7 +290,7 @@ case class CheckUpdateEventsTask(
     def updateComplete(stackName: String)(e: StackEvent): Boolean =
       isStackEvent(stackName)(e) && (e.getResourceStatus == "UPDATE_COMPLETE" || e.getResourceStatus == "CREATE_COMPLETE")
 
-    def failed(e: StackEvent): Boolean = e.getResourceStatus.contains("FAILED")
+    def failed(e: StackEvent): Boolean = e.getResourceStatus.contains("FAILED") || e.getResourceStatus.contains("ROLLBACK")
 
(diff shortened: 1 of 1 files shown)


magenta-lib/src/main/scala/magenta/tasks/UpdateCloudFormationTask.scala
```diff
@@ -1,18 +1,58 @@
...
-import magenta.{DeployReporter, DeployTarget, DeploymentPackage, KeyRing, Region, Stack, Stage}
+import magenta.{DeploymentPackage, DeployReporter, DeployTarget, KeyRing, Region, Stack, Stage}
 import org.joda.time.{DateTime, Duration}
 
 import scala.collection.convert.wrapAsScala._
 
+/**
+  * A simple trait to aid with attempting an update multiple times in the case that an update is already running.
+  */
+trait RetryCloudFormationUpdate {
```
philwills: Superficially, this looks a lot like `RepeatedPollingCheck`. Is that a real similarity, or are there some fundamental differences?

sihil: My initial version did use Repeated Polling check. But it can return a value and isn't actually a check. I concluded there was value in making a custom variant but could be convinced otherwise.

PR #468, opened 2017-10-13
Disable support for artifact.zip / deploy.json

sihil:
There have been very few deploys that rely on the old formats in the last month. This fulfils the statement in the deprecation messages that support would be removed at the end of September.


sihil:
I'm going to leave this for the weekend. Will merge first thing on Monday morning.

PR #471, opened 2017-11-03
Autoscaling: wait for stabilization rather than initial single check

TBonnin:
Riff-raff currently executes a initial single check to verify the ASG and ELB have the same number of instances.
In case the ASG is currently scaling up/down when ASG deploy is being started, the check fails and the deploy is terminated.
From now on Riff-raff will wait for the ASG to stabilize (for a short
period of time) before to fail the deploy rather than making a single check.
It shouldn't impact the deployment speed if the ASG is already in a stabilised state but it will help apps like frontend when scaling events happen on a regular basis during the day.

I picked `1 minutes` as the initial timeout so riff-raff doesn't wait to much in case something is wrong but still give a chance to the ASG to finally stabilised. Let me know what you think @sihil

### DIFF
--- a/magenta-lib/src/main/scala/magenta/deployment_type/AutoScaling.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/AutoScaling.scala
@@ -102,7 +102,7 @@ object AutoScaling  extends DeploymentType {
     val parameters = target.parameters
     val stack = target.stack
     List(
-      CheckForStabilization(pkg, parameters.stage, stack, target.region),
+      WaitForStabilization(pkg, parameters.stage, stack, 5 * 60 * 1000, target.region),
       CheckGroupSize(pkg, parameters.stage, stack, target.region),
       SuspendAlarmNotifications(pkg, parameters.stage, stack, target.region),
       TagCurrentInstancesWithTerminationTag(pkg, parameters.stage, stack, target.region),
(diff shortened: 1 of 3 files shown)


NataliaLKB:
Great! This should definitely help deployments on frontend 👍

NataliaLKB:
It just happened again: https://riffraff.gutools.co.uk/deployment/view/8caf8016-adf1-4c4d-b377-254e1bcc787e

alexduf:
It's a good idea, I've ran into that problem as well.
However, shouldn't the delay be slightly more than 1 minute? Scaling up take roughly 3 minutes on MAPI, I'd be surprised if frontend was scaling in less than a minute too.

TBonnin:
I am happy to tweak the timeout. Maybe 1 minute is too short

alexduf:
I'm actually wondering if we shouldn't use `warmupGrace` here?
As defined in the doc https://riffraff.gutools.co.uk/docs/magenta-lib/types

warmupGrace | Number of seconds to wait for the instances in the load balancer to warm up

Or maybe defining a new property defaulted at 0, that we (Frontend and MAPI) can override to something else?

TBonnin:
Change the delay to 5mins.

@sihil what do you think of this PR?

alexduf:
Thanks @TBonnin !

mbarton:
@TBonnin build statuses are back now for RiffRaff builds. Looks like there's some test failures

PR #476, opened 2018-01-17
Scheduled deploys

sihil:
Pulled this together this afternoon with @nicl 

This introduces scheduled deploys to Riff-Raff. This allows you to configure a project to deploy automatically according to a pre-determined schedule.

The aim of scheduled deploys is to have a mechanism for automatically updating AMIs even when a project is not actively being worked on.

Some details:
 - schedules are expressed as quartz cron expressions (as per AMIgo)
 - a scheduled deploy will re-deploy the most recent deploy to the specified stage
 - a scheduled deploy is skipped if there is no previous deploy or the most recent deploy failed or is still running

Possible future improvements:
 - only deploy if the most recent deploy was X days ago
 - something other than quartz crons for configuring

Scheduling screen:
screen shot 2018-01-17 at 18 26 00

Deploy is run by a "scheduled deploy" user:
screen shot 2018-01-17 at 18 26 36

### DIFF
--- a/riff-raff/app/views/schedule/form.scala.html
+++ b/riff-raff/app/views/schedule/form.scala.html
@@ -1,0 +1,41 @@
+@(configForm: Form[controllers.ScheduleController.ScheduleForm], prismLookup: resources.PrismLookup)(implicit request: Security.AuthenticatedRequest[AnyContent, com.gu.googleauth.UserIdentity], messages: Messages)
+@import b3.vertical.fieldConstructor
+@import helper.CSRF
+
+@main("Schedule", request, List("form-autocomplete")) {
+
+    <h2>Schedule Configuration</h2>
+    <hr/>
+
+    @b3.form(action=routes.ScheduleController.save) {
+        @CSRF.formField
+        @snippets.inputHidden(configForm("id"))
+
+        @if(configForm.hasGlobalErrors) {
+            <div class="alert alert-danger">
+                <h4>Error</h4>
+                <ul>
+                @configForm.globalErrors.map { error =>
+                    <li>@error.message</li>
+                }
+                </ul>
+            </div>
+        }
+
+        @b3.text(configForm("projectName"), 'id -> "projectInput", Symbol("data-url") -> "/deployment/request/autoComplete/project", '_label -> "Project Name")
+        @b3.select(
+            configForm("stage"),
+            options = helper.options(prismLookup.stages.toList),
+            '_default -> "--- Choose a stage ---",
+            '_label -> "Stage",
+            '_error -> configForm.globalError.map(_.withMessage("Please select deployment stage"))
+        )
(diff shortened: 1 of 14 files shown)


riff-raff/app/views/schedule/form.scala.html
```diff
@@ -0,0 +1,44 @@
...
+        @b3.text(configForm("projectName"), 'id -> "projectInput", Symbol("data-url") -> "/deployment/request/autoComplete/project", '_label -> "Project Name")
+        @b3.select(
+            configForm("stage"),
+            options = helper.options(prismLookup.stages.toList),
+            '_default -> "--- Choose a stage ---",
+            '_label -> "Stage",
+            '_error -> configForm.globalError.map(_.withMessage("Please select deployment stage"))
+        )
+        @b3.text(configForm("schedule"), '_label -> "Schedule expression")
+        <p>This should be a weird Quartz cron expression
```
adamnfish: Be good to give a (couple of) example(s) here. Most teams will be doing similar things and the docs are pretty dense.

sihil: Done. Think one is enough.

davidfurey:
I think it would be useful for this to default to excluding bank holidays.  Avoiding Monday and Friday fixes for most, but there is still Christmas day, Boxing Day and New Years day.

Also, is the time UTC or Europe/London?

sihil:
Bank holiday exclusions can certainly be added later if desired but isn't something I'm interested in for the first cut.

We should clarify the behaviour. My expectation would be that it is local time, but I suspect it relies on the default JVM timezone and is likely to be GMT.

nicl:
@adamnfish @davidfurey this is a first pass so things like bank holiday etc. good to add once we know it's valuable. 

We are going for a soft launch (i.e. probably not shouting about it at first). I'm planning on testing on Discussion once merged for a while as it's a good candidate project - several apps which hardly ever get deployed.

good question about timezone. I agree with @sihil let's check, but it's almost certainly GMT.

riff-raff/app/views/schedule/form.scala.html
nicl: There are a few traps we talked about here, which could happen by accident or deliberately:

* people scheduling it too frequently
* scheduling overnight/weekends
* everyone scheduling at the same time

If the feature proves useful, we'll probably need to address some or all of these.

sihil:
As guessed, Riff-Raff currently runs in UTC.

I've just pushed another commit that allows the user to explicitly select the timezone that the schedule should run in. It defaults to the timezone of the users browser (provided they have a reasonably modern browser).

When the quartz schedule is configured we provide the timezone. Seems to work based on a brief test in Kiev.

sihil:
I think this is ready to ship. Would be glad of a review / 👍whenever someone has time.

nicl:
@sihil I can 👍 but given we paired on it, another would also be good.

PR #479, opened 2018-01-22
Add support for AMIgo based encrypted AMIs

sihil:
With https://github.com/guardian/amigo/pull/164 it has become necessary to add a little more functionality to Riff-Raff to correctly identify a suitable AMI (encrypted AMIs must be in the same account as they are used). A new boolean parameter `amiEncrypted` has been introduced to support this.

Most importantly we don't want Riff-Raff to inadvertently pick up an encrypted AMI when one had not been requested. If it tried to do this and picked an AMI from a different account it would not work. At all. As such, when `amiEncrypted=false` (the default) we only consider AMIs that either have no `Encrypted` tag or those that have an `Encrypted=false` tag.

When we set `amiEncrypted=true` Riff-Raff does two things:
 - only considers owned by the AWS account ID belonging to the AWS credentials being used to update the CFN stack - this is done by adding a `meta.origin.accountNumber` filter to the prism request
 - adds a tag filter of `Encrypted=true` which is set by the AMIgo copy lambda - this can be overridden by specifying the tag explicitly but for convenience is set to `true`

### DIFF
--- a/magenta-lib/src/main/scala/magenta/Lookup.scala
+++ b/magenta-lib/src/main/scala/magenta/Lookup.scala
@@ -21,7 +21,7 @@ trait Lookup {
   def stages: Seq[String]
   def data: DataLookup
   def keyRing(stage: Stage, apps: Set[App], stack: Stack): KeyRing
-  def getLatestAmi(region: String)(tags: Map[String, String]): Option[String]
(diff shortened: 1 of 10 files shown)


tomrf1:
👍 looks good!
We're going to try out capi's elasticsearch with encrypted root volume using the new amigo feature. (We don't deploy elasticsearch using riffraff, so this change isn't required for our purposes)

PR #483, opened 2018-03-01
Expose build duration in seconds

nicl:
I want to ingest this into a Kibana stack we have so that we can track our deploy times over time.

**Note: not yet tested other than it compiles.**


nicl:
Riffraff can deploy itself in 20 seconds! 😲 If only our other projects were this quick.

screen shot 2018-03-01 at 11 57 16

PR #484, opened 2018-03-12
Use more powerful IAM capability for CF updates

adamnfish:
CAPABILITY_NAMED_IAM is a strictly more powerful than CAPABILITY_IAM. That extra power is something we need occasionally but it is fine to always pass it.

### DIFF
--- a/magenta-lib/src/main/scala/magenta/tasks/AWS.scala
+++ b/magenta-lib/src/main/scala/magenta/tasks/AWS.scala
@@ -310,7 +310,7 @@ object CloudFormation extends AWS {
   case class TemplateBody(body: String) extends Template
   case class TemplateUrl(url: String) extends Template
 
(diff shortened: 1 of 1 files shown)


johnduffell:
ok, but CAPABILITY_IAM = CAPABILITY_NAMED_IAM?
consider rename the variable too?

adamnfish:
Yeah, would be better that way. I'll have to open it in IntelliJ to do ti that way which takes a bit longer is all.

adamnfish:
@johnduffell fixed, thanks

johnduffell:
💥 🚀 any time 😉 👍

sihil:
@adamnfish Should this be merged and pushed out?

adamnfish:
Almost certainly!

I lack the access and skills to do so though. Happy to work with someone that knows Riff Raff to do this, but it probably won't be before the GDPR deadline for me.

PR #485, opened 2018-03-20
Scheduled deploy cooldown

mbarton:
Adds an optional "Cool-down Days" period to scheduled launch. If a project has been deployed in the last N days then the scheduled deploy is skipped.

The use case here is for critical yet regularly deployed apps like Composer. The chances are the project has been deployed recently and every scheduled deploy is risk (albeit small risk). As such we can achieve GDPR compliance by scheduling a deploy every week unless it has been deployed in the last 7 days.

I'd especially welcome feedback about the time ranges involved (eg are days the right unit?) and if there's interplay between the quartz expression and cool-down period that would be dangerous enough to require guards.

### DIFF
--- a/riff-raff/app/controllers/ScheduleController.scala
+++ b/riff-raff/app/controllers/ScheduleController.scala
@@ -9,7 +9,7 @@ import org.quartz.CronExpression
 import persistence.ScheduleRepository
 import play.api.data.Form
 import play.api.data.Forms._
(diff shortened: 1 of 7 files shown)


mbarton:
We've been running without this just fine for a couple of months now so I'm closing

nicl:
@mbarton ah sorry - only just saw this. If people decide this is useful in future then shout as this looks like good work!

PR #487, opened 2018-04-03
Remove legacy code

sihil:
This PR:
 - removes deploy.json support 
 - removes artifact.zip support
 - removes concepts of being able to control a deploy using a specific recipe or subset of hosts
 - makes `Stack` mandatory everywhere
 - change from `Set[App]` to `App`  (so there is only one App most of the time)
 - legacy documentation

This is a large PR and pretty much impossible to meaningfully review. I'd suggest that we try this in CODE and then give users a heads up before rolling out into PROD.


magenta-lib/src/main/scala/magenta/artifact/S3Location.scala
```diff
@@ -96,75 +93,6 @@ object S3Artifact {
...
-      artifact.deployObject.fetchContentAsString()(client)
-    }
-    possibleJson.leftMap[ArtifactResolutionError](S3ArtifactError(_)).flatMap(s =>
-      JsonInputFile.parse(s).leftMap(JsonArtifactError(_))
-    )
-  }
-
-  def convertFromZipBundle(artifact: S3JsonArtifact, deprecatedPause: Option[Int])(implicit client: AmazonS3, reporter: DeployReporter): Unit = {
-    reporter.fail(
-      """NO LONGER SUPPORTED: The artifact.zip format is no longer supported - please switch to the new format (if you
```
johnduffell: good! that error was annoying because it usually meant I messed something unrelated up, but it didn't say what 🥇

johnduffell:
image
👍 💯

Issue #482, opened 2018-02-20
'Uploaded 0 assets' warning should be more prominent


philmcmahon:
The 'frontend-static' step of this build: https://riffraff.gutools.co.uk/deployment/view/3d0241fe-411a-48e9-97ee-3424a85a3585 has a warning on it, but where the warning comes from is hard to tell until you open up the different build steps. For theguardian.com not having any assets is quite sad, so it would be good to make it immediately obvious what the problem is.

katebee:
Fixed by #489 🎉

PR #489, opened 2018-04-11
Fail the deploy fast when zero s3 artefacts are uploaded

katebee:
Fix for https://github.com/guardian/riff-raff/issues/482

### DIFF
--- a/magenta-lib/src/main/scala/magenta/tasks/tasks.scala
+++ b/magenta-lib/src/main/scala/magenta/tasks/tasks.scala
@@ -58,7 +58,7 @@ case class S3Upload(
         case (path: S3Path, _) => path.show()
         case (location, _) => location.toString()
       }).mkString("\n")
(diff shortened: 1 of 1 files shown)


AWare:
We appropriately vandalised a build: https://riffraff.code.dev-gutools.co.uk/deployment/view/73783f37-cb50-46e8-ab50-6cf41f1f0f43

AWare:
@sihil is there a valid reason for a build to attempt to upload 0 artefacts?

sihil:
If that would imply that it failed to even upload a `riff-raff.yaml` then absolutely not.

If that were to exempt a `riff-raff.yaml` I suspect it would still make no sense although there might be an exemption to that rule that I can't think of.

sihil:
I realised I was assuming that this was to do with sbt-riffraff-artifact failing the build. Actually we mean failing the deploy here and I've updated the title of the PR to reflect this.

PR #490, opened 2018-04-17
Deal with invalid date time format and log build.json parsing errors

sihil:
We've seen a `build.json` with a badly formatted date. This should deal more gracefully with that and provide logging to track down the offending file.


jacobwinch:
👍

PR #492, opened 2018-05-04
Check status of multiple classic and V2 ELBs

tomrf1:
This change makes riffraff check that all ELBs in a stack are healthy.
This also fixes an issue where a stack with both classic and V2 ELBs fails to deploy.

### DIFF
--- a/magenta-lib/src/main/scala/magenta/tasks/AWS.scala
+++ b/magenta-lib/src/main/scala/magenta/tasks/AWS.scala
@@ -27,6 +27,8 @@ import com.amazonaws.services.securitytoken.model.GetCallerIdentityRequest
 import com.gu.management.Loggable
 import magenta.{App, DeploymentPackage, DeployReporter, KeyRing, Region, Stack, Stage}
 
+import cats.implicits._
+
 import scala.annotation.tailrec
 import scala.collection.JavaConverters._
 import scala.util.Try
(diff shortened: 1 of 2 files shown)


magenta-lib/src/main/scala/magenta/tasks/AWS.scala
```diff
@@ -134,30 +139,36 @@ object ASG extends AWS {
...
 
-  def elbName(asg: AutoScalingGroup) = asg.getLoadBalancerNames.asScala.headOption
+  def elbNames(asg: AutoScalingGroup): List[String] = asg.getLoadBalancerNames.asScala.toList
 
-  def elbTargetArn(asg: AutoScalingGroup) = asg.getTargetGroupARNs.asScala.headOption
+  def elbTargetArns(asg: AutoScalingGroup): List[String] = asg.getTargetGroupARNs.asScala.toList
 
   def cull(asg: AutoScalingGroup, instance: ASGInstance, asgClient: AmazonAutoScaling, elbClient: ELB.Client) = {
-    ELB.deregister(elbName(asg), elbTargetArn(asg), instance, elbClient)
+    ELB.deregister(elbNames(asg).headOption, elbTargetArns(asg).headOption, instance, elbClient)
```
sihil: A brief look at `ELB.deregister` indicates that the first two params would nicely change from `Option` to `List` which might also be more 'correct'.

PR #499, opened 2018-07-18
[WIP] Enable lookup of Lambdas by tag when updating

sihil:
This is a work in progress PR to enable deploying to lambdas that are identified using tags.

At present none of the Riff-Raff users have the right permissions, so this will not work.


sihil:
Closing as have created a new branch to deal with the many conflicts of 11 months of sadness.

PR #500, opened 2018-07-25
Remove old riff-raff artifacts that are not used

sihil:
The aim of this PR is to delete old and unused Riff-Raff artifacts.

The strategy is:
 - list all of the project names in the root of the S3 bucket
 - for each project name
   - find builds we should keep:
     - query the riff-raff DB for the last 50 successful deploys of that project
     - find the most recent 5 for each stage in the results
     - make a list of builds to keep by combining the list above
   - tag builds that we don't want to keep:
     - get a complete list of all builds in the bucket
     - compute `allBuilds - buildsToKeep`
     - tag all objects under each `${project}/${buildId}` prefix
 - Modify the lifecycle policy on the artifact bucket to remove files that are older than 40 days **_AND_** tagged

TODO
------
 - [x] Set up configuration so this is only ever run in PROD (everywhere uses the PROD artifacts, but their own DB) - default off, switch on in PROD
 - [x] Set this up to run on a schedule
 - [x] Ensure it won't run multiple times
 - [x] Add a lifecycle policy that actually tidies up the files
 - [x] Check with @jorgeazevedo that he's not relying on the contents of the bucket
 - [x] Deal with DB errors
 - [x] Add some tests to convince ourselves that this won't delete active artifacts
 - [x] Check if we need to de-tag anything tagged during development and testing
 - [ ] Enable this feature in the PROD configuration file
 - [ ] Apply the lifecycle policy from https://github.com/guardian/deploy-tools-platform/pull/121

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -4,7 +4,7 @@ import Helpers._
 val commonSettings = Seq(
   organization := "com.gu",
   scalaVersion := "2.12.2",
-  scalacOptions ++= Seq("-deprecation", "-feature","-language:postfixOps,reflectiveCalls,implicitConversions"
+  scalacOptions ++= Seq("-deprecation", "-feature", "-language:postfixOps,reflectiveCalls,implicitConversions", "-Ypartial-unification"
 //    , "-Xfatal-warnings" TODO: Akka Agents have been deprecated. Once they have been replaced we can re-enable, but that's not trivial
   ),
   scalacOptions in(Compile, doc) ++= Seq(
--- a/riff-raff/app/housekeeping/ArtifactHousekeeping.scala
+++ b/riff-raff/app/housekeeping/ArtifactHousekeeping.scala
@@ -2,10 +2,12 @@ package housekeeping
 
 import com.amazonaws.services.s3.AmazonS3
 import com.amazonaws.services.s3.model.{ListObjectsV2Request, ObjectTagging, SetObjectTaggingRequest, Tag}
-import com.gu.management.Loggable
 import conf.Configuration
+import controllers.Logging
 import deployment.{DeployFilter, Deployments, PaginationView}
-import org.joda.time.{DateTime, Duration}
+import lifecycle.Lifecycle
+import org.joda.time.{DateTime, Duration, LocalTime}
+import utils.{DailyScheduledAgentUpdate, ScheduledAgent}
 
 import scala.annotation.tailrec
 import scala.collection.JavaConverters._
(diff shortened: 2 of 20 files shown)


jorgeazevedo:
@sihil I'm not relying on the contents of the bucket, you are free to delete. Thanks for checking in!

riff-raff/conf/routes
```diff
@@ -119,6 +119,7 @@ GET         /testing/view/:uuid                             controllers.Testing.
 GET         /testing/addStringUUID                          controllers.Testing.transferAllUUIDs
 GET         /testing/testcharset                            controllers.Testing.testcharset
 GET         /testing/deployinfo                             controllers.Testing.hosts
+GET         /testing/doHousekeeping                         controllers.Testing.doHousekeeping
```
katebee: This probably wants to be removed before release 😄

sihil: Yes. Good point. Will remove it once I've finished testing.

riff-raff/app/housekeeping/ArtifactHousekeeping.scala
```diff
@@ -0,0 +1,144 @@
...
+  }
+  def shutdown() {
+    scheduledAgent.foreach(_.shutdown())
+    scheduledAgent = None
+  }
+
+  def getBuildIdsToKeep(projectName: String): Either[Throwable, List[String]] = {
+    for {
+      deployList <- deployments.getDeploys(
+        filter = Some(DeployFilter(projectName = Some(s"^$projectName$$"), status = Some(RunState.Completed))),
```
katebee: 💯 💎 Good idea!

katebee:
asking because I am curious regarding this item of the checklist:

> - [x] Check if we need to de-tag anything tagged during development and testing

How did you go about removing all the housekeeping tags added when we ran this before?

katebee:
Hmmmm.... 🤔 

> - [x] Add some tests to convince ourselves that this won't delete active artifacts

Okay, I have kinda covered this one with some unit tests on the different methods, and by moving some logic out into a small method called `getObjectsToTag` that can then be tested.

The invoking of the housekeeping and the actual chaining of the methods together to check which builds get passed to tagBuilds is harder to test without mocking the Configuration.

build.sbt
```diff
@@ -4,7 +4,7 @@ import Helpers._
 val commonSettings = Seq(
   organization := "com.gu",
   scalaVersion := "2.12.2",
-  scalacOptions ++= Seq("-deprecation", "-feature","-language:postfixOps,reflectiveCalls,implicitConversions"
+  scalacOptions ++= Seq("-deprecation", "-feature", "-language:postfixOps,reflectiveCalls,implicitConversions", "-Ypartial-unification"
```
mario-galic, line 7: Upgrading to 2.13 I think we can remove `-Ypartial-unification` :)

PR #501, opened 2018-08-07
Skip tagging if the list of builds to keep is an empty list

katebee:
Optional since maybe we do want to tag everything for deletion if a project has never been deployed?  

However, just in case the check on the builds to keep list returns an empty list rather than an error, maybe we want to skip tagging of it for now?

This did manage to happen on PROD last night:

`h.ArtifactHousekeeping - Keeping 0 builds of <projectName> (List())`

Discuss! 

..Either way, this probably needs to get merged and deployed to PROD before tonight 😈 🔥 🔥


jacobwinch:
I think it makes sense to do this for the initial run, so 👍from me

sihil:
I did change the db code to bubble errors up, but it doesn't hurt to be paranoid...

PR #504, opened 2018-09-19
Cloudformation change sets

mbarton:
Support deploying cloudformation templates that contain [transforms](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/transform-section-structure.html) by creating and executing a change set. This means we can use Riff Raff to deploy applications written using the [AWS Serverless Model](https://docs.aws.amazon.com/lambda/latest/dg/serverless_app.html).


magenta-lib/src/main/scala/magenta/tasks/ChangeSetTasks.scala
```diff
@@ -0,0 +1,150 @@
...
+  def description = s"Create change set $changeSetName for stack $cloudFormationStackLookupStrategy with ${templatePath.fileName}"
+  def verbose = description
+}
+
+case class CheckChangeSetCreatedTask(
+  region: Region,
+  cloudFormationStackLookupStrategy: CloudFormationStackLookupStrategy,
+  changeSetName: String,
+  duration: Long
+)(implicit val keyRing: KeyRing, artifactClient: AmazonS3) extends Task with RepeatedPollingCheck {
```
desbo: Not for this PR, but it might be worth looking into the AWS SDK "Waiter" functionality to replace `RepeatedPollingCheck`.

https://aws.amazon.com/blogs/developer/waiters-in-the-aws-sdk-for-java/

https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/cloudformation/waiters/AmazonCloudFormationWaiters.html

magenta-lib/src/main/scala/magenta/tasks/AWS.scala
```diff
@@ -384,6 +384,39 @@ object CloudFormation extends AWS {
...
+    }
+
+    val request = new CreateChangeSetRequest()
+      .withChangeSetName(name)
+      .withChangeSetType(tpe)
+      .withStackName(stackName)
+      .withCapabilities(CAPABILITY_NAMED_IAM)
+      .withParameters(paramsList.toSeq.asJava)
+
+    maybeTags.foreach { tags =>
```
adamnfish: Would this be better written using the `withTags` function alongside the other `withX` calls above, rather than mutating it after creation like this?

magenta-lib/src/main/scala/magenta/tasks/ChangeSetTasks.scala
```diff
@@ -0,0 +1,150 @@
...
+      val cfnClient = CloudFormation.makeCfnClient(keyRing, region)
+      val stackName = UpdateCloudFormationTask.nameToCallNewStack(cloudFormationStackLookupStrategy)
+
+      val request = new DescribeChangeSetRequest().withChangeSetName(changeSetName).withStackName(stackName)
+      val response = cfnClient.describeChangeSet(request)
+
+      response.getStatus match {
+        case "CREATE_COMPLETE" =>
+          true
+
```
adamnfish: Surprising amount of vertical space used for this simple function, are these blank lines needed between the case statements? Very minor point and no worries if it's to be consistent with the rest of the codebase.

magenta-lib/src/main/scala/magenta/tasks/ChangeSetTasks.scala
```diff
@@ -0,0 +1,150 @@
...
+import com.amazonaws.services.s3.AmazonS3
+import magenta.artifact.S3Path
+import magenta.deployment_type.CloudFormationDeploymentTypeParameters.{CfnParam, TagCriteria}
+import magenta.tasks.CloudFormation.ParameterValue
+import magenta.tasks.UpdateCloudFormationTask._
+import magenta.{DeployReporter, KeyRing, Region, Stack, Stage}
+
+import scala.collection.JavaConverters._
+
+case class CreateChangeSetTask(
```
adamnfish: This is a controversial use of `case class`, these are behaviour not just data structures so I'd expect them to be vanilla classes.
I assume this is for consistency with the rest of the codebase in which case no worries?

sihil:
I am generally in favour of this PR, but agree with @adamnfish's comments that there should be more tests and I'd also favour replacing the existing CFN deploy type.

I'd love to see working preview functionality for this, but that is perhaps a discussion for another time.

mbarton:
Thanks for the comments everyone. @desbo and I will look to add tests and replace the existing deployment type

adamnfish:
Tests would be great, I don't think the "replace existing" has to be a blocker for this PR as long as you tackle that immediately. Nice to have small changes, but in this case if both live for too long people will start making a mess by depending on different onces.

AWare:
Is this still being actively considered?

mbarton:
@AWare yes, Sam and I are going to spend Thursday addressing the review feedback

magenta-lib/src/main/scala/magenta/tasks/ChangeSetTasks.scala
mbarton: It was actually for debugging, we've written a test for it so will reformat

magenta-lib/src/main/scala/magenta/tasks/ChangeSetTasks.scala
mbarton: Agreed. We'll make them classes (no need to copy existing odd practice)

mbarton:
Thanks @adamnfish for the detailed feedback. Sorry it has taken a while to get back to this. Sam and I have hopefully addressed this as much as we can, maintaining the existing tests and adding some more. Are there any other places we have missed?

We've replaced the existing deploy type and have tested deploying a few existing projects ([flexible-feeds](https://github.com/guardian/flexible-feeds/blob/747cdd38217f0d7ed2c7ed7ac176bafc11f595c9/riff-raff.yaml), [typerighter](https://github.com/guardian/typerighter/blob/d01f7e9b4a77be512e195b8c491e1eaa54538316/riff-raff.yaml). Any suggestions for other projects to test would be very welcome.

@desbo @sihil @AWare we're pretty much ready if anyone fancies a review?

magenta-lib/src/main/scala/magenta/tasks/ChangeSetTasks.scala
```diff
@@ -0,0 +1,119 @@
...
+
+    val (stackName, _) = stackLookup.lookup(reporter, cfnClient)
+    val changeSetName = stackLookup.changeSetName
+
+    val describeRequest = new DescribeChangeSetRequest().withChangeSetName(changeSetName).withStackName(stackName)
+    val describeResponse = cfnClient.describeChangeSet(describeRequest)
+
+    if(describeResponse.getChanges.isEmpty) {
+      reporter.info(s"No changes to perform for $changeSetName on stack $stackName")
+    } else {
```
sihil: I would be interested in seeing the changes about to be made here via a `verbose` report? I imagine it might be quite verbose but I think this would be super useful feedback, especially if it goes wrong for some reason.

magenta-lib/src/main/scala/magenta/tasks/UpdateCloudFormationTask.scala
```diff
@@ -53,6 +52,95 @@ trait RetryCloudFormationUpdate {
   }
 }
 
+class CloudFormationStackLookup(val strategy: CloudFormationStackLookupStrategy, val changeSetName: String, createStackIfAbsent: Boolean) {
```
sihil: This no longer seems like a good name for this part of the data model as it isn't just about lookup. It contains the lookup strategy, the name of the change set to create and whether a stack should be created if it doesn't already exist. Perhaps the ever ambiguous `CloudFormationStackMetadata`??

magenta-lib/src/main/scala/magenta/tasks/ChangeSetTasks.scala
```diff
@@ -0,0 +1,119 @@
...
+    )
+
+    val (stackName, changeSetType) = stackLookup.lookup(reporter, cfnClient)
+
+    val template = processTemplate(stackName, templateString, s3Client, stsClient, region, reporter)
+    val parameters = unresolvedParameters.resolve(template, accountNumber, changeSetType, reporter, cfnClient)
+
+    reporter.info("Creating Cloudformation change set")
+    reporter.info(s"Stack name: $stackName")
+    reporter.info(s"Change set name: ${unresolvedParameters.changeSetName}")
```
sihil: Does changeSetName really need to be in this `CloudFormationParameters` object. It's already available in the `CloudFormationStackLookup` object in this scope. I don't believe it is used outside of here.

magenta-lib/src/main/scala/magenta/deployment_type/CloudFormation.scala
```diff
@@ -44,44 +49,72 @@ object CloudFormation extends DeploymentType with CloudFormationDeploymentTypePa
...
+
+    val userParams = globalParams ++ stageParams
+    val amiLookupFn = getLatestAmi(pkg, target, reporter, resources.lookup)
+
+    val stackTags = cloudFormationStackLookupStrategy match {
+      case LookupByTags(tags) => Some(tags)
+      case _ => None
     }
+
+    val changeSetName = s"riff-raff-${UUID.randomUUID().toString}"
```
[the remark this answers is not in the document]
desbo: > I imagine the concern is that it could fail if it exists?
Yep, all changesets for a CFN stack must have unique names. Is there a unique identifier for each deploy we can access from the params?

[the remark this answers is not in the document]
sihil: Unique even after it is deleted?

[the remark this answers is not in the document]
desbo: Ah, I forgot we do that. I tested it in the console and you can indeed re-use names after deletion.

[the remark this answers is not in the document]
sihil: It would be better I think if it were something like `riff-raff-$escapedProjectName-$stage-$buildId-$deploymentName-$region`. Perhaps you don't need region as it will already be unique to that.

magenta-lib/src/main/scala/magenta/deployment_type/CloudFormation.scala
```diff
@@ -44,44 +50,74 @@ object CloudFormation extends DeploymentType with CloudFormationDeploymentTypePa
...
+
+    val userParams = globalParams ++ stageParams
+    val amiLookupFn = getLatestAmi(pkg, target, reporter, resources.lookup)
+
+    val stackTags = cloudFormationStackLookupStrategy match {
+      case LookupByTags(tags) => Some(tags)
+      case _ => None
     }
+
+    val rawChangeSetName = s"riff-raff-${target.parameters.build.projectName}-${target.parameters.stage}-${target.parameters.build.id}-${target.region.name}"
```
sihil: Thinking more I think you may need to have:
 - `pkg.name` to distinguish between two different CFN blocks in the same riff-raff.yaml
 - `target.stack` to distinguish between stacks when deploying to multiple stacks (that could be in the same account)

I don't think you need region here as I presume it only needs to be unique in an account/region.

sihil: As discussed IRL this smacks of being somewhat brittle (if another dimension is added to riff-raff for example). The Cloudformation deployment type already uses a millis wall clock to create a unique key for the template when the template is over the limit. Given that precedent we may as well use that approach here too and just suffer that it is impure.

Reminder that you can mess with joda time if that seems like a sane way of testing this: https://markhneedham.com/blog/2008/09/24/testing-with-joda-time/

PR #506, opened 2018-10-17
Fishing for DB error messages that get swallowed

regiskuckaertz:
Better to look at each commit separately. Each tackles a specific case where errors are being logged but the user is unaware that something wrong happened.

I'm adding new ones as I see them, perhaps @sihil you have some specific examples in mind?

Another area that could be impacted is the API, but it would affect all client applications, so probably not a good call for this PR.


sihil:
I don't have any particular example - they should eventually all be fixed. The one that has a significant impact is when a deploy update or log event fails to be written to the DB. Any of:
```
def writeDeploy(deploy: DeployRecordDocument)
def writeLog(log: LogDocument)
def updateStatus(uuid: UUID, status: RunState.Value)
def updateDeploySummary(uuid: UUID, totalTasks:Option[Int], completedTasks:Int, lastActivityTime:DateTime, hasWarnings:Boolean)
```

In these cases I believe that the DB update should be retried incase it was a transient error - possibly a new `logExceptionAndRetry` method would help here. Ideally we should have a list of exception types that are transient - but perhaps we wait until we have migrated and it will cease to be a problem anyway...

regiskuckaertz:
@sihil this is ready for a thorough review, let me know what you think 🙏

riff-raff/app/persistence/mongodb.scala
```diff
@@ -184,9 +196,11 @@ class MongoDatastore(database: MongoDB) extends DataStore with DocumentStore wit
       deployCollection.findOneByID(uuid).flatMap(DeployRecordDocument.fromDBO(_))
     }
 
-  override def writeLog(log: LogDocument) {
-    logAndSquashExceptions(Some("Writing new log document with id %s for deploy %s" format (log.id, log.deploy)),()) {
-      deployLogCollection.insert(log.toDBO, WriteConcern.Safe)
+  override def writeLog(log: LogDocument) = {
+    retry(MAX_RETRIES) { 
+      logExceptions(Some("Writing new log document with id %s for deploy %s" format (log.id, log.deploy))) {
```
sihil: I suspect that the way this composition currently does logging might not give us as much visibility as we might want. The logging will not clearly inform us if a retry eventually succeeded or not - or how many retries were attempted. I'd consider baking that into the `retry` function.

riff-raff/app/controllers/authentication.scala
```diff
@@ -87,14 +87,14 @@ class Login(deployments: Deployments, val controllerComponents: ControllerCompon
...
-    val records = deployments.getDeploys(Some(DeployFilter(deployer=Some(request.user.fullName)))).logAndSquashException(Nil).reverse
+    val records = deployments.getDeploys(Some(DeployFilter(deployer=Some(request.user.fullName)))).map(_.reverse)
     Ok(views.html.auth.profile(request, records))
   }
 
   val authorisationForm = Form( "email" -> nonEmptyText )
 
   def authList = authAction { request =>
-    Ok(views.html.auth.list(request, Persistence.store.getAuthorisationList.sortBy(_.email)))
+    Ok(views.html.auth.list(request, Persistence.store.getAuthorisationList.map(_.sortBy(_.email))))
```
sihil: Rather than pushing this down into each controller, I'd consider using one of the `views.errorPage` or `views.errorContent` templates. It should also use a 5xx status code.

riff-raff/app/persistence/mongodb.scala
regiskuckaertz: Yes, good idea 👍

PR #507, opened 2018-10-17
Apply scale-in protection to existing instances during autoscaling deploy

sihil:
A failed riff-raff deploy is somewhat laborious to recover from as you need to identify and kill the newly created instances. This aims to make that simpler by applying the scale in protection attribute to existing instances at the start of a deploy.

This attribute protects an instance from being selected for termination during scale-in events. By setting this users will be able to recover from a failed Riff-Raff deploy by simply changing the desired size of an ASG back to the original number.

The ASG console looks like this during a deploy (note the last column):
screen shot 2018-10-17 at 17 22 59

More details about instance protection are here: https://docs.aws.amazon.com/autoscaling/ec2/userguide/as-instance-termination.html#instance-protection-instance


magenta-lib/src/main/scala/magenta/tasks/ASGTasks.scala
```diff
@@ -35,6 +35,23 @@ case class TagCurrentInstancesWithTerminationTag(pkg: DeploymentPackage, stage:
   lazy val description = "Tag existing instances of the auto-scaling group for termination"
 }
 
+case class ProtectCurrentInstances(pkg: DeploymentPackage, stage: Stage, stack: Stack, region: Region)(implicit val keyRing: KeyRing) extends ASGTask {
+  override def execute(asg: AutoScalingGroup, reporter: DeployReporter, stopFlag: => Boolean, asgClient: AmazonAutoScaling) {
+    if (asg.getInstances.asScala.nonEmpty) {
+      val instanceIds = asg.getInstances.asScala.toList.map(_.getInstanceId)
+      val request = new SetInstanceProtectionRequest()
+        .withAutoScalingGroupName(asg.getAutoScalingGroupName)
+        .withInstanceIds(instanceIds: _*)
```
regiskuckaertz: Ha, the question indeed: do you need to supply instance IDs?

sihil: I believe so. The docs say this `should be one or more instance IDs` and I am happy to believe them.

regiskuckaertz: Yeah, it is unfortunate the API forces the whole conversion ceremony on you, but it wasn't written with Scala in mind.

sihil: Any time you interact with a Java API it's the same. The splat operator is actually quite nice in this regard as it skips the need to do a asJava conversion...

PR #512, opened 2018-11-07
De-duplicate log entries when reading from the DB

sihil:
I've noticed a number of duplicated log lines which are causing chaos in the log viewer. I suspect that this is down to our policy of trying Mongo updates repeatedly if we get an error, meaning that we ensure that a log line is successfully written at the cost of introducing duplicates in the case that it has written successfully but still threw an error.

The proper fix is to switch from inserts to upserts to ensure that any deduplication happens in the DB. However this can't be done until duplicates have already been removed so this is a short term kludge to fix the immediate problem by filtering out duplicates on read.

### DIFF
--- a/riff-raff/app/persistence/package.scala
+++ b/riff-raff/app/persistence/package.scala
@@ -37,14 +37,14 @@ object `package` {
 
   implicit class richList[T](list: List[T]) {
     def distinctOn[N](f: T => N): List[T] = {
-      list.foldLeft((Set.empty[N], List.empty[T])){ case ((distinctElements, acc), item) =>
+      list.foldRight((Set.empty[N], List.empty[T])){ case (item, (distinctElements, acc)) =>
           val element = f(item)
           if (distinctElements.contains(element)) {
             distinctElements -> acc
           } else {
             (distinctElements + element) -> (item :: acc)
           }
-      }._2.reverse
+      }._2
     }
   }
 }
--- a/riff-raff/app/persistence/package.scala
+++ b/riff-raff/app/persistence/package.scala
@@ -1,12 +1,12 @@
 package persistence
 
-import magenta._
-import deployment.{PaginationView, DeployFilter}
 import com.mongodb.casbah.commons.MongoDBObject
 import com.mongodb.casbah.Imports._
 import com.mongodb.DBObject
 import com.mongodb.casbah.MongoCursor
-import org.joda.time.{LocalDate, DateMidnight}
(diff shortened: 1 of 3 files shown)


sihil:
Deployed to CODE and shown to work as expected.

riff-raff/app/persistence/package.scala
```diff
@@ -1,12 +1,15 @@
...
-import deployment.{PaginationView, DeployFilter}
+import deployment.{DeployFilter, PaginationView}
 import com.mongodb.casbah.commons.MongoDBObject
 import com.mongodb.casbah.Imports._
 import com.mongodb.DBObject
 import com.mongodb.casbah.MongoCursor
-import org.joda.time.{LocalDate, DateMidnight}
+import org.joda.time.{DateMidnight, LocalDate}
+
+import scala.collection.mutable
```
regiskuckaertz: I don't think this is needed as `+` creates a new set anyway.

riff-raff/app/persistence/package.scala
```diff
@@ -34,4 +37,17 @@ object `package` {
...
+  implicit class richList[T](list: List[T]) {
+    def distinctOn[N](f: T => N): List[T] = {
+      list.foldLeft((mutable.Set.empty[N], List.empty[T])){ case ((distinctElements, acc), item) =>
+          val element = f(item)
+          if (distinctElements.contains(element)) {
+            distinctElements -> acc
+          } else {
+            (distinctElements + element) -> (item :: acc)
+          }
+      }._2.reverse
```
regiskuckaertz: You can avoid that extra traversal with a lazy fold (`foldRight`).

riff-raff/app/persistence/package.scala
sihil: Ah yes! I originally added that when I was planning to mutate a Set and didn't remove it. Good spot.

riff-raff/app/persistence/package.scala
sihil: I don't think I can as foldRight will remove the duplicates in a different order (the last copy in the list will be retained rather than the first copy).

regiskuckaertz: Ah yes indeed, I thought that would not be a problem.

sihil: It might not be to be fair, I haven't really thought about it. I'll try it and see if it is still working as expected.

sihil: Yep - that works happily.

PR #514, opened 2018-12-05
Migration from Mongo to PostgresQL

regiskuckaertz:
Documentation to follow.

- [x] Move code into the riff raff project
- [x] Hook up script with migration view
- ~~Deal with failures: (1) failures to decode/read from mongo, (2) failures to write to postgre~~


migration/src/main/scala/migration/App.scala
```diff
@@ -0,0 +1,48 @@
...
+
+object MigrationApp extends App with Args with Migrator {
+
+  val WINDOW_SIZE: Int = 1000
+
+  override def run(args: List[String]): IO[Nothing, ExitStatus] = 
+    parseArgs(args) redeem (
+      _ => usage.const(ExitStatus.ExitNow(1)), 
+      cfg => IO.bracket(
+        Mongo.connect(cfg.mongoUri) <* Postgre.connect(cfg.pgUri, cfg.pgUser, cfg.pgPassword)
```
alexduf: Nitpick: I think the shorthand for PostgreSQL is Postgres

migration/src/main/scala/migration/Args.scala
```diff
@@ -0,0 +1,45 @@
...
+    indent, putStrLn("options:"),
+    indent, indent, putStrLn("-t <table>,...,<table>  List of tables to migrate"),
+    indent, indent, putStrLn("-n <int>                Max number of rows to migrate"),
+    indent, indent, putStrLn("-d                      Dry run, does not actually write anything"),
+    lineFeed
+  ))(identity _).void.catchAll {
+    case t => IO.unit // this won't happen
+  }
+
+  def catastrophe(e: MigrationError): IO[Nothing, Unit] = IO.unit
```
alexduf: Don't you want to trace something here?

migration/src/main/scala/migration/Migrator.scala
```diff
@@ -0,0 +1,47 @@
...
+
+      iterations = Math.ceil(n / WINDOW_SIZE).toInt
+
+      _ <- (0 until iterations).toList.traverse { i =>
+        val skip = i * WINDOW_SIZE
+        val limit = WINDOW_SIZE.toLong.min(n - skip.toLong).toInt
+        for {
+          _ <- putStr(s"Moving items ${skip} to ${skip + limit}...").leftMap[MigrationError](ConsoleError(_)).lift[Migration]
+          // TODO log errors
+          recs <- getCursor[A](coll, skip, limit)
```
alexduf: If I understand correctly the mongo collection is "sliced" into portions of size "WINDOW_SIZE".
The conditions for this to work are:
 - Mongo guarantees the order of its collection
 - Absolutely no write operation can be done on the collection during the migration as this might affect the order, and lead to data loss

I don't know if the first one is true, and the second can be enforced by turning the UI off during the migration, but wouldn't there be a way to use the same cursor for the whole table? That may or may not be desirable so I'm more asking the question rather than suggesting a solution.

migration/src/main/scala/migration/Args.scala
regiskuckaertz: Oh yeah, will add a stack trace 👍

migration/src/main/scala/migration/Migrator.scala
regiskuckaertz: 🤦‍♂️  great observation! I'll update and move the slicing _after_ the point of backpressure.

regiskuckaertz:
> I've reviewed under the assumption the migration code would be thrown away at some point, is that correct?

That is my understanding, yes.

migration/src/main/scala/migration/Args.scala
regiskuckaertz: In fact all this will go away, since now the UI will be the app itself, not the command line

regiskuckaertz:
@alexduf I did not see the dependency on cats, which is all I needed so I've removed the extra scalaz dep. Also got rid of the free algebra which was unnecessary. The last point is to figure out how to run the IO loop in a separate thread so that users can see the progress in the UI while the migration is running in the background.

project/Dependencies.scala
```diff
@@ -75,4 +75,15 @@ object Dependencies {
     m.excludeAll(ExclusionRule(organization = "com.google.code.findbugs", name = "jsr305"))
   )
 
+  lazy val migrationDeps = Seq(
+    "org.mongodb.scala" %% "mongo-scala-driver" % "2.5.0",
+    "org.scalikejdbc"   %% "scalikejdbc"        % "3.3.0",
+    "org.scalaz"        %% "scalaz-zio"         % "0.5.1",
```
regiskuckaertz: NB: despite the name, this is a zero-dependency library that does not inject the whole scalaz ecosystem (as opposed to cats-effect).

marialivia16:
[RiffRaff DB migration](https://trello.com/c/zFGjwapM/132-riffraff-db-migration)

PR #515, opened 2018-12-10
Configuration improvements

marialivia16:
Riff-raff is currently using this https://github.com/guardian/guardian-configuration library for dealing with configuration. This uses an 8 step strategy to decide what config to use - it is hard to follow and it feels unnecessary for the amount of config riff-raff deals with. Also, I've briefly tried to have a test config file, but failed to make the tests use that file because of this library.

I'm proposing a simplified version of doing configuration. This PR is reducing the number of config files to 1 (`.gu/riff-raff.properties`). I managed to successfully run this locally and on CODE.

I'm opened to any other way of doing config, your suggestions will help a lot. I've created this PR mainly so we can have a discussion about how config works.


jacobwinch:
I think simplifying this is a good idea 👍 

If we can replace the slightly ambiguous line [here](https://github.com/guardian/riff-raff/blob/master/readme.md#how-do-i-run-riff-raff-locally-if-i-want-to-hack-on-it), with a simple instruction to download all the required config from S3 then I think that would be big improvement in terms of getting more developers involved with this project!

Perhaps @sihil can offer a view as to why the initial approach was chosen (and whether it is still helping us in a way that I've overlooked).

sihil:
The initial approach was taken because it was the standard way that configuration was done across all applications. It can seem complicated - which is why it used to have a management page that shows all of the current config and where it has come from (this was on port 18080 but doesn't seem to be there anymore). It's also logged out at startup.

I'm up for simplifying but if we do that then I'd suggest adopting Play's built in config stuff and re-writing into HOCON. It's also an opportunity to move config from being a global object into a class that plays better with compile time DI. Finally we might want to consider using [ficus](https://github.com/iheartradio/ficus).

Worth having a chat about this - especially in light of the fact that there is a small team looking at solving config for the department. Perhaps we can be a guinea pig for their solution?

riff-raff/app/conf/context.scala
```diff
@@ -1,40 +1,48 @@
...
+import scala.concurrent.duration._
 import scala.util.{Success, Try}
 
-class Configuration(val application: String, val webappConfDirectory: String = "env") extends Logging {
-  protected val configuration = ConfigurationFactory.getConfiguration(application, webappConfDirectory)
+object Config extends Logging {
+
+  val applicationConf = ConfigFactory.parseResources("application.conf")
+  val userConf = ConfigFactory.parseFile(new File(s"${scala.util.Properties.userHome}/.gu/riff-raff.conf"))
+  val configuration = userConf.withFallback(applicationConf).resolve()
```
sihil: maybe make these private to stop users taking shortcuts

riff-raff/app/conf/context.scala
```diff
@@ -1,40 +1,48 @@
...
+
+  private def getString(path: String): String = configuration.getString(path)
+  private def getStringOpt(path: String): Option[String] = Try(configuration.getString(path)).toOption
+  private def getStringList(path: String): List[String] = configuration.getString(path).split(",").map(_.trim).toList
+  private def getBooleanOpt(path: String): Option[Boolean] = Try(configuration.getBoolean(path)).toOption
+  private def getIntOpt(path: String): Option[Int] = Try(configuration.getInt(path)).toOption
```
sihil: Ditto, these could also be private

PR #527, opened 2019-02-27
Few changes to migration

regiskuckaertz:
The most important change is that the mongo queries were running on a yielding thread pool, which is bounded by the number of CPUs. They now run on an unbounded pool suited for synchronous, blocking IO.


regiskuckaertz:
I've tested this on CODE->CODE, let's do PROD->CODE asap.

PR #531, opened 2019-03-05
Instantiate the metrics scheduler only once

marialivia16:
Following the config migration we were instantiating one metrics object per postgres request 😱

### DIFF
--- a/riff-raff/app/persistence/datastore.scala
+++ b/riff-raff/app/persistence/datastore.scala
@@ -2,7 +2,7 @@ package persistence
 
 import java.util.UUID
 
-import conf.{Config, DatastoreMetrics}
+import conf.{Config, DatastoreMetrics, DatastoreRequest}
 import controllers.{ApiKey, AuthorisationRecord, Logging}
 import org.joda.time.DateTime
 import play.api.Logger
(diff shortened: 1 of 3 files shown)


mchv:
@marialivia16 🥇

riff-raff/app/persistence/datastore.scala
```diff
@@ -48,7 +48,7 @@ abstract class DataStore(config: Config) extends DocumentStore with Retriable {
         Left(t)
     }
 
-  def run[T](block: => T): T = new DatastoreMetrics(config, this).DatastoreRequest.measure(block)
```
regiskuckaertz: nice catch 👍

regiskuckaertz:
It is weird this behaviour was only noticed now, because the shape of that part of the program hasn't changed during migration.

In hindsight, the `DatastoreMetrics` class is very dangerous and should be marked private, with only a singleton instance in the companion object.

regiskuckaertz:
(allocating an object per request is nothing compared to creating _a scheduled job_ per request, which is what `new DatastoreMetrics` is doing 😨)

marialivia16:
@regiskuckaertz Good point, we should definitely address it.

PR #533, opened 2019-03-05
Use sql function to convert startTime to TIMESTAMP when running queries

philmcmahon:
https://github.com/guardian/riff-raff/pull/532/files adds a new index - this changes modifies the app to make use of the same timestamp conversion function as is used in the index (otherwise the index will not be used)>


nicl:
This is 2.5 years old so going to close for now. If it was required I suspect someone would have shouted by now!

PR #538, opened 2019-04-12
Upgrade to AWS SDK v2

marialivia16:
I got too much into upgrading libraries and this happened...

Most of the changes are syntax related, yay!

It is a bit annoying that some libs use v1 so we have to pass an old client around in a couple of cases, but that will probably go away with time.

I've tested deployment on CODE and seems to be happy.


marialivia16:
[RIffRaff GDPR](https://trello.com/c/1xymD0TE/176-riffraff-gdpr)

philmcmahon:
wow

philmcmahon:
Nice! This feels too long to review properly, but if it's working on CODE then 👍 👍

PR #547, opened 2019-06-06
Target lambda deployments using tags

sihil:
A phoenix of #499.

Making a deployment generic enough to run across several stages or stacks is currently difficult and requires a high level of discipline with the naming of functions. For example it is currently necessary to hardcode the function name in the cloudformation rather than use dynamically generated names.

This PR allows lambdas to be targeted based on the set of tags that they have. This is similar to the way that autoscaling groups are discovered in `autoscaling` deploys.

This allows a more general approach to deployment. https://github.com/guardian/cloudwatch-logs-management is a good example of where this is helpful: a stack that can be deployed to multiple AWS accounts in which the function names are not specified.

The code follows the same pattern used for `cloud-formation` where a stack can be discovered by either name or tags. A trait and case classes are used to express the target function as `LambdaFunctionName` or `LambdaFunctionTags`. The latter is then resolved via the AWS APIs if necessary at execution time.

Note that the following permissions are required to be added to the account's riff-raff user prior to this working (although some have `lambda:*` which already covers it):
```
 - lambda:ListFunctions
 - lambda:ListTags
```

Finally - the code currently fetches all lambdas and then all tags for all lambdas in a given account (we must fetch all to ensure there isn't a second lambda with the same tags to prevent non-deterministic deployments). This means we make a lot of `ListTags` requests (one for each function) that will gradually get worse as more functions are rolled out. In the future a short term cache might be considered to ease the burden on the AWS API.

### DIFF
--- a/magenta-lib/src/main/scala/magenta/deployment_type/Lambda.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/Lambda.scala
@@ -44,6 +44,10 @@ object Lambda extends DeploymentType  {
   val prefixStackParam = Param[Boolean]("prefixStack",
     "If true then the values in the functionNames param will be prefixed with the name of the stack being deployed").default(true)
 
+  val prefixStackToKeyParam = Param[Boolean]("prefixStackToKey",
+    documentation = "Whether to prefix `package` to the S3 location"
+  ).default(true)
+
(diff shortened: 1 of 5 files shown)


sihil:
@guardian/deploy-infrastructure

PR #549, opened 2019-06-10
Use automatically create distribution buckets (lambda only)

sihil:
This is an implementation of an idea that @philmcmahon and I were throwing around to solve the scenario in which you want to deploy a stack to multiple accounts easily. There are two main approaches:
 - put the artifact into a shared bucket and point multiple instances/lambdas across the multiple accounts to the same shared bucket 
 - have one artifact bucket per account and point each instance/lambda at the local bucket

The second of these is preferable for a few reasons but it is not possible to create two buckets with the same name in different accounts. One alternative is to use the `bucketResource` attribute (currently only supported on `aws-s3`) but another option is to automatically create a local bucket in each account/region for uploading artifacts to.

There are a few parts to this:
 - code to create unique buckets in each account
 - a way of communicating this bucket name into a cloudformation template
 - a way of using this bucket in lambda and autoscaling deploys

It turns out that we already have some code to create per account buckets from the CloudFormation work and so this can be reused to create a workable solution. Initially we'll only support this in Lambda and see whether it is useful.

I've added a `autoDistBucketParameter` to the `cloud-formation` deployment type. Similar to `amiParameter` this specifies the CFN template parameter that should be injected with the name of the automatically created bucket. This is guaranteed to be account local and will look something like `riff-raff-artifact-bucket-123456789-eu-west-1`.

In addition I've added an `autoDistBucket` parameter to the `aws-lambda` deployment type as an alternative to the `bucket` parameter. When true it uses the same bucket name creation logic as described above to upload the artifact to and when calling `UpdateFunctionCode`.

### DIFF
--- a/magenta-lib/src/test/scala/magenta/deployment_type/CloudFormationTest.scala
+++ b/magenta-lib/src/test/scala/magenta/deployment_type/CloudFormationTest.scala
@@ -5,15 +5,21 @@ import java.util.UUID
 import magenta._
 import magenta.artifact.S3Path
 import magenta.fixtures._
-import magenta.tasks.CloudFormation.{SpecifiedValue, UseExistingValue}
+import magenta.tasks.CloudFormation.{SpecifiedValue, TemplateBody, UseExistingValue}
 import magenta.tasks.UpdateCloudFormationTask._
 import magenta.tasks._
+import org.scalatest.mockito.MockitoSugar
+import org.mockito.Mockito._
 import org.scalatest.{FlatSpec, Inside, Matchers}
 import play.api.libs.json.{JsBoolean, JsString, JsValue, Json}
-import software.amazon.awssdk.services.cloudformation.model.{Change, ChangeSetType, Parameter}
+import software.amazon.awssdk.services.cloudformation.CloudFormationClient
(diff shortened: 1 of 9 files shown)


philmcmahon:
Sorry, I reviewed this yesterday but got distracted halfway through. My only thought was maybe we can think of a better name than `autoDist` to refer to the bucket - it works with the convention composer-dist, frontend-dist etc. but might not be obvious to others. Maybe just `autoArtifact`?

magenta-lib/src/test/scala/magenta/deployment_type/CloudFormationTest.scala
```diff
@@ -189,6 +208,8 @@ class CloudFormationTest extends FlatSpec with Matchers with Inside {
     CloudFormationDeploymentTypeParameters.unencryptedTagFilter(Map("Bob" -> "bobbins", "Encrypted" -> "true")) shouldBe false
   }
 
+  it should ""
```
philmcmahon: is this deliberately empty?

philmcmahon:
Other than above comments looks great, I think it's possible to follow what's happening in the code.

sihil:
I've been pondering more about this since playing around with SSM parameter store CFN parameters and am now wondering if that might be a better solution - although still requires some special magic around the lambda deployment type. Might come and chat with you about it tomorrow.

sihil:
Closing in favour of #551

PR #545, opened 2019-05-10
Upgrade to AWS SDK v2 

marialivia16:
Fixed a couple of issues from last time (https://github.com/guardian/riff-raff/pull/538) and tested on CODE.

PR #550, opened 2019-06-13
Fix large cloudformation deploys

philmcmahon:
Clouformation deploys involving a large template are currently broken due to the recent [upgrade of the AWS sdk](https://github.com/guardian/riff-raff/pull/545) in riffraff. This fixes the problem. I couldn't find a method in the new sdk for generating a url given a bucket and key so I've manually constructed it here.

Example failed deploy: https://riffraff.code.dev-gutools.co.uk/deployment/view/84fe9dd4-3182-4d38-9226-edf97512c9f5


philmcmahon:
Thanks! I checked the frontend one and it isn't public

PR #551, opened 2019-06-17
Add option to lookup bucket from an SSM key in the target account

sihil:
NOTE: this is built on top of #547 and should be merged _after_ that

An alternative approach to #549 which avoids creating buckets automatically and is more aligned to existing behaviour.

This is an implementation of an idea that @philmcmahon and I were throwing around to solve the scenario in which you want to deploy a stack to multiple accounts easily. There are two main approaches:
 - put the artifact into a shared bucket and point multiple instances/lambdas across the multiple accounts to the same shared bucket 
 - have one artifact bucket per account and point each instance/lambda at the local bucket

The second of these is preferable for a few reasons (avoidance of cross account permissions primarily) but it is not possible to create two buckets with the same name in different accounts. That means buckets will be differently named in each account. We could create unique buckets automatically or allow a bucket to be configured. In the latter case we could configure it in riff-raff or in the target account.

SSM is a good fit to this as there is already functionality in CloudFormation templates for doing lookups. We just need to add lookup capabilities to the deployment tasks in riff-raff.

This adds `bucketSsmLookup` and `bucketSsmKey` parameters to the Lambda type. When `bucketSsmLookup` is set to true then an SSM `getParameter` API call will be made with the value of `bucketSsmKey` (default is `/account/service/artifact.bucket`). The retrieved value will be used as the bucket to upload artifacts to and to update functions from.

It would make sense to roll this functionality out to the `autoscaling` deployment type in the future.

### DIFF
--- a/magenta-lib/src/main/scala/magenta/deployment_type/BucketParameters.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/BucketParameters.scala
@@ -1,0 +1,36 @@
+package magenta.deployment_type
+
+import magenta.tasks.S3.{Bucket, BucketByName, BucketBySsmKey}
+import magenta.{DeployReporter, DeployTarget, DeploymentPackage}
+
+trait BucketParameters {
+  this: DeploymentType =>
+
+  val bucketParam = Param[String]("bucket",
+    documentation =
+      """
+        |Name of the S3 bucket where the distribution artifacts should be uploaded.
+      """.stripMargin,
+    optional = true
+  )
+
+  val bucketSsmLookupParam = Param[Boolean]("bucketSsmLookup",
+    """Lookup the bucket for uploading distribution artifacts from SSM in the target account and region. This is
+      |designed to be used in conjunction with a matching `AWS::SSM::Parameter::Value<String>` CFN parameter
+      |on any related CloudFormation template.""".stripMargin
+  ).default(false)
+
+  val bucketSsmKeyParam = Param[String]("bucketSsmKey",
+    """The SSM key used to lookup the bucket name for uploading distribution artifacts.""".stripMargin
+  ).default("/account/services/artifact.bucket")
+
+  def resolveBucket(pkg: DeploymentPackage, target: DeployTarget, reporter: DeployReporter): Bucket = {
+    val bucketSsmLookup = bucketSsmLookupParam(pkg, target, reporter)
+    val maybeExplicitBucket = bucketParam.get(pkg)
+    (bucketSsmLookup, maybeExplicitBucket) match {
+      case (true, None) => BucketBySsmKey(bucketSsmKeyParam(pkg, target, reporter))
+      case (false, Some(explicitBucket)) => BucketByName(explicitBucket)
+      case _ => reporter.fail("One and only one of the following must be set: the bucket parameter or bucketSsmLookup=true")
+    }
+  }
+}
--- a/magenta-lib/src/main/scala/magenta/deployment_type/BucketParameters.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/BucketParameters.scala
@@ -24,12 +24,12 @@ trait BucketParameters {
     """The SSM key used to lookup the bucket name for uploading distribution artifacts.""".stripMargin
   ).default("/account/services/artifact.bucket")
 
-  def resolveBucket(pkg: DeploymentPackage, target: DeployTarget, reporter: DeployReporter): Bucket = {
+  def getTargetBucketFromConfig(pkg: DeploymentPackage, target: DeployTarget, reporter: DeployReporter): Bucket = {
     val bucketSsmLookup = bucketSsmLookupParam(pkg, target, reporter)
     val maybeExplicitBucket = bucketParam.get(pkg)
+
     val bucket = (bucketSsmLookup, maybeExplicitBucket) match {
-      case (true, None) =>
-        BucketBySsmKey(bucketSsmKeyParam(pkg, target, reporter))
+      case (true, None) => BucketBySsmKey(bucketSsmKeyParam(pkg, target, reporter))
       case (false, Some(explicitBucket)) => BucketByName(explicitBucket)
       case _ => reporter.fail("One and only one of the following must be set: the bucket parameter or bucketSsmLookup=true")
     }
--- a/magenta-lib/src/main/scala/magenta/deployment_type/BucketParameters.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/BucketParameters.scala
@@ -27,10 +27,13 @@ trait BucketParameters {
   def resolveBucket(pkg: DeploymentPackage, target: DeployTarget, reporter: DeployReporter): Bucket = {
     val bucketSsmLookup = bucketSsmLookupParam(pkg, target, reporter)
     val maybeExplicitBucket = bucketParam.get(pkg)
-    (bucketSsmLookup, maybeExplicitBucket) match {
-      case (true, None) => BucketBySsmKey(bucketSsmKeyParam(pkg, target, reporter))
+    val bucket = (bucketSsmLookup, maybeExplicitBucket) match {
+      case (true, None) =>
+        BucketBySsmKey(bucketSsmKeyParam(pkg, target, reporter))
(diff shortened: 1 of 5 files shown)


philmcmahon:
This LGTM but the diff is weird- it seems to include changes from previous merged branches

johnduffell:
this would be great as i was bumping up against a problem just last week about not being able to cloudform an assets bucket separately for code and prod!

magenta-lib/src/main/scala/magenta/deployment_type/BucketParameters.scala
```diff
@@ -0,0 +1,39 @@
...
+  def resolveBucket(pkg: DeploymentPackage, target: DeployTarget, reporter: DeployReporter): Bucket = {
+    val bucketSsmLookup = bucketSsmLookupParam(pkg, target, reporter)
+    val maybeExplicitBucket = bucketParam.get(pkg)
+    val bucket = (bucketSsmLookup, maybeExplicitBucket) match {
+      case (true, None) =>
+        BucketBySsmKey(bucketSsmKeyParam(pkg, target, reporter))
+      case (false, Some(explicitBucket)) => BucketByName(explicitBucket)
+      case _ => reporter.fail("One and only one of the following must be set: the bucket parameter or bucketSsmLookup=true")
+    }
+    reporter.verbose(s"Resolved artifact bucket as $bucket")
```
johnduffell: is verbose the stuff that appears when you tick the box on the deploy screen? if so, then this is a good place to log it :)

magenta-lib/src/main/scala/magenta/tasks/AWS.scala
```diff
@@ -81,6 +82,24 @@ object S3 extends AWS {
     }
     bucketName
   }
+
+  def resolveBucket(bucket: Bucket, ssmClient: => SsmClient, reporter: => DeployReporter): String = {
+    bucket match {
+      case BucketByName(name) => name
+      case bucketBySsm: BucketBySsmKey =>
+        val resolvedBucket = bucketBySsm.resolve(ssmClient)
```
johnduffell: just a thought to reduce the number of methods calling each other and make things more cohesive:
In general you can implement polymorphism by either pattern matching or inheritance (or typeclasses if you have a special use case) wheras here we pattern match and then call a method.
```suggestion
        val resolvedBucket = SSM.getParameter(ssmClient, bucketBySsm.ssmKey)
```
then you can remove the whole resolve method from the BucketBySsmKey thereby making it into an ADT

magenta-lib/src/main/scala/magenta/tasks/AWS.scala
```diff
@@ -81,6 +82,24 @@ object S3 extends AWS {
...
+        reporter.verbose(s"Resolved bucket from SSM key ${bucketBySsm.ssmKey} to be $resolvedBucket")
+        resolvedBucket
+    }
+  }
+
+  trait Bucket
```
johnduffell:
```suggestion
  sealed trait Bucket
```

magenta-lib/src/main/scala/magenta/tasks/AWS.scala
```diff
@@ -81,6 +82,24 @@ object S3 extends AWS {
     }
     bucketName
   }
+
+  def resolveBucket(bucket: Bucket, ssmClient: => SsmClient, reporter: => DeployReporter): String = {
```
johnduffell: we have two `resolveBucket` functions, the one above turn the deploy data into a `Bucket` wheras this one turns that `Bucket` into a String.
I'd prefer them to have different names (and resolve is a bit of a generic word) - could we call the top one something like `getMagentaBucket` and this one `getAWSBucketName` or something?

project/Dependencies.scala
```diff
@@ -34,6 +34,7 @@ object Dependencies {
     "software.amazon.awssdk" % "lambda" % Versions.aws,
     "software.amazon.awssdk" % "cloudformation" % Versions.aws,
     "software.amazon.awssdk" % "sts" % Versions.aws,
+    "software.amazon.awssdk" % "ssm" % Versions.aws,
```
johnduffell: minor point in terms of this PR, but given there are only one file that directly depends on this lib, and there are so many other deps that are needed be other things, have you considered splitting magentaLib into separate sbt projects each with more fine grained dependencies (but all still building into one artifact)

johnduffell:
this looks great and I look forward to it!  I have made a few random comments but partly for my own interest as much as they are suggestions.
👍

sihil:
This won't help you cloudform artifact buckets separately for CODE and PROD, or are you saying that it becomes unnecessary?

johnduffell:
oh good point, for that it would have to look up in SSM based on tags (stage/stack/app) rather than just a hard coded lookup.
In the end I'm getting around it by having one cfn and deploy for the bucket and then once that bucket is out in prod, we just copy the assets to the prod one in the /PROD/... or /CODE/... keys

magenta-lib/src/main/scala/magenta/deployment_type/Lambda.scala
```diff
@@ -1,11 +1,14 @@
...
-import magenta.tasks.{S3Upload, UpdateS3Lambda}
+import magenta.tasks.{S3Upload, SSM, STS, UpdateS3Lambda, S3 => S3Tasks}
 import magenta.{DeployParameters, DeployReporter, DeployTarget, DeploymentPackage, KeyRing, Region, Stack}
 import software.amazon.awssdk.services.s3.S3Client
+import software.amazon.awssdk.services.ssm.SsmClient
 
-object Lambda extends DeploymentType  {
+object Lambda extends Lambda
+
+trait Lambda extends DeploymentType with BucketParameters {
```
johnduffell: just interested why the object now is a trait? is it for testability or something else?

sihil: Mainly to allow this SSM lookup behaviour to be shared by any task that uses S3. Ideally the autoscaling and s3upload tasks should also be able to use this behaviour.

johnduffell: fair enough, so I take it the way things are designed, inheritance is the only way to reuse it.  I generally avoid inheritance as it can get out of hand, but it's ok when used judiciously as it cleans up the syntax https://www.thoughtworks.com/insights/blog/composition-vs-inheritance-how-choose

magenta-lib/src/main/scala/magenta/tasks/AWS.scala
sihil: Some of the complexity comes from some things happening when the task list is being made and some during task execution but you're quite right - this is confusing in the way it mixes two approaches incoherently. I'll look to simplify this.

magenta-lib/src/main/scala/magenta/tasks/AWS.scala
```diff
@@ -459,11 +474,26 @@ object STS extends AWS {
...
+object SSM extends AWS {
+  def withSsmClient[T](keyRing: KeyRing, region: Region)(block: SsmClient => T): T = {
+    withResource(SsmClient.builder()
+      .credentialsProvider(provider(keyRing))
+      .overrideConfiguration(clientConfiguration)
+      .region(region.awsRegion)
+      .build())(block)
+  }
+
+  def getParameter(ssmClient: SsmClient, key: String): String = {
```
johnduffell: the two `String` have the same type but actually mean something different.
when I see the same type twice for different purposes in the same function signature, I try wrapping in a value class so the purpose is still clear from the type.
This would make the withSsmClient on line 88 clearer especially.
However there is also a syntactic overhead of wrapping and unwrapping for pay for the extra clarity elsewhere.
```suggestion
  case class SsmKey(value: String) extends AnyVal
  case class SsmValue(value: String) extends AnyVal

  def getParameter(ssmClient: SsmClient, key: SsmKey): SsmValue = {
```

johnduffell:
I have made a comment but it depends whether you want to ship it or polish it more! Thanks for the updates 👍

magenta-lib/src/main/scala/magenta/tasks/AWS.scala
sihil: We do that in some places but it's not consistent across the codebase so I'm inclined to leave this as is at the moment.

PR #554, opened 2019-07-15
Provide a way to create connections with a dynamic password.

alexduf:
# What I'm trying to solve

We currently provide a fixed pool to access our database. IAM provide passwords that are valid for 15 minutes, which means any connection dropped after that mark can't and won't be replaced.
We shouldn't rely on the network being reliable, this mean if some connection gets dropped, the number of active connection to the DB will slowly decrease until it's depleted.

# How is this helping

I'm providing a connection pool, with a special `ConnectionFactory`. The connection factory is invoked every time a fresh connection is attempted. This gives us an opportunity window to go to IAM and request a new password.

# Limits

- The call to IAM is blocking, this will impact the application performance, but it should work. This is alleviated by adding a 12 minute cache on the validity of the password.
- The cache doesn't attempt to be synchronous. It's perfectly possible to have two threads accessing the password and triggering two password refresh. I deemed it acceptable.

# TODO

- [x] Test on CODE

### DIFF
--- a/riff-raff/app/AppComponents.scala
+++ b/riff-raff/app/AppComponents.scala
@@ -35,7 +35,7 @@ import utils.{ChangeFreeze, ElkLogging, HstsFilter, ScheduledAgent}
 import scala.concurrent.{ExecutionContext, Future}
 import scala.util.control.NonFatal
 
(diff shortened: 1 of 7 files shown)


alexduf:
Thanks Phil!

PR #556, opened 2019-07-22
make project field exact match on deploy screen PLUS pop-out link on same mini history table

twrichards:
On the deploy screen, the small deploy history table on the right doesn't display the project name and yet the project name filter WAS a wildcard search `like  %$pn%` which meant clicking on projects like `identity` would produce a very confusing/alarming list containing builds from multiple projects (often with very different build numbers).

NOW its an exact match search on the deploy screen...
image
... feature co-authored with @TBonnin 

PLUS @philmcmahon suggested adding some sort of link to those historical deploys (from that small history table on the deploy screen), so made sense to do it at the same time...
image
... the entire row is clickable but opens in a new tab when on the deploy screen.

### DIFF
--- a/riff-raff/app/deployment/filter.scala
+++ b/riff-raff/app/deployment/filter.scala
@@ -18,6 +18,7 @@ trait QueryStringBuilder {
 
 case class DeployFilter(
   projectName: Option[String] = None,
+  isExactMatchProjectName: Option[Boolean] = None,
   stage: Option[String] = None,
   deployer: Option[String] = None,
   status: Option[RunState] = None,
@@ -27,6 +28,7 @@ case class DeployFilter(
   lazy val queryStringParams: List[(String, String)] = {
     Nil ++
       projectName.map("projectName" -> _.toString) ++
+      isExactMatchProjectName.map("isExactMatchProjectName" -> _.toString) ++
       stage.map("stage" -> _.toString) ++
       deployer.map("deployer" -> _.toString) ++
       status.map("status" -> _.toString) ++
@@ -36,7 +38,10 @@ case class DeployFilter(
 
   lazy val sqlParams: List[SQLSyntax] = List(
     projectName.map { pn =>
-      val likeString = s"%$pn%"
+      val likeString = isExactMatchProjectName match {
+        case Some(true) => pn
+        case _ => s"%$pn%"
+      }
       sqls"content->'parameters'->>'projectName' ilike $likeString"
     },
     stage.map(s => sqls"content->'parameters'->>'stage' = $s"),
--- a/riff-raff/app/views/deploy/form.scala.html
+++ b/riff-raff/app/views/deploy/form.scala.html
@@ -8,12 +8,22 @@
 
     <div class="clearfix"><p>&nbsp;</p></div>
 
+    <style type="text/css">
+        #projectInput_field > label::after {
+            content: '(exact match)';
+            display: inline-block;
+            font-style: italic;
+            font-weight: normal;
+            margin-left: 5px;
+        }
+    </style>
+
     <div class="col-md-6">
     @b3.form(routes.DeployController.processForm, 'class -> "well") {
         @CSRF.formField
         <fieldset>
             <legend>What would you like to deploy?</legend>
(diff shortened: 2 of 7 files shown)


riff-raff/app/deployment/filter.scala
```diff
@@ -36,7 +38,10 @@ case class DeployFilter(
 
   lazy val sqlParams: List[SQLSyntax] = List(
     projectName.map { pn =>
-      val likeString = s"%$pn%"
+      val likeString = isExactMatchProjectName match {
+        case Some(true) => pn
```
TBonnin: no quotes needed?

twrichards: ahhh so its using `sqls` template string interpreter... https://github.com/guardian/riff-raff/blob/fff87f9547be7fbaadf0cb3ef45e7b757a09b003/riff-raff/app/deployment/filter.scala#L45 

Which according to their documentation...
> Don’t worry again, this code is safely protected. sqls"" always treats external input values as binding parameters.
http://scalikejdbc.org/documentation/sql-interpolation.html

riff-raff/app/views/deploy/form.scala.html
```diff
@@ -8,12 +8,22 @@
 
     <div class="clearfix"><p>&nbsp;</p></div>
 
+    <style type="text/css">
+        #projectInput_field > label::after {
+            content: '(exact match)';
```
adamnfish: Is this where someone coming later would expect to find this string? Feels like this would be simpler added to the template directly, instead of splitting the content across the HTML and the CSS.

It's a minor point though, up to you.

riff-raff/app/views/snippets/recordTable.scala.html
```diff
@@ -34,9 +34,7 @@
         <tr class="rowlink@if(record.hasWarnings){ has-warnings}">
             <td><time class="makeRelativeDate" withinhours="24" datetime="@record.time">@utils.DateFormats.Medium.print(record.time)</time></td>
             <td><span class="label label-default">@record.deployerName</span></td>
-            @if(allColumns) {
-                <td><a href="@routes.DeployController.viewUUID(record.uuid.toString)" class="rowlink">@record.buildName</a></td>
-            }
+            <td><a href="@routes.DeployController.viewUUID(record.uuid.toString)" target="@if(popOutLink){ _blank }" class="rowlink">@record.buildName</a></td>
```
adamnfish: I'm not seeing this on CODE, but the change looks ok so presumably another branch is up there now?

riff-raff/conf/routes
```diff
@@ -39,7 +39,7 @@ GET         /deployment/updates/:uuid                       controllers.DeployCo
 
 GET         /deployment/request/autoComplete/project        controllers.DeployController.autoCompleteProject(term:String)
 GET         /deployment/request/autoComplete/build          controllers.DeployController.autoCompleteBuild(project:String,term:String)
-GET         /deployment/request/history                     controllers.DeployController.deployHistory(project:String, stage:Option[String])
+GET         /deployment/request/history                     controllers.DeployController.deployHistory(project:String, stage:Option[String], isExactMatchProjectName:Option[Boolean])
```
adamnfish: Should this new flag be exposed in the UI? The "navbar" that contains the controls could include this option as a checkbox?

twrichards: as discussed in person, that would require adding to the route for the 'history' page and wiring through to the `DeployFilter` plus the HTML etc to display such a tickbox - a worthwhile feature but something for another PR

riff-raff/app/views/deploy/form.scala.html
twrichards: as discussed in person, because the input box and label is not directly but rather via a couple of layers of indirection , see https://github.com/guardian/riff-raff/blob/62817fe1a1d58bc27f40e779a24efe0e75530521/riff-raff/app/views/deploy/form.scala.html#L26 
...it was cleanest to add the 'secondary label' via pseudo `::after` element

riff-raff/app/views/snippets/recordTable.scala.html
twrichards: I looked into this more and I'd actually made a big boo-boo... it was actually putting in the full project name in the middle (don't know how I missed it), I've now wrapped the contents `Status` cell in the `<a>` instead so that there are no extra `<td>` which means the cell containing the `@record.buildName` can remain conditional (`@if(allColumns)`)

PR #560, opened 2019-08-16
Set -Xmx property when starting riffraff.

philmcmahon:
So as to help avoid out of memory errors. The default MaxHeapSize on the boxes we're using is around 512MB.


philmcmahon:
Closing this after logging into a box and noticing that riffraff is already consuming 70% of memory.

PR #565, opened 2019-09-30
Minor UI improvements

regiskuckaertz:
In the preview screen, when you have lots of tasks but only want to kick-off some at the top, it's annoying to have to scroll all the way down.

Also, the "run tasks" label isn't active, i.e. the HTML should be

```html
<input type="checkbox" id="selectedKey_1">
<label for="selectedKey_1">Run tasks</label>
```

so that when you click on the text, it (de)activates the checkbox.

**Before (preview, deploy, reset not at top)**
image

**After (preview, deploy, reset at top and bottom**
image

- [x] Test on CODE

### DIFF
--- a/riff-raff/app/views/preview/yaml/showTasks.scala.html
+++ b/riff-raff/app/views/preview/yaml/showTasks.scala.html
@@ -30,6 +30,24 @@ to make it fit in the Play idea of forms but it's far from trivial. Be warned!
           }
       }
     }
+    <div class="actions">
+        @b3.submit(
+            'name -> "action",
+            'value -> "preview",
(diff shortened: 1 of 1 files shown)


riff-raff/app/views/preview/yaml/showTasks.scala.html
```diff
@@ -41,10 +59,11 @@ <h4 class="pull-left">
                         <span>
                             @b3.checkbox(
                                 form("selectedKeys[]"),
+                                'id -> s"selectedKey_$i",
```
akash1810: YES

PR #567, opened 2019-11-07
Correct redeploy docs

philmcmahon:
https://github.com/guardian/riff-raff/pull/566 wasn't quite correct - I think this is more accurate!

### DIFF
--- a/riff-raff/public/docs/index.md
+++ b/riff-raff/public/docs/index.md
@@ -26,4 +26,4 @@ Using Riff-Raff
 
 Riff-Raff not picking up builds?
 ---------------
- - Usually, redeploying Riff-Raff fixes this; we use Riff-Raff itself to deploy itself 🎉. Be sure to check that no deploys are in progress first!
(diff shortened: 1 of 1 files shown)


riff-raff/public/docs/index.md
```diff
@@ -26,4 +26,4 @@ Using Riff-Raff
 
 Riff-Raff not picking up builds?
 ---------------
- - Usually, redeploying Riff-Raff fixes this; we use Riff-Raff itself to deploy itself 🎉. Be sure to check that no deploys are in progress first!
+ - Usually, redeploying Riff-Raff fixes this; we use Riff-Raff itself to deploy itself 🎉. The deploy will complete immediately  but riffraff will only redeploy itself when there are no other running deploys. The last deploy time is shown in the bottom right of this page.
```
akash1810: Just had a thought: What's the reason for displaying the last deploy time? And would [prout](https://github.com/guardian/prout) do a similar role? That is, should we add prout to riffraff?

PR #568, opened 2019-11-11
switch to the slate bootstrap theme

akash1810:
dark mode = best mode

This diff of PR is pretty horrible as it copy+pastes the css from https://bootswatch.com/3/slate/, however I couldn't find a better way to handle web assets in this project.

**Before**
image

**After**
image


riff-raff/app/assets/stylesheets/graph.less
```diff
@@ -40,14 +40,14 @@
   margin-left: 20px;
   padding: 2px 0;
   background-color: transparent;
-  color: black;
+  color: #ffffff;
```
akash1810: `black` didn't provide enough contrast, so move to `ffffff` aka white

**before**
image

**after**
image

PR #571, opened 2020-01-15
Add AWS account ID as Anghammarad target.

philmcmahon:
Some projects have lots of different stack names so it's easier to identify an owner using the AWS account id. This change adds 'aws account id' to the list of 'targets' that riffraff sends to [anghammarad](https://github.com/guardian/anghammarad) after a deploy failure, so that this value can be used to determine the project owner.

### DIFF
--- a/riff-raff/app/notification/DeployFailureNotifications.scala
+++ b/riff-raff/app/notification/DeployFailureNotifications.scala
@@ -11,14 +11,19 @@ import lifecycle.Lifecycle
 import magenta.Message.Fail
 import magenta.deployment_type.DeploymentType
 import magenta.input.resolver.Resolver
-import magenta.{DeployParameters, DeployReporter}
+import magenta.tasks.STS
+import magenta.{DeployParameters, DeployReporter, Lookup, Region}
 import schedule.ScheduledDeployer
+import magenta.{App => MagentaApp, Stack => MagentaStack}
+import rx.lang.scala.Subscription
 
 import scala.concurrent.ExecutionContext
+import scala.util.{Failure, Success, Try}
 
 class DeployFailureNotifications(config: Config,
                                  deploymentTypes: Seq[DeploymentType],
-                                 targetResolver: TargetResolver)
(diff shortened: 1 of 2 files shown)


riff-raff/app/notification/DeployFailureNotifications.scala
```diff
@@ -31,19 +36,36 @@ class DeployFailureNotifications(config: Config,
...
-        List(App(target.app), Stack(target.stack))
+        List(App(target.app), Stack(target.stack))++ getAwsAccountIdTarget(target, parameters).toList
       } ++ List(Stage(parameters.stage.name))
     }
 
+
     deriveAnghammaradTargets match {
       case Right(targets) =>
-        val failureMessage = s"${parameters.deployer.name} for ${parameters.build.projectName} (build ${parameters.build.id}) to stage ${parameters.stage.name} failed."
+        log.info(s"Targets to notify: ${targets.toSet}")
```
Reettaphant: Do we want to remove this log message?

riff-raff/app/notification/DeployFailureNotifications.scala
```diff
@@ -31,19 +36,36 @@ class DeployFailureNotifications(config: Config,
...
+        val failureMessage = s"${parameters.deployer.name} for ${parameters.build.projectName} (build ${parameters.build.id}) to stage ${parameters.stage.name} failed. Targets: $targets"
```
Reettaphant: do we still need to the targets variable at the end of this message?

philmcmahon: Good spot, I've removed it

riff-raff/app/notification/DeployFailureNotifications.scala
philmcmahon: I've given this a clearer label but think maybe it's worth keeping - failed deploys don't happen super often so it shouldn't be too noisy.

Reettaphant: 👍

PR #572, opened 2020-02-13
Source credentials from SSM

sihil:
This changes the way credentials are built to source them from SSM param store instead of a config file in S3. At a later point we might migrate the whole config to SSM param store but this is the most important bit.

I initially wrote this to support both the config file and SSM but that just added compexity for no reason. I also initially wrote this so that keyrings were built from live data in AWS. I switched to fetch credentials at startup to make it more robust. It could be changed to have a refresh endpoint if so desired.

I've imported credentials for CODE and PROD as of 9am on the 13th Feb and the build is currently on CODE and tested as working.

There are currently no credentials for DEV. When there are, they will at least no longer be on a developers machine, although they will be shared by anyone that works on riff-raff.

## TODO
 - [ ] Remove credentials from config file, perhaps leaving a comment for those that follow

### DIFF
--- a/magenta-lib/src/main/scala/magenta/tasks/AWS.scala
+++ b/magenta-lib/src/main/scala/magenta/tasks/AWS.scala
@@ -34,11 +34,11 @@ import scala.collection.JavaConverters._
 import scala.util.Try
 import magenta.withResource
 
-object S3 extends AWS {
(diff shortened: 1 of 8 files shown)


philmcmahon:
I think leaving a comment in the credentials file would be a good shout.

PR #574, opened 2020-03-06
GCP deployment manager type

sihil:
Add support to push out updates via Google Cloud's CFN equivalent: Deployment Manager.

This is a large PR but is isolated to new features and doesn't touch existing code (except a small DRYing refactor).

There are a few components to this:
 - a tiny test deployment at `contrib/deployment-manager-test` that has helped to prove that this works
 - new `Gcp` API code that can create a client and provides get, insert and update actions on deployments plus the get action on an "operation" that represents an in flight change
 - new GCP Deployment Manager Update task that coordinates the use of the API to update or upsert a deployment
 - new GCP Deployment Manager deployment type accessible to users
 - a small change to the logging initialisation that forwards JUL logs into SLF4J

### Outstanding questions
 - At the moment we unusually allow different template names for different stages rather than passing in a stage property with the deploy's value. We should consider fixing this and removing this feature.

### Outstanding work
 - [x] Add datalake GCP credentials to parameter store
 - [x] Add CODE and PROD GCP project name mappings to Prism data
 - [x] Create a new trivial project that can be deployed using this task
 - [x] Test running this locally to execute updates to a test deployment in the datalake code
 - [x] Review the failure modes
 - [x] Evaluate whether we need an upsert task instead of only update

### DIFF
--- a/magenta-lib/src/main/scala/magenta/deployment_type/GcpDeploymentManager.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/GcpDeploymentManager.scala
@@ -101,14 +101,14 @@ object GcpDeploymentManager extends DeploymentType {
       val upsert = upsertParam(pkg, target, reporter)
       val preview = previewParam(pkg, target, reporter)
 
-      val configPathLookup: PartialFunction[String, String] = (configPathByStageParam.get(pkg), configPathParam.get(pkg)) match {
-        case (Some(map), None) => map
-        case (None, Some(default)) => { case _ => default }
+      val configPath: String = (configPathByStageParam.get(pkg), configPathParam.get(pkg)) match {
+        case (Some(map), None) =>
+          map.getOrElse(
+            target.parameters.stage.name,
+            reporter.fail(s"No config path supplied for stage ${target.parameters.stage.name} in configPathByStage")
+          )
+        case (None, Some(path)) => path
         case (None, None) | (Some(_), Some(_)) => reporter.fail(s"One and only one of configPathByStage or configPath must be provided")
-      }
-
-      val configPath = configPathLookup.lift(target.parameters.stage.name).getOrElse {
-        reporter.fail(s"No config path supplied for stage ${target.parameters.stage.name} in configPathByStage or configPath")
       }
 
       val maybeBundle = createDeploymentBundle(pkg.s3Package, configPath, reporter)
--- a/magenta-lib/src/main/scala/magenta/deployment_type/GcpDeploymentManager.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/GcpDeploymentManager.scala
@@ -62,6 +62,15 @@ object GcpDeploymentManager extends DeploymentType {
     optional = true // one of this or configPath
   )
 
+  val upsertParam: Param[Boolean] = Param[Boolean](
+    name = "upsert",
+    documentation =
+      """
+        |When true a deployment should be created if it doesn't already exist. When false a deployment will fail
+        |if it doesn't exist but still updated if it does exist.
+        |""".stripMargin
+  ).default(false)
+
   val updateAction: Action = Action(
     name = "update",
     documentation =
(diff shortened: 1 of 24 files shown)


magenta-lib/src/main/scala/magenta/deployment_type/GcpDeploymentManager.scala
```diff
@@ -0,0 +1,158 @@
...
+      implicit val artifactClient: S3Client = resources.artifactClient
+
+      val maxWaitDuration = maxWaitParam(pkg, target, reporter).seconds
+      val deploymentName = deploymentNameParam(pkg, target, reporter)
+      val upsert = upsertParam(pkg, target, reporter)
+      val preview = previewParam(pkg, target, reporter)
+
+      val configPathLookup: PartialFunction[String, String] = (configPathByStageParam.get(pkg), configPathParam.get(pkg)) match {
+        case (Some(map), None) => map
+        case (None, Some(default)) => { case _ => default }
```
philmcmahon, line 106: cant this just be `=> default`?

magenta-lib/src/main/scala/magenta/tasks/gcp/DeploymentManagerTasks.scala
```diff
@@ -0,0 +1,84 @@
...
+        }
+      }{ deployment =>
+        reporter.verbose(s"Fetched details for deployment ${deploymentName}; using fingerprint ${deployment.getFingerprint} for update")
+        Gcp.DeploymentManagerApi.update(client, project, deploymentName, deployment.getFingerprint, bundle, preview)(reporter)
+      }
+    }
+
+    def pollOperation(client: DeploymentManager, operation: DMOperation)(reporter: DeployReporter, stopFlag: => Boolean): Unit = {
+      check(reporter, stopFlag) {
+        val maybeUpdatedOperation = Gcp.DeploymentManagerApi.operationStatus(client, operation)(reporter)
```
philmcmahon, line 62: might be worth adding a type definition here so it's clear what is being folded on in the next line?

riff-raff/app/views/preview/yaml/showTasks.scala.html
```diff
@@ -53,7 +53,7 @@ <h4 class="pull-left">
                 <div class="panel-body">
                     <ul class="magenta-list task-list">
                     @deploymentTasks.tasks.map { task =>
-                        <li><span class="preview-task"><strong>@task.name</strong> @task.description</span></li>
+                        <li><span class="preview-task"><strong>@task.name</strong> @task.verbose</span></li>
```
philmcmahon, line 56: is this a deliberate change?

sihil: Yes - not sure the verbose is used in many places but it's often more descriptive. I'll review.

magenta-lib/src/main/scala/magenta/tasks/gcp/DeploymentManagerTasks.scala
sihil: Good point.

magenta-lib/src/main/scala/magenta/deployment_type/GcpDeploymentManager.scala
sihil: It's expecting a `PartialFunction[String, String]` and `default` is a String, hence the `{ case _ => default}`. Perhaps I can express this more clearly by omitting the intermidiate lookup function.

sihil: I've cleaned it up to make it simpler. Improves the error messaging too.

magenta-lib/src/main/scala/magenta/tasks/gcp/DeploymentManagerTasks.scala
sihil: Done, although it's a wrapper type that is really an either so perhaps helps less than expected.

riff-raff/app/views/preview/yaml/showTasks.scala.html
sihil: So I killed `verbose` entirely. Well... not entirely since I didn't want to mess with persistance, but mostly.

magenta-lib/src/main/scala/magenta/deployment_type/GcpDeploymentManager.scala
philmcmahon: nice

PR #580, opened 2020-04-24
Change param logic to stop default template values overriding live stack values

sihil:
This PR changes the logic that resolves the set of parameters to use when creating a CloudFormation change set. It fixes a long standing bug where a default value for a parameter in the template took priority over a live value in the CloudFormation stack.

Note that this PR cannot be merged until #579 as it builds on restructuring that was introduced there.

This is a relatively risky change so it would be good to have a couple of pairs of eyes on it. 

 - [x] Try this on a handful of CODE stacks using Riff-Raff CODE prior to merging.

### DIFF
--- a/magenta-lib/src/test/scala/magenta/deployment_type/CloudFormationTest.scala
+++ b/magenta-lib/src/test/scala/magenta/deployment_type/CloudFormationTest.scala
@@ -6,7 +6,7 @@ import magenta._
 import magenta.artifact.S3Path
 import magenta.fixtures._
 import magenta.tasks.CloudFormation.{SpecifiedValue, UseExistingValue}
-import magenta.tasks.CloudFormationParameters.{InputParameter, TemplateParameter}
+import magenta.tasks.CloudFormationParameters.{ExistingParameter, InputParameter, TemplateParameter}
 import magenta.tasks.UpdateCloudFormationTask._
 import magenta.tasks._
 import org.scalatest.{EitherValues, FlatSpec, Inside, Matchers}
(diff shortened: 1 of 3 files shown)


magenta-lib/src/test/scala/magenta/deployment_type/CloudFormationTest.scala
```diff
@@ -119,11 +119,118 @@ class CloudFormationTest extends FlatSpec with Matchers with Inside with EitherV
...
+      List(ExistingParameter("param1", "monkey", None), ExistingParameter("param2", "monkey", None))
+    val templateParameters =
+      List(TemplateParameter("param1", default = false), TemplateParameter("param2", default = false), TemplateParameter("param3", default = false))
+    val combined = combineParameters(deployParameters, existingParameter, templateParameters, Map.empty)
+
+    combined.left.value should startWith("Missing parameters for param3:")
+  }
+
+  it should "fail when the user provides parameters that are not in the template" in {
+    val deployParameters = Map("Stack" -> "cfn", "Stage" -> "PROD", "BuildId" -> "543")
```
mbarton, line 216: The user here being the contents of `riff-raff.yaml`?

sihil: Yeah - either the `templateParameters` directly or the result of the `amiParameters`/`amiParametersToTags` params in the `riff-raff.yaml`.

magenta-lib/src/test/scala/magenta/deployment_type/CloudFormationTest.scala
```diff
@@ -119,11 +119,118 @@ class CloudFormationTest extends FlatSpec with Matchers with Inside with EitherV
...
+    val combined = combineParameters(deployParameters, existingParameter, templateParameters, Map("param3" -> "user-provided"))
+
+    combined.right.value should be(Map(
+      "param1" -> UseExistingValue,
+      "param2" -> UseExistingValue,
+      "param3" -> SpecifiedValue("user-provided")
+    ))
+  }
+
+  it should "always specify parameters when the template contains deploy parameters" in {
```
amyhughes, line 189: I think this partly shows that riff-raff would have resolved our case in a sensible way with this change - if I'm reading this right the deploy parameter "PROD" is chosen for "Stage" over the existing parameter "CODE".

I think our example was slightly different in that there was no existing parameter (we were deploying the stack for the first time), only a default parameter for "Stage" which was different to the deploy parameter (default was "CODE" but we were deploying "PROD"). A sensible behaviour in this case would have been for the deploy parameter to be chosen over the default parameter.

I'm sorry if I've missed a test which spells this out, or something else which obviously says deploy parameters should be used over anything else.

sihil: That particular behaviour hasn't changed. If you have a parameter called `Stage` then it would always have been populated with the stage name that Riff-Raff was targetting to. 

As you surmise this test is just trying to prove that the existing values are irrellevant.

amyhughes:
This looks really sensible, I'd be happy to try this on a couple of CODE stacks later this week.

magenta-lib/src/test/scala/magenta/deployment_type/CloudFormationTest.scala
amyhughes: Excellent

sihil:
Let mw know if you get a chance. Keen to get this merged sooner rather than later so I can redeploy cloudwatch-logs-management but perhaps I should progress that separately.

PR #584, opened 2020-05-06
adjust filter in TargetController

akash1810:
Use the exact projectName when filtering for recent deploys, we've already retrieved an entry from the dynamo repository (in the lines before this) so we can use the exact projectName rather than regex it as we've validated it exists.

# Before
`DeployFilter(Some(^editorial-tools:remote-machines$),None,Some(PROD),None,None,None,None)`

# After
`DeployFilter(Some(editorial-tools:remote-machines),None,Some(PROD),None,None,None,None)`

Related to https://github.com/guardian/amiable/pull/55.

### DIFF
--- a/riff-raff/app/controllers/TargetController.scala
+++ b/riff-raff/app/controllers/TargetController.scala
@@ -41,7 +41,7 @@ class TargetController(config: Config,
     val maybeTargetId = targetDynamoRepository.get(targetKey, projectName)
     maybeTargetId.map { targetId =>
       // find recent deploys of this project / stage
-      val filter = DeployFilter(projectName = Some(s"^${targetId.projectName}$$"), stage = Some(stage))
+      val filter = DeployFilter(projectName = Some(targetId.projectName), stage = Some(stage))
       val records = deployments.getDeploys(Some(filter), PaginationView(pageSize = Some(20))).logAndSquashException(Nil).reverse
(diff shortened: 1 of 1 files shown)


riff-raff/app/controllers/TargetController.scala
```diff
@@ -41,7 +41,7 @@ class TargetController(config: Config,
     val maybeTargetId = targetDynamoRepository.get(targetKey, projectName)
     maybeTargetId.map { targetId =>
       // find recent deploys of this project / stage
-      val filter = DeployFilter(projectName = Some(s"^${targetId.projectName}$$"), stage = Some(stage))
+      val filter = DeployFilter(projectName = Some(targetId.projectName), stage = Some(stage))
```
sihil, line 44: I think the aim of this was to anchor to start and end in mongo adn thus only return exact matches. Looks like `isExactMatchProjectName` (see [here](https://github.com/guardian/riff-raff/blob/master/riff-raff/app/deployment/filter.scala#L21)) now exists to handle this case and I suggest that you need to set it to true here.

akash1810: Ah interesting! Have added this now, though it was behaving as expected without it, so perhaps this is just being more explicit?

sihil: I think you'd find corner cases where it wouldn't such as trying to do this with `Editorial Tools::Workflow` when `Editorial Tools::Workflow::Workflow Frontend` exists.

PR #585, opened 2020-05-06
add --debug flag to make interactive debugging easier

akash1810:
Running `./script/start --debug` will allow you to attach a remote debugger on port 9999.

An alternative to this would be to have a `./script/debug` script instead - thoughts welcome!

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -58,8 +58,8 @@ How do I run Riff-Raff locally if I want to hack on it?
 Assuming you have a reasonably recent version of Java installed, 
 
  * Create a basic configuration file at ~/.gu/riff-raff.conf (S3 and postgres config is probably the minimum)
- * Run `./script/start` from the project root
- * visit http://localhost:9000/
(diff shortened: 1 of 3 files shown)


sihil:
Part of me feels like we could get rid of the `start` and `sbt` scripts entirely and change the instructions to be `sbt riffraff/run` or `sbt --jvm-debug 9999 riffraff/run`.

akash1810:
> Part of me feels like we could get rid of the `start` and `sbt` scripts entirely and change the instructions to be `sbt riffraff/run` or `sbt --jvm-debug 9999 riffraff/run`.

Not sure I agree; the `script/task` pattern is nice as it means we can setup the environment first, then start sbt (though I concede we're not currently doing this).

I'd be more in favour of removing `sbt` and `sbt-debug` and encapsulate everything in `script/start`.

sihil:
Fair enough. Either way this is clearly a useful addition so happy for it to be :shipit:

PR #586, opened 2020-05-11
Update multiple lambdas by tag

mbarton:
I am building a Step Functions pipeline with multiple lambdas (https://github.com/guardian/investigations-platform/pull/234). Each lambda has a different handler from a single bundle I build using [zeit/ncc](https://github.com/zeit/ncc).

At the moment even though Riff-Raff looks up the lambdas to update using tags it will fail if more than one lambda has the same tag:

```
More than one function matched for Map(Stack -> pfi, App -> pfi-video-pipeline, Stage -> PROD) (matched pfi-video-pipeline-PROD-CheckProgressFunctionAABF1-5GNB10FB5STK, pfi-video-pipeline-PROD-TriggerTranscribeFunctionD-1SET5CFQHJ43A). Failing fast since this may be non-deterministic.
```

(https://riffraff.gutools.co.uk/deployment/view/869b91c2-f0d9-49cf-abc1-d6ac9b6e2117)

For my use case it would be easiest to tag all my functions with the same Stack, App, Stage and have Riff-Raff update them all with my bundle.

The code to look up lambdas by tag is not called anywhere else in Riff-Raff apart from the `aws-lambda` task.

I can't remember why this check was added originally in https://github.com/guardian/riff-raff/pull/547):

> we must fetch all to ensure there isn't a second lambda with the same tags to prevent non-deterministic deployments

Hopefully this PR can be a discussion about it and any pros and cons of this change I have not thought of yet 😃

### DIFF
--- a/magenta-lib/src/main/scala/magenta/deployment_type/Lambda.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/Lambda.scala
@@ -31,7 +31,7 @@ trait Lambda extends DeploymentType with BucketParameters {
   )
 
   val lookupByTags = Param[Boolean]("lookupByTags",
-    """When true, this will lookup the function to deploy to by using the Stack, Stage and App tags on a function.
+    """When true, this will lookup the functions to deploy to by using the Stack, Stage and App tags.
       |The values looked up come from the `stacks` and `app` in the riff-raff.yaml and the stage deployed to.
     """.stripMargin
   ).default(false)
--- a/magenta-lib/src/main/scala/magenta/tasks/AWS.scala
+++ b/magenta-lib/src/main/scala/magenta/tasks/AWS.scala
@@ -118,7 +118,7 @@ object Lambda {
       .s3Key(s3Key)
       .build()
 
-  def findFunctionByTags(tags: Map[String, String], reporter: DeployReporter, client: LambdaClient): Option[FunctionConfiguration] = {
+  def findFunctionsByTags(tags: Map[String, String], reporter: DeployReporter, client: LambdaClient): List[FunctionConfiguration] = {
 
     def tagsMatch(function: FunctionConfiguration): Boolean = {
       val tagResponse = client.listTags(ListTagsRequest.builder().resource(function.functionArn).build())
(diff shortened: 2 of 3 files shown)


philmcmahon:
I don't know about the non deterministic stuff but the scala looks good to me 👍

sihil:
So in my design I was expecting the scenario where there were two lambda's with the same set of tags to have been a deployment error. i.e. someone has deployed two stacks of the same thing.

I see the advantage of deploying multiple things that form part of a step function - otherwise you get this silliness: https://github.com/guardian/editions/blob/master/riff-raff.yaml#L31-L39

However I'm not convinced that any two lambdas that undertake different tasks should have exactly the same set of tags. In the example above the app tag is different for each function and the resulting logging in ELK is rather pleasant. They are prefixed with the same tag but are otherwise different.

Perhaps one solution is to special case this with a separate `StepFunction` tag and a new mode of function selection that can select multiple functions using Stack, Stage and StepFunction tags.

mbarton:
I agree about having unique sets of tags for each lamba. `app` to me feels like it would apply to the whole state machine including all lambdas but I think your solution of having a common prefix is good and fits nicely with the conventions we have already.

I'm a bit wary to add `StepFunctions` as a named tag in Riff-Raff itself. I think that there are other use-cases for updating multiple lambdas eg Cognito User Pool triggers (https://github.com/guardian/investigations-platform/pull/214). Maybe we could extend the `lookupByTags` parameter to fully customise which dynamic injected and static tags are used?

```yaml
lookupByTags:
  dynamic: [Stack, Stage]
  static:
    StepFunctions: pfi-video-pipeline
```

Another possible solution could be to add another (!) parameter: `lookUpFromCloudformationStackByTags`. We would do a `DescribeStacks` filtered by `Stack, App, Stage` to find the Cloudformation stack and then read the function names from the `Resources`.

sihil:
I'm not sure we should bake in an assumption that all functions in a CFN stack are a step function. Targetting a step function directly would perhaps make more sense, but even then you might have one function special cased?

I'd also shy away from making it too customisable. I'd rather make it easy to understand solution for the use case at hand.

Rather than hardcoding a tag could we have a `multiLambdaAppTag` which takes a replacement tag name, the value of which is still the app of the deployment in question?

All feels a bit complicated. It might be easier to have an entirely new deployment type that behaves in a different way.

mbarton:
> It might be easier to have an entirely new deployment type that behaves in a different way.

Yeah that sounds sensible. I was thinking the other day that the state machine (as far as I know) will invoke whatever code is in your lambda  function when it reaches that state so an execution running prior to a deployment would pick up the new code as it goes along which might not be what we want and requires the state JSON to be backwards/forwards compatible. That could be covered by a new deployment type as well, although I'm not 100% sure how.

Maybe have Riff-Raff run CDK to create a new state machine definition referencing a lambda alias related to the build number, like we're doing with our containers in investigations-platform atm.

Either way, more discussion required. Closing for now, thanks for your input @sihil and thanks for the original review @philmcmahon!

PR #587, opened 2020-05-15
Support role based credential provider

Nirvikalpa108:
Currently riffraff uses users (which have access keys and IDs - the same as usernames and passwords) to gain access to other application’s aws accounts so that they can deploy them. If someone gets hold of this, they can use Riffraff which is a security risk. This change switches riffraff to using roles, which only riffraff is allowed to assume.

So far, we've tested our changes on apps in the deploy tools account using riffraff CODE. Next, we will create a Stackset, so that we can deploy a role to all guardian aws accounts.

### DIFF
--- a/magenta-lib/src/main/scala/magenta/Credentials.scala
+++ b/magenta-lib/src/main/scala/magenta/Credentials.scala
@@ -7,6 +7,6 @@ case class KeyRing(apiCredentials: Map[String, ApiCredentials] = Map.empty, riff
   override def toString = apiCredentials.values.toList.mkString(", ")
 }
 
-case class ApiCredentials(service: String, id: String, secret: String, comment: Option[String], role: Option[String] = None) {
+case class ApiCredentials(service: String, id: String, secret: String, comment: Option[String]) {
   override def toString = s"$service:$id${comment.map(c => s" ($c)").getOrElse("")}"
 }
--- a/magenta-lib/src/main/scala/magenta/Credentials.scala
+++ b/magenta-lib/src/main/scala/magenta/Credentials.scala
@@ -1,8 +1,5 @@
 package magenta
 
-import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider
-
-
 case class KeyRing(apiCredentials: Map[String, ApiCredentials] = Map.empty) {
   override def toString = apiCredentials.values.toList.mkString(", ")
 }
--- a/magenta-lib/src/main/scala/magenta/Credentials.scala
+++ b/magenta-lib/src/main/scala/magenta/Credentials.scala
@@ -5,6 +5,6 @@ case class KeyRing(apiCredentials: Map[String, ApiCredentials] = Map.empty) {
   override def toString = apiCredentials.values.toList.mkString(", ")
 }
 
-case class ApiCredentials(service: String, id: String, secret: String, comment: Option[String]) {
+case class ApiCredentials(service: String, id: String, secret: String, comment: Option[String], role: Option[String] = None) {
   override def toString = s"$service:$id${comment.map(c => s" ($c)").getOrElse("")}"
 }
--- a/magenta-lib/src/main/scala/magenta/Credentials.scala
+++ b/magenta-lib/src/main/scala/magenta/Credentials.scala
@@ -3,7 +3,7 @@ package magenta
 import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider
 
 
-case class KeyRing(apiCredentials: Map[String, ApiCredentials] = Map.empty, riffRaffCredentialsProvider: Option[AwsCredentialsProvider] = None) {
+case class KeyRing(apiCredentials: Map[String, ApiCredentials] = Map.empty) {
   override def toString = apiCredentials.values.toList.mkString(", ")
 }
 
--- a/magenta-lib/src/main/scala/magenta/Credentials.scala
+++ b/magenta-lib/src/main/scala/magenta/Credentials.scala
@@ -3,7 +3,7 @@ package magenta
 import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider
 
 
-case class KeyRing(apiCredentials: Map[String, ApiCredentials] = Map.empty, riffRaffCredentialsProvider: AwsCredentialsProvider) {
+case class KeyRing(apiCredentials: Map[String, ApiCredentials] = Map.empty, riffRaffCredentialsProvider: Option[AwsCredentialsProvider] = None) {
   override def toString = apiCredentials.values.toList.mkString(", ")
 }
 
--- a/magenta-lib/src/main/scala/magenta/Credentials.scala
+++ b/magenta-lib/src/main/scala/magenta/Credentials.scala
@@ -1,7 +1,9 @@
 package magenta
 
+import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider
 
-case class KeyRing(apiCredentials: Map[String, ApiCredentials] = Map.empty) {
+
+case class KeyRing(apiCredentials: Map[String, ApiCredentials] = Map.empty, riffRaffCredentialsProvider: AwsCredentialsProvider) {
   override def toString = apiCredentials.values.toList.mkString(", ")
 }
 
--- a/magenta-lib/src/main/scala/magenta/deployment_type/AutoScaling.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/AutoScaling.scala
@@ -126,20 +126,19 @@ object AutoScaling  extends DeploymentType {
     """.stripMargin
   ){ (pkg, resources, target) =>
     implicit val keyRing = resources.assembleKeyring(target, pkg)
-    implicit val artifactClient = resources.artifactClient
-    implicit val stsClient = resources.stsClient
-    val reporter = resources.reporter
     val prefix = S3Upload.prefixGenerator(
-      stack = if (prefixStack(pkg, target, reporter)) Some(target.stack) else None,
-      stage = if (prefixStage(pkg, target, reporter)) Some(target.parameters.stage) else None,
-      packageName = if (prefixPackage(pkg, target, reporter)) Some(pkg.name) else None
+      stack = if (prefixStack(pkg, target, resources.reporter)) Some(target.stack) else None,
+      stage = if (prefixStage(pkg, target, resources.reporter)) Some(target.parameters.stage) else None,
+      packageName = if (prefixPackage(pkg, target, resources.reporter
+      )) Some(pkg.name) else None
     )
     List(
       S3Upload(
         target.region,
-        bucket(pkg, target, reporter),
+        bucket(pkg, target, resources.reporter),
         Seq(pkg.s3Package -> prefix),
-        publicReadAcl = publicReadAcl(pkg, target, reporter)
+        publicReadAcl = publicReadAcl(pkg, target, resources.reporter),
+        resources = resources
       )
     )
   }
--- a/magenta-lib/src/main/scala/magenta/deployment_type/AutoScaling.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/AutoScaling.scala
@@ -127,18 +127,19 @@ object AutoScaling  extends DeploymentType {
   ){ (pkg, resources, target) =>
     implicit val keyRing = resources.assembleKeyring(target, pkg)
     implicit val artifactClient = resources.artifactClient
+    val reporter = resources.reporter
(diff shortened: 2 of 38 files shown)


magenta-lib/src/main/scala/magenta/tasks/AWS.scala
```diff
@@ -33,15 +33,14 @@ import software.amazon.awssdk.services.sts.model.AssumeRoleRequest.Builder
 import scala.annotation.tailrec
 import scala.collection.JavaConverters._
 import scala.util.{Random, Try}
-import magenta.withResource
 import software.amazon.awssdk.services.sts.auth.StsAssumeRoleCredentialsProvider
 import software.amazon.awssdk.services.sts.model.AssumeRoleRequest
 
 object S3 {
-  def withS3client[T](keyRing: KeyRing, region: Region, config: ClientOverrideConfiguration = AWS.clientConfiguration)(block: S3Client => T): T =
+  def withS3client[T](keyRing: KeyRing, region: Region, config: ClientOverrideConfiguration = AWS.clientConfiguration, resources: DeploymentResources)(block: S3Client => T): T =
```
sihil, line 40: I'd consider putting the resources as the first argument here.

magenta-lib/src/main/scala/magenta/tasks/ChangeSetTasks.scala
```diff
@@ -18,9 +19,9 @@ class CreateChangeSetTask(
                            val unresolvedParameters: CloudFormationParameters
 )(implicit val keyRing: KeyRing, artifactClient: S3Client) extends Task {
 
-  override def execute(reporter: DeployReporter, stopFlag: => Boolean) = if (!stopFlag) {
+  override def execute(reporter: DeployReporter, stopFlag: => Boolean, stsClient: StsClient) = if (!stopFlag) {
     CloudFormation.withCfnClient(keyRing, region) { cfnClient =>
-      S3.withS3client(keyRing, region) { s3Client =>
+      S3.withS3client(keyRing, region, stsClient = stsClient) { s3Client =>
         STS.withSTSclient(keyRing, region) { stsClient =>
```
sihil, line 25: S3.withS3client should take a resources now (as it needs both a reporter to grab the deploy ID and also the stsClient). I would suggest that the artifactClient passed in on 19/20 here becomes a DeploymentResources. This is something we know at the time we create the task. Although to be fair it's not completely clear cut as that would provide two routes to the reporter.

sihil: Needs more thought 🤔

philmcmahon: Hey @sihil , thanks for these comments - I only picked them up today. We ended up changing the signature of execute to take a `resources` instead of a `reporter` everywhere (see recent commits) - what do you think?

magenta-lib/src/main/scala/magenta/tasks/AWS.scala
philmcmahon: I wish I saw this last week...have repeated the same pattern in every client builder 😭

magenta-lib/src/main/scala/magenta/tasks/tasks.scala
```diff
@@ -13,19 +13,21 @@ import software.amazon.awssdk.core.internal.util.Mimetype
...
 import software.amazon.awssdk.services.s3.model.{GetObjectRequest, ObjectCannedACL, PutObjectRequest}
+import software.amazon.awssdk.services.sts.StsClient
 
 import scala.util.control.NonFatal
 
 case class S3Upload(
   region: Region,
   bucket: String,
   paths: Seq[(S3Location, String)],
+  resources: DeploymentResources,
```
philmcmahon, line 24: we possibly don't actually need to pass this here as we've gone down the route of changing the signature of `execute` - so all that's needed in this task is an artifactclient (for `resolveMappings` on line 97`)

philmcmahon: update: I've removed this and gone back to having an implicit artifactClient everywhere

philmcmahon:
> Not super clear whether to change the execute signature or pass resources in for each task. Possibly worth an experiment.

We've gone down the `execute` route which has involved a lot of turning `reporter` into `resources.reporter` but not too drastic code changes in most places. I realised part of the way through that it was quite annoying having to construct a full DeploymentResources object when all that `provider` needs is UUID and STS client so ended up creating StsAssumeRoleCredentialsProvider

Nirvikalpa108:
We've tested this by deploying amiable using role credentials and editions using static credentials 👍🏽

magenta-lib/src/main/scala/magenta/tasks/ChangeSetTasks.scala
sihil: I think that's a reasonable choice. I think it makes it a little easier to test the delpoyment types too.

magenta-lib/src/main/scala/magenta/tasks/AWS.scala
sihil: The difficulty with leaving it at the end is losing the behaviour of default parameters at the end of a parameter list. I see you've dealt with that by naming the param and this is only an issue for this one example due to the way the testing is done.

magenta-lib/src/main/scala/magenta/Credentials.scala
```diff
@@ -1,5 +1,7 @@
 package magenta
 
+import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider
+
```
sihil, line 4: Import no longer used.

magenta-lib/src/main/scala/magenta/deployment_type/AutoScaling.scala
```diff
@@ -127,18 +127,18 @@ object AutoScaling  extends DeploymentType {
...
+      )) Some(pkg.name) else None
     )
     List(
       S3Upload(
         target.region,
-        bucket(pkg, target, reporter),
+        bucket(pkg, target, resources.reporter),
         Seq(pkg.s3Package -> prefix),
-        publicReadAcl = publicReadAcl(pkg, target, reporter)
+        publicReadAcl = publicReadAcl(pkg, target, resources.reporter)
```
sihil, line 141: There is no actual change here unless I missed something? I'd leave it as was if you're able.

magenta-lib/src/main/scala/magenta/tasks/Task.scala
```diff
@@ -1,10 +1,12 @@
...
 package tasks
 
+import software.amazon.awssdk.services.sts.StsClient
+
 trait Task {
   // execute this task (should throw on failure)
-  def execute(reporter: DeployReporter, stopFlag: => Boolean)
-  def execute(reporter: DeployReporter) { execute(reporter, stopFlag = false) }
+  def execute(resources: DeploymentResources, stopFlag: => Boolean)
+  def execute(resources: DeploymentResources) { execute(resources, stopFlag = false) }
```
sihil, line 9: Small side note - the second line here is a legacy interface and could be removed from the `Task` definition and the `stopFlag = false` added to the calls from tests. Not really part of this PR though so this comment is just noise 😏

magenta-lib/src/test/scala/magenta/tasks/TasksTest.scala
```diff
@@ -18,11 +18,14 @@ import software.amazon.awssdk.core.client.config.ClientOverrideConfiguration
...
 import software.amazon.awssdk.services.s3.S3Client
 import software.amazon.awssdk.services.s3.model._
+import software.amazon.awssdk.services.sts.StsClient
 
 
 class TasksTest extends FlatSpec with Matchers with MockitoSugar {
   implicit val fakeKeyRing: KeyRing = KeyRing()
-  implicit val reporter: DeployReporter = DeployReporter.rootReporterFor(UUID.randomUUID(), fixtures.parameters())
+  val reporter = DeployReporter.rootReporterFor(UUID.randomUUID(), fixtures.parameters())
+//  val resources = DeploymentResources(reporter, null, mock[S3Client], mock[StsClient])
```
sihil, line 27: superflous comment?

riff-raff/app/deployment/actors/TasksRunner.scala
```diff
@@ -37,7 +37,7 @@ class TasksRunner(stopFlagAgent: Agent[Map[UUID, String]]) extends Actor with Lo
                   log.debug(s"Running task $taskId")
                   deployMetricsProcessor ! TaskStart(uuid, taskId, queueTime, new DateTime())
                   deployReporter.taskContext(task) { taskReporter =>
-                    task.execute(taskReporter, stopFlagAsker)
+                    task.execute(rootResources, stopFlagAsker)
```
sihil, line 40: You're dropping the reporter context here. The context changes here on 28 and also 39 with the calls to `infoContext` and `deployContext`. This creates the nesting in the logs which I suspect may have gone missing in your test runs?

It is somewhat less elegant as you'll need to recreate the DeployResources with the new context. Can't immediately think of a neater way of solving it though, other than passing in the stsClient elsewhere again 🙄

riff-raff/app/deployment/actors/TasksRunner.scala
```diff
@@ -69,5 +69,5 @@ class TasksRunner(stopFlagAgent: Agent[Map[UUID, String]]) extends Actor with Lo
 
 object TasksRunner {
   trait Message
-  case class RunDeployment(uuid: UUID, deployment: DeploymentTasks, rootReporter: DeployReporter, queueTime: DateTime) extends Message
+  case class RunDeployment(uuid: UUID, deployment: DeploymentTasks, rootResources: DeploymentResources, queueTime: DateTime) extends Message
```
sihil, line 72: `rootResources` is a bit of a misnomer here. It makes sense for a reporter to be a rootReporter, less so for a resource. A reporter has a context attached which determines where in the log tree any reports are appended. This ensures that logs for a task are all bundled together.

riff-raff/app/resources/PrismLookup.scala
```diff
@@ -28,8 +28,11 @@ class PrismLookup(config: Config, wsClient: WSClient, secretProvider: SecretProv
     val apiCredentials = data.keys flatMap {
       case key@KeyPattern(service) =>
         data.datum(key, app, stage, stack).flatMap { data =>
-          secretProvider.lookup(service, data.value).map { secret =>
-            service -> ApiCredentials(service, data.value, secret, data.comment)
+          if(service == "aws-roles") {
```
sihil, line 31: This should really be singular but 🤷‍♂️

riff-raff/app/resources/PrismLookup.scala
```diff
@@ -28,8 +28,11 @@ class PrismLookup(config: Config, wsClient: WSClient, secretProvider: SecretProv
...
+            Some(service -> ApiCredentials(service, data.value, " ", data.comment))
+          } else {
+            secretProvider.lookup(service, data.value).map { secret =>
+              service -> ApiCredentials(service, data.value, secret, data.comment)}
```
sihil, line 35: If we have both a role and a static, both are added to the keychain, right? But later in AWS.provider we don't try one and then fallback to the other? I reckon we should either do that fallback trick or single out which one we will use here. Both don't need to be in the keychain if we won't use them both.

Not quite sure how to do that.

philmcmahon:
@sihil thanks for that review.

> the reporter context is not passed through correctly which I think will mean the logs are flattened (deploy context is lost)
This should be resolved now

> the keyring contains both static and role creds but one is preferred over the other - I wonder if we should reflect this preferential logic in the lookup phase? although now I'm typing this I can see failure modes where what has been done is a smart move

Happy to make this change, but also happy to be lazy and not do so! I guess eventually the static credential route will only exist for running locally so we might want to remove those access keys from the prism data file anyway...

> there are no tests to confirm the behaviour of AWS.provider or the PrismLookup as far as I can see - this would be nice to have so we have confidence in the behaviour of this domain specific logic.

We've added tests for these specific function

katebee:
> deploy a role to all guardian aws accounts

It might be best to only roll out the StackSet to the subset of accounts that are used by Digital and already have a RiffRaff user, and talk to the other account owners to discuss integration.

I think some accounts have different allowances of RiffRaff access, and not all accounts currently have a RiffRaff user.

sihil:
@katebee We should only roll out to AWS accounts that use Riff-Raff. No need to put it anywhere else.

In the new world the Riff-Raff role will be consistent across all accounts that do use Riff-Raff. This kind of relies on us doing an extra piece of work to split the CFN executions from the main Riff-Raff role which follows this piece of work.

PR #589, opened 2020-06-30
Add logging for Grafana annotations

stephengeller:
## What does this change?

Grafana has an [annotations](https://grafana.com/docs/grafana/latest/reference/annotations/) feature that allows users to show lines on panels. We would like to add deployment lines to panels so that it is easier for users to correlate deployments with changes in metrics.

This is currently done using a bit of a hacky query (`"Editorial Tools::Integration Tests" AND "Started deploying" AND app:"riff-raff"`) in Grafana, and the log lines do not necessarily dictate the actual start of a deployment.

This PR adds loglines to  

Deploys in Riff-Raff create log entries as shown below, so that Grafana can query them for deployment annotations.

## Images

Before: using the message field in the logs to parse tags
image

New: example loglines (from Riff-Raff CODE deploys)
image

### DIFF
--- a/riff-raff/app/AppComponents.scala
+++ b/riff-raff/app/AppComponents.scala
@@ -14,7 +14,7 @@ import housekeeping.ArtifactHousekeeping
 import lifecycle.ShutdownWhenInactive
 import magenta.deployment_type._
 import magenta.tasks.AWS
-import notification.{DeployFailureNotifications, HooksClient}
+import notification.{DeployFailureNotifications, GrafanaAnnotationLogger, HooksClient}
 import persistence._
 import play.api.ApplicationLoader.Context
 import play.api.db.evolutions.EvolutionsComponents
@@ -154,6 +154,7 @@ class AppComponents(context: Context, config: Config, passwordProvider: Password
     builds,
     targetResolver,
     DeployMetrics,
+    new GrafanaAnnotationLogger,
     hooksClient,
     new SummariseDeploysHousekeeping(config, datastore),
     continuousDeployment,
--- a/riff-raff/app/notification/GrafanaAnnotationLogger.scala
+++ b/riff-raff/app/notification/GrafanaAnnotationLogger.scala
@@ -1,0 +1,49 @@
+package notification
+
+import controllers.Logging
+import lifecycle.Lifecycle
+import magenta.ContextMessage.{FailContext, FinishContext, StartContext}
+import magenta.{DeployParameters, DeployReporter}
+import magenta.Message.Deploy
+import net.logstash.logback.marker.LogstashMarker
+import net.logstash.logback.marker.Markers.appendEntries
+import play.api.MarkerContext
+import rx.lang.scala.Subscription
+
+import scala.collection.JavaConverters._
+
+trait LogMarker {
+  def toLogMarker: LogstashMarker = appendEntries(markerContents.asJava)
+
+  def markerContents: Map[String, Any]
+}
+
+class GrafanaAnnotationLogger extends Lifecycle with Logging {
+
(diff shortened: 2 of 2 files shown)


rtyley:
Ah cool! Prout has a similar feature for Librato (added in https://github.com/guardian/prout/commit/cc34e5f6e6e842b2cf3e22c2f4c8ecff3e8dc11f) using their 'annotations' API- it looks like this:

image

...the time-range shown is the time between the PR being merged and the commit-id being observed by Prout on the checkpoint url.


The log-lines approach in this PR a nice approach - shows how flexible Grafana is.

stephengeller:
> Ah cool! Prout has a similar feature for Librato (added in [guardian/prout@cc34e5f](https://github.com/guardian/prout/commit/cc34e5f6e6e842b2cf3e22c2f4c8ecff3e8dc11f)) using their 'annotations' API- it looks like this:
> 
> image
> 
> ...the time-range shown is the time between the PR being merged and the commit-id being observed by Prout on the checkpoint url.
> 
> The log-lines approach in this PR a nice approach - shows how flexible Grafana is.

Nice! Where does that information get sourced from? Does Librato query an endpoint or does it get metrics shipped to it?

rtyley:
> Nice! Where does that information get sourced from? Does Librato query an endpoint or does it get metrics shipped to it?

[Prout sends the data to Librato](https://github.com/guardian/prout/blob/0ed5b38249226dc4e8336c410af969c6d9d77022/app/lib/librato/LibratoApiClient.scala#L22-L34) using the Librato API.

stephengeller:
> > Nice! Where does that information get sourced from? Does Librato query an endpoint or does it get metrics shipped to it?
> 
> [Prout sends the data to Librato](https://github.com/guardian/prout/blob/0ed5b38249226dc4e8336c410af969c6d9d77022/app/lib/librato/LibratoApiClient.scala#L22-L34) using the Librato API.

Awesome. I like querying ES for this, but if it gets too slow it might be worth looking into [Grafana's equivalent](https://grafana.com/docs/grafana/latest/http_api/annotations/) to this.

philmcmahon:
Would it be possible to use the [riffraff api](https://riffraff.gutools.co.uk/docs/riffraff/api.md) for this? Looks like it has a history endpoint which might do the job but I'm not sure.

stephengeller:
> Would it be possible to use the [riffraff api](https://riffraff.gutools.co.uk/docs/riffraff/api.md) for this? Looks like it has a history endpoint which might do the job but I'm not sure.

I think it _could_, but I think it would require creating additional infrastructure (maybe a lambda that ships the events from Riff-Raff to Grafana's DB) that I feel is maybe unnecessary work given we can just query Elasticsearch directly?

image

sihil:
Another approach might be a plugin for Grafan that queries the Riff-Raff API directly but this sticks with vanilla Grafana and feels like the simplest solution. Pushing directly into Grafana annotations would be another nice alternative and might be something we could switch to. We'd need to figure out how to use the flat tags in grafana to reflect KV pairs.

I'm generally in favour of this approach for now as a simple solution but we did pair on it.

stephengeller:
> Another approach might be a plugin for Grafan that queries the Riff-Raff API directly but this sticks with vanilla Grafana and feels like the simplest solution. Pushing directly into Grafana annotations would be another nice alternative and might be something we could switch to. We'd need to figure out how to use the flat tags in grafana to reflect KV pairs.
> 
> I'm generally in favour of this approach for now as a simple solution but we did pair on it.

I agree. I think that, for our uses, this solution is simple and effective. Developing a Grafana Riff-Raff plugin may well necessary if/when Grafana gets more widely adopted, I'd imagine mainly if we might want to take some pressure off Elasticsearch to provide the annotations via queries.

PR #591, opened 2020-07-02
Ensure that tags are acutally consumed

sihil:
## What does this change?
Tags are currently being added to a request object that is discarded. This fixes the issue and adds some tests to make sure we are constructing a fairly complex object correctly.

This fixes a feature that hasn't worked for a while. Should we notify users or understand the impact of this? It should just mean that a whole load of resources become correctly tagged along with the stacks themselves. However this is likely to change stuff at scale and might have unintended consequences?


sihil:
@akash1810 I suspect this might be somewhat painful to roll out but flagging. Feel free to close.

nicl:
Closing for now as it's been a while. We can reopen if needed though.

PR #592, opened 2020-07-02
Use service role for Cloudformation updates

sihil:
## What does this change?
Uses a less privileged role for most work but then uses a CFN service role when creating or updating CFN stacks.

This uses roles as defined in https://github.com/guardian/deploy-tools-platform/pull/209 

RiffRaff previously used a pile of static credentials to access other accounts. In #587 is was changed to assume a role to access other accounts but this had to be very powerful due to CFN stacks being created and updated.

This PR goes one step further. Riff-Raff assumes a least privileged role to make changes in another account and then uses a CFN service role to allow wide ranging changes to be made to resources.

Deploys still work and we can delete static credentials.

### DIFF
--- a/magenta-lib/src/main/scala/magenta/Credentials.scala
+++ b/magenta-lib/src/main/scala/magenta/Credentials.scala
@@ -4,6 +4,12 @@ case class KeyRing(apiCredentials: Map[String, ApiCredentials] = Map.empty) {
   override def toString = apiCredentials.values.toList.mkString(", ")
 }
 
-case class ApiCredentials(service: String, id: String, secret: String, comment: Option[String]) {
+trait ApiCredentials {
+  val service: String
+  val id: String
+  val comment: Option[String]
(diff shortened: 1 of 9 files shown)


magenta-lib/src/main/scala/magenta/Credentials.scala
```diff
@@ -4,6 +4,12 @@ case class KeyRing(apiCredentials: Map[String, ApiCredentials] = Map.empty) {
...
+  val service: String
+  val id: String
+  val comment: Option[String]
   override def toString = s"$service:$id${comment.map(c => s" ($c)").getOrElse("")}"
-}
\ No newline at end of file
+}
+
+case class ApiStaticCredentials(service: String, id: String, secret: String, comment: Option[String]) extends ApiCredentials
+case class ApiRoleCredentials(service: String, id: String, comment: Option[String]) extends ApiCredentials
```
philmcmahon, line 15: great improvement

PR #593, opened 2020-07-03
Update node-riffraff-artifact repo links

tjmw:
## What does this change?

Tiny documentation change: while browsing the docs I noticed that these link to a legacy library, `node-riffraff-artefact` (with an e), instead of `node-riffraff-artifact` (with an i).

Ever-so-minor improvement to DevEx 👩‍💻


philmcmahon:
thanks Tom!

PR #594, opened 2020-07-08
Ensure scheduled deploys are matched exactly.

alexduf:
## What does this change?
We've had an issue in the mobile team where some of our EC2 image got out of date even though we had a weekly scheduled deployment.
Upon investigating we found out two of our projects (`mobile-n10n:notificationworkerlambda` and `mobile-n10n:notification`) have a name including the other, and riff-raff could pick the wrong one.

@twrichards Quickly pointed out the potential use of the `isExactMatchProjectName` flag.

If we deploy the right project.

## What is the cost of failure?
Riff-raff fails to deploy on a schedule.

## Images
YES

### DIFF
--- a/riff-raff/app/schedule/DeployJob.scala
+++ b/riff-raff/app/schedule/DeployJob.scala
@@ -62,7 +62,8 @@ object DeployJob extends Logging with LogAndSquashBehaviour {
     } else {
       val filter = DeployFilter(
         projectName = Some(projectName),
-        stage = Some(stage)
+        stage = Some(stage),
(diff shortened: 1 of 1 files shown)


twrichards:
@alexduf probably worth a department email on this one to encourage teams to review their list of scheduled deploys, as it will probably change the behaviour of other teams scheduled deploys (for example you'll note in the description of https://github.com/guardian/riff-raff/pull/556 which introduced `isExactMatchProjectName` that the `identity` is both its own project and prefix for other projects and I imagine there are countless other examples)

PR #595, opened 2020-07-30
Disable automatic AMI update temporarily.

alexduf:
This is following a department wide incident where deploying the latest AMI prevent the autoscaling group from scaling up.

I'm not sure I have covered all the cases nor am I sure it's the right approach, please provide feedback.

## What does this change?
This temporarily disable automatic AMI upgrades on riff-raff to prevent deploying the latest AMI which stops our servers from booting.

## How to test
Unit tests + deployment on CODE

no down time 🤞 

Could potentially break all deployments. Needs to be carefully tested on CODE


alexduf:
Closing as trying a different approach

PR #599, opened 2020-08-21
add warning message when someone deploys a branch called "master"

johnduffell:
Since the name "master" is not inclusive, it makes sense to have regular reminders so as not to normalise it.
Although most people don't look at riffraff on a daily basis, it is a good place to put it back in people's minds whenever they do.

This will highlight the offending deploys and put a barber pole pattern on them, which should attract people's attention.

Tested in CODE
image
--
image

### DIFF
--- a/riff-raff/app/deployment/actors/DeployGroupRunner.scala
+++ b/riff-raff/app/deployment/actors/DeployGroupRunner.scala
@@ -129,6 +129,7 @@ class DeployGroupRunner(
   def receive = {
     case Start =>
       try {
+        warnDeprecatedBranch()
         self ! ContextCreated(createContext)
         self ! StartDeployment
       } catch {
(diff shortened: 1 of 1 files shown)


shtukas (approved):

philmcmahon:
Sorry I've only just spotted this, not great devxing! But I am worried this is going to result in warnings on every project and have the 'cry wolf' effect of people not paying attention to warnings. Maybe it would be better to wait until we have a majority of projects migrated before implementing this?

philmcmahon:
(though maybe it will mean we get rid of main really quickly so perhaps best to leave it!) 

Might be worth waiting for github to release the auto migration stuff Akash shared the link to though....

philmcmahon:
Update! John and I just spoke about this and decied to leave it in for now, but if we get a chance maybe change the history page styling for this warning so that it is more obvious what the warning is about, perhaps for the sake of someone looking at the history to try debug an incident.

Mullefa:
Hey, I'm all for the sentiment of this, but am concerned that logging it as a warning could hide issues with the build itself:

Screenshot 2020-09-07 at 10 44 17

Could we could consider logging it at level `INFO` instead?

johnduffell:
@Mullefa if people are finding the disadvantage of not being able to distinguish the different types of warning outweighs the benefit of denormalising using the word "master" then I'm happy to this to be reverted.
I find it very hard to compare the two, in fact the whole task of renaming branches is hard to work out the priority against OKR work in my mind.
I can totally see how annoying a page full of warnings is and it rather loses its effect when it's wall to wall, but equally, if you are triggered by seeing such words, maybe the effect is similarly bad, so at least this way the pain is shared with everyone.
I would be happy with alternative styling, as long as it expresses the sentiment that this word is not ok, we just didn't fix it yet.  But unfortunately it would be more work than I have time for.
Happy to chat through it, given you are the second person to suggest it's a bit much, I am happy to see it reverted and use other approaches to draw this to people's attention.

PR #601, opened 2020-09-07
make "master" branch deploy a info rather than warning

johnduffell:
according to #599 I added a warning for deploying "master" as it is not inclusive.

Since we use warnings for problems that might cause unexpected behaviour in some situations, the fact that the majority of deploys have a warning anyway, makes it less likely that we will spot other problems if they arise.  This is because you have to click in to see the actual warning, and people might assume it's just because of the branch name.

With the current approach there is a trade off between us saying clearly "this word is not ok, but we haven't fixed it yet" and being able to spot a minority of problems which would be more obvious as a result.

Of course the best approach is if we implemented different styling (more work) then this would be more clear.  However I don't have time for this at present.

What do people think? - this PR is to check if consensus is overall people prefer to do without the warnings and just make it a normal log message.

### DIFF
--- a/riff-raff/app/deployment/actors/DeployGroupRunner.scala
+++ b/riff-raff/app/deployment/actors/DeployGroupRunner.scala
@@ -174,7 +174,7 @@ class DeployGroupRunner(
   private def warnDeprecatedBranch(): Unit = {
     val maybeBranch: Option[String] = record.metaData.get("branch")
     if (maybeBranch.contains("master"))
-      rootReporter.warning("branch name 'master' is not inclusive, please rename - see https://docs.google.com/document/d/1nqC5Uk6n3y7S79m1yiG6oqT8NtXWpDcsS-uLI_RK7Mg/edit#")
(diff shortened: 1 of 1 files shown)


Mullefa:
Hey John, firstly just want to say I think it's commendable the effort and thought you've put into making the change from master to main. Thanks!

I was one of the people concerned that all the "warnings encountered" in the summarised builds was making it hard to distinguish between branch name issues and build issues. But would also be keen to see if there was a consensus on this?

Another option could be to the render the information on the deployment rather than in the logs themselves:

Screenshot 2020-09-07 at 14 33 22

Whatever's decided on, I'm happy to try and find some time to help support this initiative and contribute to implementing the changes.

akash1810:
I'd suggest it's easier for teams to perform the migration (it needs to be done at some point anyway!) than to find the best UX for this message. Having it as a `warn` in it's garish red provides incentive too 😄 .

Tools such as https://github.com/guardian/master-to-main provides automation for a lot of the migration process.

Mullefa:
Hmm, the suggestion wasn't about trying to find the best UX message; rather a temporary mitigation whilst we transition from master to main (regrettably taking longer than we'd like on CAPI due to the amount of time sensitive work at the moment). 

However, it appears the consensus is that the warnings are fine as is. Happy to take on board the suggestion, and thanks for the link to the tool.

PR #602, opened 2020-09-08
decorate log line with extra information

akash1810:
## What does this change?

Log the branch being deployed as a marker. This is primarily to help understand our progress of moving from `master` to `main`.

Logging this information as a marker means we can create graphs within the Central ELK stack.

A graph of branch name used over time would allow us to track our progress of `master` to `main`; `master` reaching 0 is a proxy for our success on this work.

## How to test

- Deploy to CODE
- Perform a deploy
- Witness log lines have extra markers

We have visibility of which branch is being deployed within the logs.

I think this log line occurs pretty close to the start of a deploy. That is, we don't know if the deploy has succeeded or not. I think this is fine as we only really care about deploying intent for this particular use case.

## Images

**Log filter (logs from RIffRaff CODE)**
image

**Dashboard (logs from RIffRaff CODE)**
image

### DIFF
--- a/riff-raff/app/deployment/Deployments.scala
+++ b/riff-raff/app/deployment/Deployments.scala
@@ -66,7 +66,12 @@ class Deployments(deploymentEngine: DeploymentEngine,
     val uuid = java.util.UUID.randomUUID()
     val hostNameMetadata = Map(Record.RIFFRAFF_HOSTNAME -> java.net.InetAddress.getLocalHost.getHostName)
     val record = deployRecordFor(uuid, params) ++ hostNameMetadata
-    val mc = MarkerContext(appendEntries(record.metaData.asJava))
+
+    // information for creating a master to main progress dashboard
+    val markers = Map(
+      "branch" -> record.metaData.getOrElse("branch", "unknown")
+    )
+    val mc = MarkerContext(appendEntries(markers.asJava))
     log.info(s"Creating deploy record for $params")(mc)
     library send { _ + (uuid -> Agent(record)) }
     documentStoreConverter.saveDeploy(record)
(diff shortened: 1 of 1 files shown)


riff-raff/app/deployment/Deployments.scala
```diff
@@ -60,10 +63,11 @@ class Deployments(deploymentEngine: DeploymentEngine,
   def getStopFlag(uuid: UUID): Boolean = deploymentEngine.getDeployStopFlag(uuid)
 
   def create(params: DeployParameters): Record = {
-    log.info(s"Creating deploy record for $params")
     val uuid = java.util.UUID.randomUUID()
     val hostNameMetadata = Map(Record.RIFFRAFF_HOSTNAME -> java.net.InetAddress.getLocalHost.getHostName)
     val record = deployRecordFor(uuid, params) ++ hostNameMetadata
+    val mc = MarkerContext(appendEntries(record.metaData.asJava))
+    log.info(s"Creating deploy record for $params")(mc)
```
akash1810, line 70: This is the meaningful change in the branch. The rest is my IDE being helpful.

sihil: 👍 for not creating an extra log line. 

Am I right in thinking that this will just splat the metadata map keys into elk? Do we want to do that (given the chat recently about ES common schema etc?

sihil: Is `labels` the right place for this? https://www.elastic.co/guide/en/ecs/current/ecs-base.html

akash1810: I don't know enough about ECS to be confident in this response...

I think ECS is a sort of all or nothing deal. That is, we'd need the entire app/log producers in the index to adhere to the schema to see the benefits of it. With this in mind, I'd be tempted to push that work to a later time as it's quite big.

sihil (approved):

PR #604, opened 2020-09-22
Support for GCS deployment type

bfil:
## What does this change?

Adds a task to deploy artifacts to Google Cloud Storage.

## How to test

Use the `gcp-gcs` deployment type in `riff-raff.yml`.

Deployments to GCS succeeds.

None, since the task is new.

### DIFF
--- a/magenta-lib/src/main/scala/magenta/deployment_type/GCS.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/GCS.scala
@@ -143,12 +143,10 @@ object GCS extends DeploymentType {
       ))
       List(
         GCSUpload(
-          target.region,
           bucket = bucketName,
           paths = Seq(pkg.s3Package -> prefix),
           cacheControlPatterns = cacheControl(pkg, target, reporter),
-          extensionToMimeType = mimeTypes(pkg, target, reporter),
-          publicReadAcl = publicReadAcl(pkg, target, reporter)
+          publicReadAcl = publicReadAcl(pkg, target, reporter),
         )
       )
     }
--- a/magenta-lib/src/main/scala/magenta/deployment_type/GCS.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/GCS.scala
@@ -1,0 +1,150 @@
+package magenta.deployment_type
+
+import magenta.Datum
+import magenta.deployment_type.param_reads.PatternValue
+import magenta.tasks.gcp.GCSUpload
+
+object GCS extends DeploymentType {
+  val name = "gcp-gcs"
+  val documentation = "For uploading files into a GCS bucket."
+
+  val prefixStage = Param("prefixStage",
+    "Prefix the GCS bucket key with the target stage").default(true)
+  val prefixPackage = Param("prefixPackage",
+    "Prefix the GCS bucket key with the package name").default(true)
+  val prefixStack = Param("prefixStack",
+    "Prefix the GCS bucket key with the target stack").default(true)
+  val pathPrefixResource = Param[String]("pathPrefixResource",
+    """Deploy Info resource key to use to look up an additional prefix for the path key. Note that this will override
+       the `prefixStage`, `prefixPackage` and `prefixStack` keys - none of those prefixes will be applied, as you have
+       full control over the path with the resource lookup.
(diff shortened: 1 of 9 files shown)


magenta-lib/src/main/scala/magenta/tasks/gcp/GCSTasks.scala
```diff
@@ -0,0 +1,130 @@
...
+import magenta.{DeploymentResources, KeyRing, Region, Stack, Stage}
+import magenta.artifact.{S3Location, S3Object, S3Path}
+import magenta.deployment_type.param_reads.PatternValue
+import magenta.tasks.Task
+import software.amazon.awssdk.core.internal.util.Mimetype
+import software.amazon.awssdk.services.s3.S3Client
+import software.amazon.awssdk.services.s3.model.GetObjectRequest
+
+case class GCSUpload(
+  region: Region,
```
sihil, line 21: region isn't used so could be removed.

magenta-lib/src/main/scala/magenta/deployment_type/GCS.scala
```diff
@@ -0,0 +1,158 @@
...
+      |      },
+      |      {
+      |        "pattern": ".*",
+      |        "value": "max-age=3600"
+      |      }
+      |    ]
+    """.stripMargin
+  )
+
+  val mimeTypes = Param[Map[String,String]]("mimeTypes",
```
sihil, line 81: As mentioned elsewhere this is not actually consumed by the task at present so might be best to remove.

magenta-lib/src/main/scala/magenta/tasks/gcp/GCSTasks.scala
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    val withClient = withClientFactory(keyRing, resources)
+    withClient { client =>
+
+      resources.reporter.verbose(s"Starting transfer of ${fileString(objectMappings.size)} ($totalSize bytes)")
+      transfers.zipWithIndex.par.foreach { case (transfer, index) =>
+        logger.debug(s"Transferring $transfer")
+        index match {
+          case x if x < 10 => resources.reporter.verbose(s"Transferring $transfer")
```
sihil, line 78: the `detailedLoggingThreshold` is intended to be used in place of the `10` here.

magenta-lib/src/main/scala/magenta/tasks/gcp/GCSTasks.scala
```diff
@@ -0,0 +1,130 @@
...
+case class GCSUpload(
+  region: Region,
+  bucket: String,
+  paths: Seq[(S3Location, String)],
+  cacheControlPatterns: List[PatternValue] = Nil,
+  extensionToMimeType: Map[String,String] = Map.empty,
+  publicReadAcl: Boolean = false,
+  detailedLoggingThreshold: Int = 10,
+)(implicit val keyRing: KeyRing, artifactClient: S3Client,
+  withClientFactory: (KeyRing, DeploymentResources) => (Storage => Unit) => Unit = GCS.withGCSClient[Unit]) extends Task with Loggable {
```
sihil, line 29: This complicated `withClientFactory` signature exists so that a mock can be provided in tests. If you don't plan to write a test then this could be simplified.

sihil (commented):
Hi Bruno, great work!
I've not tested the functionality, just done a code review.
Have made a few comments about deployment type and function parameters that aren't actually hooked up - suggest they are either removed or hooked up as appropriate. It would also be good to see a couple of simple tests as there is some relatively complex code here which would be great to avoid breaking in the future. There are some tests for the S3 upload that you could adapt.
Thanks again for tackling this.

katebee:
What are we intending to deploy to GCS?

magenta-lib/src/main/scala/magenta/deployment_type/GCS.scala
sihil: This isn't used so the parameter can be removed.

sihil (approved):
LGTM. Thanks for adding this @bfil

sihil:
@katebee I think DBT and Airflow (apps that control/run data tech EMR jobs/graphs).

PR #605, opened 2020-09-23
Update to more inclusive language.

akash1810:
We've recently updated this repo to use `main` as the default branch instead of `master`.

This PR updates some files in the repo to reflect this. I've stricken through `master` to imply its legacy. These files are documentation, until all our repos have been migrated, the docs should support legacy default branches.

Note, some tests still reference `master`. I'm not sure if it is worth updating those too?
﻿


sihil:
Just get rid of it instead of strikethrough?

akash1810:
> Just get rid of it instead of strikethrough?

Have opted to rephrase as "default branch" to make it as generic as possible.

sihil (approved):

PR #607, opened 2020-10-13
Add activeTab parameter from querystring in order to preserve tab on page refresh

jamie-lynch:
## What does this change?
This PR adds the `activeTab` parameter in the call when using the `tabbable` snippet. This value is taken from the appropriate query parameter. This allows the active tab to be preserved across page refreshes.

## How to test
On PROD navigate to `/deployinfo/hosts` and/or `/deployinfo/data` and select a tab that isn't the first. Refresh the page and you'll see that the first tab becomes active again. 
Repeat on the current branch and you should see that the previously selected tab is shown still.

The selected tab is persisted across page refreshes.


philmcmahon (approved):
👍

PR #608, opened 2020-10-19
Show loading and error status for auto refreshing components

jamie-lynch:
## What does this change?
This PR updates the `auto-refresh` script and the views that make use of it in order that loading and error messages are shown around the auto update process. Those components that auto refresh now contain `loading`, `error` and `content` div tags. While the request is loading, the `loading` div is shown. If the request results in an error, the `error` div is displayed. The current content is left unchanged. If the request is successful then the `error` div is hidden and the `content` div is populate with the result of the request. 

The inspiration for this change was confusing behaviour when an error occurred on the `/deployment/history` page where when an error occurred the loading message continued to be shown with no change.

## How to test
You can test the loading message by navigating to any page that auto refreshes. At the point where a new request is fired, the load message should be shown. To test the error messages, while running locally make a change within the API to ensure an error is returned and you should see the error `div` being displayed with a generic error message.

Loading and error messages are displayed as required.

## Images
history

### DIFF
--- a/riff-raff/app/assets/javascripts/auto-refresh.coffee
+++ b/riff-raff/app/assets/javascripts/auto-refresh.coffee
@@ -34,12 +34,23 @@ enableRefresh = (interval=1000) ->
       $('[data-ajax-refresh]').each ->
         if $(".ajax-refresh-disabled").length == 0
           divBottomWasInView = bottomInView($(this).get(-1))
-          $(this).load(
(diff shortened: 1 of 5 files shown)


stephengeller (approved):
Looks good to me!

philmcmahon:
👍 excellent jQuerying 😄

stephengeller (approved):

PR #610, opened 2020-12-03
Update the docs for aws-s3 param publicReadAcl

tjmw:
## What does this change?

We ran into an issue using the `aws-s3` deployment type where the default value for publicReadAcl (true) doesn't play nicely with the default policy for new buckets, which is to restrict all public access. Update the docs to specify that if your bucket blocks public access then you should set this param to false.

Slightly clearer docs == slightly happier developers 👩‍💻 👨‍💻

### DIFF
--- a/magenta-lib/src/main/scala/magenta/deployment_type/S3.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/S3.scala
@@ -34,7 +34,11 @@ object S3 extends DeploymentType {
   )
 
   val publicReadAcl = Param[Boolean]("publicReadAcl",
-    "Whether the uploaded artifacts should be given the PublicRead Canned ACL. (Default is true!)"
+    """
+      |Whether the uploaded artifacts should be given the PublicRead Canned ACL. (Default is true!) Note that the
+      |AWS default S3 bucket setting is to block all public access, which doesn't play nicely with `publicReadAcl`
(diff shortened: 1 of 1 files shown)


magenta-lib/src/main/scala/magenta/deployment_type/S3.scala
```diff
@@ -34,7 +34,11 @@ object S3 extends DeploymentType {
   )
 
   val publicReadAcl = Param[Boolean]("publicReadAcl",
-    "Whether the uploaded artifacts should be given the PublicRead Canned ACL. (Default is true!)"
+    """
+      |Whether the uploaded artifacts should be given the PublicRead Canned ACL. (Default is true!) Note that the
+      |default bucket setting is to block all public access, which doesn't play nicely with this parameter when set
```
adamnfish, line 39: I'd mention AWS here. We're using the word default to mean two different things (default in Riff Raff, default for S3 buckets in AWS). We're in RiffRaff land so we can help the reader by signposting the latter.

tjmw: Great shout, I'll update this 👍

tjmw: (Hopefully) improved/fixed this in 1bcc2df9e03d66493198c26dac8809e177f803c6.

adamnfish (approved):
LGTM!

PR #611, opened 2021-01-15
Prepare to send notifications if scheduled deploys fail to start

jacobwinch:
## What does this change?

This change refactors the code so that we are able to send failure notifications if scheduled deploys fail to start.

At the moment, we will not actually send notifications. Instead, we will log an event whenever we _would have_ notified a team, so that we can collect data on the common failure scenarios and work out how to resolve them.

Once we've done this, it should be trivial to start sending notifications in these scenarios.

## How to test

Prior to commenting out the line which sends the notifications to Anghammarad, we confirmed that this was working as expected.

We've also tested that deploy failure notifications for other scenarios (e.g. a scheduled deploy which fails after starting due to lack of capacity) still results in a message being sent to Anghammarad.

Existing failure notifications should continue to work as before. The logs should contain information which helps us to decide when to send alerts to teams.

This doesn't impact upon the behaviour of the app, so should be low risk.

### DIFF
--- a/riff-raff/app/notification/DeployFailureNotifications.scala
+++ b/riff-raff/app/notification/DeployFailureNotifications.scala
@@ -35,16 +35,32 @@ class DeployFailureNotifications(config: Config,
     prefix + path.url
   }
 
+  def getAwsAccountIdTarget(target: ci.Target, parameters: DeployParameters, uuid: UUID): Option[Target] = {
+    Try {
+      val keyring = lookup.keyRing(parameters.stage, MagentaApp(target.app), MagentaStack(target.stack))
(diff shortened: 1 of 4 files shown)


riff-raff/app/notification/DeployFailureNotifications.scala
```diff
@@ -31,15 +31,15 @@ class DeployFailureNotifications(config: Config,
   lazy private val prefix = config.urls.publicPrefix
 
   def url(uuid: UUID): String = {
-    val path = routes.DeployController.viewUUID(uuid.toString,true)
+    val path = routes.DeployController.viewUUID(uuid.toString, verbose = true)
```
stephengeller, line 34: This is just to appease IntelliJ's Scala linter

philmcmahon (approved):
LGTM 👍🏻

PR #614, opened 2021-02-09
Enable failure notifications if scheduled deploys fail to start

jacobwinch:
## What does this change?

This PR continues the work started in https://github.com/guardian/riff-raff/pull/611. We have identified the common reasons for scheduled deploys failing to start and are now ready to start sending notifications in these scenarios. 

Where there is a clearly identifiable owner and set of recovery actions, we will notify teams directly.

However, some alerts will be sent to the maintainers of Riff-Raff, rather than the teams who 'own' the scheduled deployments. There are two reasons for this:

1. In some scenarios teams are not actually empowered to fix the problems for themselves (e.g. 'stuck' deploys). Consequently these alerts will also be sent to Riff-Raff maintainers so that we can try to fix the underlying problem(s).
2. In some scenarios (e.g. configuration issues) we lack information about who to send alerts to. Consequently these alerts will also be sent to Riff-Raff maintainers so that we can try to (manually) identify affected users and help them to resolve their problem.

Both of these scenarios should be pretty rare (once the initial backlog of cases has been cleared), so I don't anticipate it generating too much work for the Developer Experience team.

There are also some scenarios where we have intentionally decided _not to_ notify anyone:

1. [Scheduled deployments are disabled](https://github.com/guardian/riff-raff/blob/main/riff-raff/app/schedule/DeployJob.scala#L57) - we assume that this is an intentional choice, so we do not notify about it.
1. [Deployments are switched off via RiffRaff](https://github.com/guardian/riff-raff/blob/main/riff-raff/app/deployment/Deployments.scala#L40) - again, we assume this would only happen in rare, intended scenarios. I don't think notifications would be helpful here!
1. Scheduled deployments fail to start as they are [in conflict with restrictions](https://github.com/guardian/riff-raff/blob/main/riff-raff/app/deployment/Deployments.scala#L38) - in my opinion this is a [bug](https://trello.com/c/kBjw5ZyR/526-make-restrictions-and-scheduled-deploys-compatible) we should fix, rather than a problem that we should notify teams about.

## How to test

I'll deploy this change to `CODE` and test it via manually creating some failure scenarios.

* Teams should find it easier to keep their AMIs up to date, as they should be alerted when scheduled deploys fail to start.

* There is a risk that we end up sending (or receiving) too many alerts. If this becomes too noisy, we might need to consider reverting this PR, or rethinking the logic slightly.

### DIFF
--- a/riff-raff/app/deployment/model.scala
+++ b/riff-raff/app/deployment/model.scala
@@ -1,7 +1,6 @@
 package deployment
 
 import java.util.UUID
-
 import com.gu.googleauth.UserIdentity
 import controllers.ApiKey
 import magenta.ContextMessage._
@@ -17,7 +16,12 @@ case object ContinuousDeploymentRequestSource extends RequestSource
 case class ApiRequestSource(key: ApiKey) extends RequestSource
 case object ScheduleRequestSource extends RequestSource
 
-case class Error(message: String) extends AnyVal
+sealed trait ScheduledDeployError
+case object NoDeploysFoundForStage extends ScheduledDeployError
+case object SkippedDueToPreviousFailure extends ScheduledDeployError
+case object SkippedDueToPreviousWaitingDeploy extends ScheduledDeployError
+
+case class Error(message: String, scheduledDeployError: Option[ScheduledDeployError] = None)
 
 object Record {
   val RIFFRAFF_HOSTNAME = "riffraff-hostname"
--- a/riff-raff/app/deployment/model.scala
+++ b/riff-raff/app/deployment/model.scala
@@ -16,12 +16,23 @@ case object ContinuousDeploymentRequestSource extends RequestSource
 case class ApiRequestSource(key: ApiKey) extends RequestSource
 case object ScheduleRequestSource extends RequestSource
 
-sealed trait ScheduledDeployError
-case class NoDeploysFoundForStage(projectName: String, stage: String) extends ScheduledDeployError
-case class SkippedDueToPreviousFailure(failedDeployRecord: Record) extends ScheduledDeployError
-case class SkippedDueToPreviousWaitingDeploy(waitingDeployRecord: Record) extends ScheduledDeployError
+sealed trait RiffRaffError {
+  val message: String
+}
 
-case class Error(message: String, scheduledDeployError: Option[ScheduledDeployError] = None)
+case class Error(message: String) extends RiffRaffError
+
+sealed trait ScheduledDeployNotificationError extends RiffRaffError
+case class NoDeploysFoundForStage(projectName: String, stage: String) extends ScheduledDeployNotificationError {
+  val message = s"A scheduled deploy didn't start because Riff-Raff has never deployed $projectName to $stage before. " +
+    "Please inform the owner of this schedule as it's likely that they have made a configuration error."
+}
+case class SkippedDueToPreviousFailure(failedDeployRecord: Record) extends ScheduledDeployNotificationError {
+  val message = s"Scheduled Deployment for ${failedDeployRecord.parameters.build.projectName} to ${failedDeployRecord.parameters.stage.name} didn't start because the most recent deploy failed."
+}
+case class SkippedDueToPreviousWaitingDeploy(waitingDeployRecord: Record) extends ScheduledDeployNotificationError {
+  val message = s"Scheduled Deployment for ${waitingDeployRecord.parameters.build.projectName} to ${waitingDeployRecord.parameters.stage.name} failed to start as a previous deploy was still waiting to be deployed."
+}
 
 object Record {
   val RIFFRAFF_HOSTNAME = "riffraff-hostname"
--- a/riff-raff/app/deployment/model.scala
+++ b/riff-raff/app/deployment/model.scala
@@ -17,9 +17,9 @@ case class ApiRequestSource(key: ApiKey) extends RequestSource
 case object ScheduleRequestSource extends RequestSource
 
 sealed trait ScheduledDeployError
-case object NoDeploysFoundForStage extends ScheduledDeployError
-case object SkippedDueToPreviousFailure extends ScheduledDeployError
-case object SkippedDueToPreviousWaitingDeploy extends ScheduledDeployError
+case class NoDeploysFoundForStage(projectName: String, stage: String) extends ScheduledDeployError
+case class SkippedDueToPreviousFailure(failedDeployRecord: Record) extends ScheduledDeployError
+case class SkippedDueToPreviousWaitingDeploy(stuckDeployRecord: Record) extends ScheduledDeployError
 
 case class Error(message: String, scheduledDeployError: Option[ScheduledDeployError] = None)
 
--- a/riff-raff/app/deployment/model.scala
+++ b/riff-raff/app/deployment/model.scala
@@ -19,7 +19,7 @@ case object ScheduleRequestSource extends RequestSource
 sealed trait ScheduledDeployError
 case class NoDeploysFoundForStage(projectName: String, stage: String) extends ScheduledDeployError
 case class SkippedDueToPreviousFailure(failedDeployRecord: Record) extends ScheduledDeployError
-case class SkippedDueToPreviousWaitingDeploy(stuckDeployRecord: Record) extends ScheduledDeployError
+case class SkippedDueToPreviousWaitingDeploy(waitingDeployRecord: Record) extends ScheduledDeployError
 
 case class Error(message: String, scheduledDeployError: Option[ScheduledDeployError] = None)
 
--- a/riff-raff/app/notification/DeployFailureNotifications.scala
+++ b/riff-raff/app/notification/DeployFailureNotifications.scala
@@ -5,15 +5,15 @@ import ci.{ContinuousDeployment, TargetResolver}
 import com.gu.anghammarad.Anghammarad
 import com.gu.anghammarad.models._
 import conf.Config
-import controllers.{Logging, routes}
+import controllers.Logging
 import deployment.Error
-import deployment.{NoDeploysFoundForStage, ScheduledDeployError, SkippedDueToPreviousFailure, SkippedDueToPreviousWaitingDeploy}
 import lifecycle.Lifecycle
 import magenta.Message.Fail
 import magenta.deployment_type.DeploymentType
 import magenta.input.resolver.Resolver
 import magenta.tasks.STS
 import magenta.{DeployParameters, DeployReporter, Lookup, Region, StsDeploymentResources, App => MagentaApp, Stack => MagentaStack}
+import notification.FailureNotificationContents._
 import schedule.ScheduledDeployer
 import rx.lang.scala.Subscription
 
@@ -31,14 +31,6 @@ class DeployFailureNotifications(config: Config,
   lazy private val snsClient = config.scheduledDeployment.snsClient
   lazy private val prefix = config.urls.publicPrefix
   lazy private val riffRaffTargets = List(App("riff-raff"), Stack("deploy"))
-
-  case class NotificationContents(subject: String, message: String, actions: List[Action])
-  case class NotificationContentsWithTargets(notificationContents: NotificationContents, targets: List[Target])
-
-  def problematicDeployUrl(uuid: UUID): String = {
-    val path = routes.DeployController.viewUUID(uuid.toString, verbose = true)
-    prefix + path.url
-  }
 
   def getAwsAccountIdTarget(target: ci.Target, parameters: DeployParameters, uuid: UUID): Option[Target] = {
     Try {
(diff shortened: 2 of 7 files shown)


riff-raff/app/notification/DeployFailureNotifications.scala
```diff
@@ -50,46 +55,90 @@ class DeployFailureNotifications(config: Config,
...
+      } yield {
+        TargetResolver.extractTargets(deployGraph).toList.flatMap { target =>
+          val maybeAccountId = uuid.flatMap(uuid => getAwsAccountIdTarget(target, parameters, uuid)).toList
+          List(App(target.app), Stack(target.stack), Stage(parameters.stage.name)) ++ maybeAccountId
+        }
       }
     }
+    attemptToDeriveTargets match {
+      case Some(Right(targets)) => targets
+      case _ => riffRaffTargets
```
jacobwinch, line 72: This is new behaviour, previously if we couldn't find the targets we just dropped the notification entirely: https://github.com/guardian/riff-raff/pull/614/files#diff-0dceec64b87cbc45ffb3aabb2b6f1fb0abd9b04fcc1e1cde546cd01b49275362L80-L81

riff-raff/app/notification/DeployFailureNotifications.scala
```diff
@@ -50,46 +55,90 @@ class DeployFailureNotifications(config: Config,
...
+    def viewProblematicDeploy(uuid: UUID, status: String) = Action(s"View $status deploy", problematicDeployUrl(uuid))
+    scheduledDeployError match {
+      case SkippedDueToPreviousFailure(record) =>
+        val message = s"A scheduled deploy of ${record.parameters.build.projectName} to ${record.parameters.stage} didn't start because the most recent deploy failed."
+        val redeployAction = Action("Redeploy manually", prefix + routes.DeployController.deployAgainUuid(record.uuid.toString).url)
+        val contents = NotificationContents(subject, message, List(viewProblematicDeploy(record.uuid, "failed"), redeployAction))
+        NotificationContentsWithTargets(contents, getTargets(None, Some(record.parameters)))
+      case SkippedDueToPreviousWaitingDeploy(record) =>
+        val message = s"A scheduled deploy of ${record.parameters.build.projectName} to ${record.parameters.stage} failed to start as a previous deploy was still waiting to be deployed."
+        val contents = NotificationContents(subject, message, List(viewProblematicDeploy(record.uuid, "waiting")))
```
jacobwinch, line 101: Here's an [example](https://riffraff.gutools.co.uk/deployment/view/699d166c-c0e5-4e12-90f0-9f37f89d2190) of this problem. @sihil - any ideas how we should resolve problems like this?

stephengeller:
This looks great! My only thinking is whether we can have tests added for some of the new logic in `DeployFailureNotifications.scala`? Not sure how easy it is to add, but might be nice to have some tests ensure that our logic and behaviour is _behaving_ as intended, and might make it easier to digest too.

jacobwinch:
> My only thinking is whether we can have tests added for some of the new logic in `DeployFailureNotifications.scala`?

Thanks for the suggestion @stephengeller, I've added some tests here https://github.com/guardian/riff-raff/pull/614/commits/68f26e5cc0752286db0e1b4c15d93f0cf6ffd4cd. It made me notice a few typos/inconsistencies in the notification copy (which I've fixed in the same commit), so I think it was worth adding these.

> Not sure how easy it is to add...

It wasn't too difficult. We basically want to test the parts of the code which have 0 side effects, so I refactored slightly to pull them into a separate object: https://github.com/guardian/riff-raff/pull/614/commits/aef690c450e68d9c6177c1e7633b20bd7b0f3627. This means that we don't need to mock all of the class's dependencies, which are irrelevant to the logic under test.

riff-raff/app/notification/FailureNotificationContents.scala
```diff
@@ -0,0 +1,61 @@
...
+
+  def deployFailedNotificationContents(uuid: UUID, parameters: DeployParameters, urlPrefix: String): NotificationContents = {
+    NotificationContents(
+      subject = s"${parameters.deployer.name} failed",
+      message = s"${parameters.deployer.name} for ${parameters.build.projectName} (build ${parameters.build.id}) to stage ${parameters.stage.name} failed.",
+      actions = List(viewProblematicDeploy(uuid, "failed", urlPrefix))
+    )
+  }
+
+  def deployUnstartedNotificationContents(scheduledDeployError: ScheduledDeployError, urlPrefix: String, teamTargetsSearch: (Option[UUID], Option[DeployParameters]) => List[Target], riffRaffTargets: List[Target]): NotificationContentsWithTargets = {
```
jacobwinch, line 40: We pass in a function for `teamTargetsSearch`, which makes it easy to use a mock function in unit tests:

`def fakeGetTargets(uuid: Option[UUID], maybeParameters: Option[DeployParameters]): List[Target] = List(Stack("another-team-stack"))`

This is necessary as the real `getTargets` function talks to s3 etc.

riff-raff/app/deployment/model.scala
```diff
@@ -17,7 +16,12 @@ case object ContinuousDeploymentRequestSource extends RequestSource
 case class ApiRequestSource(key: ApiKey) extends RequestSource
 case object ScheduleRequestSource extends RequestSource
 
-case class Error(message: String) extends AnyVal
+sealed trait ScheduledDeployError
+case class NoDeploysFoundForStage(projectName: String, stage: String) extends ScheduledDeployError
+case class SkippedDueToPreviousFailure(failedDeployRecord: Record) extends ScheduledDeployError
+case class SkippedDueToPreviousWaitingDeploy(waitingDeployRecord: Record) extends ScheduledDeployError
+
+case class Error(message: String, scheduledDeployError: Option[ScheduledDeployError] = None)
```
sihil, line 24: I would rearrange this into an error hierarchy such that Error is a trait with a message and all the other errors extend from it setting the message appropriately. This will make the Error hierarchy more generic so it can be adopted elsewhere as required.

riff-raff/app/notification/DeployFailureNotifications.scala
```diff
@@ -50,46 +47,61 @@ class DeployFailureNotifications(config: Config,
...
+    error.scheduledDeployError.map { failedToStartReason =>
+      val contentsWithTargets = deployUnstartedNotificationContents(failedToStartReason, prefix, getTargets, riffRaffTargets)
+      notifyViaAnghammarad(contentsWithTargets.notificationContents, contentsWithTargets.targets)
+    }.getOrElse {
+      log.warn("Scheduled deploy failed to start but notification was not sent...")
     }
   }
 
+  def deployFailedNotification(uuid: UUID, parameters: DeployParameters, targets: List[Target]): Unit = {
+    val notificationContents = deployFailedNotificationContents(uuid, parameters, prefix)
```
sihil, line 92: Personally I prefer refering explicitly to methods in an external object (i.e. qualified with the object name) rather than importing the methods. Makes it easier to comprehend.

riff-raff/app/notification/DeployFailureNotifications.scala
```diff
@@ -50,46 +47,61 @@ class DeployFailureNotifications(config: Config,
...
-
-    val deriveAnghammaradTargets = for {
-      yaml <- targetResolver.fetchYaml(parameters.build)
-      deployGraph <- Resolver.resolveDeploymentGraph(yaml, deploymentTypes, magenta.input.All).toEither
-    } yield {
-      TargetResolver.extractTargets(deployGraph).toList.flatMap { target =>
-        val maybeAccountId = uuid.flatMap(uuid => getAwsAccountIdTarget(target, parameters, uuid)).toList
-        List(App(target.app), Stack(target.stack), Stage(parameters.stage.name)) ++ maybeAccountId
+  def getTargets(uuid: Option[UUID], maybeParameters: Option[DeployParameters]): List[Target] = {
+    val attemptToDeriveTargets = maybeParameters.map { parameters =>
```
sihil, line 51: If you coerce maybeParameters into an `Either` with `toRight` then you can make this a part of the for comprehension.

This would require creating a common `Error` hierarchy so that the left hand side of the eithers match up - but this would be a good thing to do.

riff-raff/app/notification/FailureNotificationContents.scala
```diff
@@ -0,0 +1,61 @@
+package notification
+
+import com.gu.anghammarad.models.{Action, Target}
+import controllers.routes
+import deployment.{NoDeploysFoundForStage, ScheduledDeployError, SkippedDueToPreviousFailure, SkippedDueToPreviousWaitingDeploy}
+import magenta.DeployParameters
+
+import java.util.UUID
+
+object FailureNotificationContents {
```
sihil, line 10: This might work better as a class that takes `prefix`, `teamTargetsSearch` and `riffRaffTargets` as parameters? This would clean up the signatures but might make it worse in other ways I haven't thought of...  just a thought.

riff-raff/app/notification/FailureNotificationContents.scala
```diff
@@ -0,0 +1,61 @@
...
+      message = s"${parameters.deployer.name} for ${parameters.build.projectName} (build ${parameters.build.id}) to stage ${parameters.stage.name} failed.",
+      actions = List(viewProblematicDeploy(uuid, "failed", urlPrefix))
+    )
+  }
+
+  def deployUnstartedNotificationContents(scheduledDeployError: ScheduledDeployError, urlPrefix: String, teamTargetsSearch: (Option[UUID], Option[DeployParameters]) => List[Target], riffRaffTargets: List[Target]): NotificationContentsWithTargets = {
+    val subject = "Scheduled Deployment failed to start"
+    scheduledDeployError match {
+      case SkippedDueToPreviousFailure(record) =>
+        val message = s"Scheduled Deployment for ${record.parameters.build.projectName} to ${record.parameters.stage.name} didn't start because the most recent deploy failed."
```
sihil, line 44: If you create a common error hierarchy with a `message` field then you can build these messages in the errors themselves so they carry the context.

riff-raff/app/schedule/DeployJob.scala
```diff
@@ -22,32 +22,28 @@ class DeployJob extends Job with Logging {
...
+      record <- DeployJob.getLastDeploy(deployments, projectName, stage)
+      params <- DeployJob.createDeployParameters(record, scheduledDeploymentEnabled)
+      uuid <- deployments.deploy(params, ScheduleRequestSource)
+    } yield uuid
+
+    attemptToStartDeploy match {
+      case Left(error) =>
+        scheduledDeployNotifier.deployUnstartedNotification(error)
+      case Right(uuid) =>
+        log.info(s"Started scheduled deploy $uuid")
```
sihil, line 38: This is all much nicer 💯

riff-raff/app/schedule/DeployJob.scala
```diff
@@ -56,8 +52,12 @@ object DeployJob extends Logging with LogAndSquashBehaviour {
         } else {
           Left(Error(s"Scheduled deployments disabled. Would have deployed $params"))
         }
+      case RunState.Failed =>
+        Left(defaultError(RunState.Failed).copy(scheduledDeployError = Some(SkippedDueToPreviousFailure(lastDeploy))))
+      case RunState.NotRunning =>
+        Left(defaultError(RunState.Failed).copy(scheduledDeployError = Some(SkippedDueToPreviousWaitingDeploy(lastDeploy))))
```
sihil, line 58: This looks like you should be passing in `RunState.NotRunning` here? The `defaultError` method actually has access to `lastDeploy.state` so might be easier to get it directly as per `uuid`.

riff-raff/app/schedule/DeployJob.scala
```diff
@@ -56,8 +52,12 @@ object DeployJob extends Logging with LogAndSquashBehaviour {
         } else {
           Left(Error(s"Scheduled deployments disabled. Would have deployed $params"))
         }
+      case RunState.Failed =>
+        Left(defaultError(RunState.Failed).copy(scheduledDeployError = Some(SkippedDueToPreviousFailure(lastDeploy))))
```
sihil, line 56: You could avoid the copy by adding another parameter with a default value to `defaultError` - perhaps renamed to `stateError` or some such.
This might look different with a common error hierarchy.

sihil (commented):
The shape of this is great. I'd encourage the adoption of a common error hierarchy in the codebase so that `Either`s can be composed more easily. I'd also move the responsbility for the error messages into the hierarchy.

riff-raff/app/deployment/model.scala
jacobwinch: > I would rearrange this into an error hierarchy such that Error is a trait with a message and all the other errors extend from it setting the message appropriately.

Thanks for the suggestion, I've tried to do that here: https://github.com/guardian/riff-raff/pull/614/commits/4c70adbf78fee91f6f88e7bea4cc120b6c61d932. 

The `Error(message: String)` case class is already used widely throughout the codebase, so I've given the trait a different name to avoid conflicts!

riff-raff/app/notification/DeployFailureNotifications.scala
jacobwinch: I've realised that we had unnecessary optionality here, so I've managed to [simplify the function](https://github.com/guardian/riff-raff/pull/614/commits/0095af06be3764f08cb1b8ad315b52c6ef32698b) and avoid this problem entirely!

riff-raff/app/schedule/DeployJob.scala
jacobwinch: Thanks for spotting this - you caught me copy and pasting 🙈 

Fortunately, this problem goes away anyway now that we've moved to an error hierarchy.

riff-raff/app/notification/FailureNotificationContents.scala
jacobwinch: I've converted to a class as it helps us to avoid passing `prefix` around so much: https://github.com/guardian/riff-raff/pull/614/commits/02f5c5f195018e2363bf6216b1cb27a9be461b1f. As `teamTargetsSearch` and `riffRaffTargets` are only used by one function I've left those as function parameters.

riff-raff/app/notification/DeployFailureNotifications.scala
```diff
@@ -50,36 +47,44 @@ class DeployFailureNotifications(config: Config,
...
+      sourceSystem = "riff-raff",
+      channel = All,
+      target = targets,
+      actions = notificationContents.actions,
+      topicArn = anghammaradTopicARN,
+      client = snsClient
+    ).recover { case ex => log.error(s"Failed to send notification (via Anghammarad)", ex) }
+  }
+
+  def deployUnstartedNotification(error: ScheduledDeployNotificationError): Unit = {
```
akash1810, line 80: Would `scheduledDeployFailureNotification` be a more explicit name here?

riff-raff/app/notification/DeployFailureNotifications.scala
```diff
@@ -50,36 +47,44 @@ class DeployFailureNotifications(config: Config,
...
+    val contentsWithTargets = failureNotificationContents.deployUnstartedNotificationContents(error, getTargets, riffRaffTargets)
+    notifyViaAnghammarad(contentsWithTargets.notificationContents, contentsWithTargets.targets)
+  }
+
+  def deployFailedNotification(uuid: UUID, parameters: DeployParameters, targets: List[Target]): Unit = {
```
akash1810, line 85: Would `midDeployFailureNotification` be a more explicit name here?

akash1810 (approved):
Discussed this PR IRL with @jacobwinch . All looks sane, so 👍🏽 .

A couple of naming suggestions, but nothing that's blocking.

riff-raff/app/notification/DeployFailureNotifications.scala
jacobwinch: I think the names you've suggested are an improvement. I've updated them here: https://github.com/guardian/riff-raff/pull/614/commits/5dd6bea2d47343ba8f500e70482f3c823f621cf0.

PR #616, opened 2021-02-23
Update deployment types documentation

jacobwinch:
## What does this change?

This PR removes some unnecessary information from [Riff-Raff's documentation page](https://riffraff.gutools.co.uk/docs/magenta-lib/types).

Historically, teams managed their own Riff-Raff users. Consequently, they needed to ensure that these users had the correct IAM permissions to support various Riff-Raff deployment types. 

Riff-Raff now assumes [a role](https://github.com/guardian/deploy-tools-platform/blob/main/cloudformation/riffraff/riffraff-roles.yaml) which is maintained by DevX. Consequently, other teams / users don't need to think about the IAM permissions which support various deployment types. Similarly, they no longer need to take any steps to add permissions to IAM users.

For more details on the new approach, see [new docs](https://github.com/guardian/deploy-tools-platform/blob/main/cloudformation/riffraff/riff-raff-roles.md).

## How to test
N/A

The docs are slightly easier to follow / no longer contain outdated instructions.

N/A


jorgeazevedo (approved):
Good to clean this stuff up!

PR #618, opened 2021-03-08
Permit scheduled deploys when restrictions are enabled if CD is enabled

mbarton:
## What does this change?

Giant has restrictions on who can deploy it which form part of our defence in depth strategy as it can contain sensitive data. There is an issue currently where it stalls processing work after a few days and so we set up a daily scheduled deploy to keep it available for users whilst we investigate the underlying issue.

Unfortunately the scheduled deploy didn't work:

```
Scheduled deploy failed to start due to Error(Unable to queue deploy as restrictions are currently in place: Michael Barton: Sensitive data). A notification will not be sent...
```

We had configured the restriction to allow continuous deployment but scheduled deploys were still not allowed. This PR changes the behaviour to allow both if the checkbox is set and updates the description accordingly.

## How to test

It's fully unit tested but to reproduce manually:

- Set up a restriction on a deploy
- Set up a scheduled deploy

Currently the scheduled deploy is suppressed with the error above. After this PR the deploy proceeds.

The scheduled deploy works

I think this is a clear, well tested change that follows what users would expect the behaviour to be. I haven't renamed the config variable though as I think it's serialised.

### DIFF
--- a/riff-raff/app/restrictions/RestrictionChecker.scala
+++ b/riff-raff/app/restrictions/RestrictionChecker.scala
@@ -1,7 +1,7 @@
 package restrictions
 
 import com.gu.googleauth.UserIdentity
-import deployment.{ContinuousDeploymentRequestSource, Error, RequestSource, UserRequestSource}
+import deployment.{ContinuousDeploymentRequestSource, Error, RequestSource, ScheduleRequestSource, UserRequestSource}
 
 object RestrictionChecker {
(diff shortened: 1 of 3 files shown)


aug24 (approved):

philmcmahon (approved):
Sounds good to me

joelochlann (approved):

PR #619, opened 2021-03-24
s3 - surrogate cache parameter

tomrf1:
We currently have the ability to set the `cache-control` header on the `aws-s3` task.
I'd also like to be able to set the [surrogate-control](https://docs.fastly.com/en/guides/configuring-caching#surrogate-control) header as well.

Screen Shot 2021-03-24 at 09 10 21

### DIFF
--- a/magenta-lib/src/main/scala/magenta/deployment_type/S3.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/S3.scala
@@ -76,6 +76,12 @@ object S3 extends DeploymentType {
     """.stripMargin
   )
 
+  val surrogateCacheControl = Param[List[PatternValue]]("surrogateCacheControl",
+    """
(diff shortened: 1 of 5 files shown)


sihil:
Looks sane - do you want to add a noddy test to ensure that it is passed through?

tomrf1:
> Looks sane - do you want to add a noddy test to ensure that it is passed through?

I've added a very basic check [here](https://github.com/guardian/riff-raff/pull/619/files#diff-bb7f963ee457aacae0f840c075006a82dde729a229d0d6c4c01bba257645f59fR87)
I'm not sure how else this logic is tested

sihil (approved):
LGTM

sihil:
I missed that - there is also scope to check it gets passed through to the generated requests, but I think this is sufficient.

tomrf1:
Tested in CODE (with support-dotcom-components):
Screen Shot 2021-03-24 at 10 32 29

PR #620, opened 2021-03-24
Highlight all table rows

mxdvl:
## What does this change?

This highlights all row hovered in tables, not just clickable ones.

## How to test

Hover over tables outside of the deployment history and get the same highlighting, minus the `pointer`.

More accessibility and confidence in making changes on long lists.

## Images

example

### DIFF
--- a/riff-raff/app/assets/stylesheets/magenta.less
+++ b/riff-raff/app/assets/stylesheets/magenta.less
@@ -20,13 +20,6 @@ tr.rowlink td {
 tr.rowlink td.nolink {
   cursor: auto;
 }
-.table tbody tr.rowlink:hover td {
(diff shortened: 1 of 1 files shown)


akash1810 (approved):

mxdvl:
I’ve been unable to run this locally to test, as I’m not sure what to write in `~/.gu/riff-raff.conf`.

akash1810:
> I’ve been unable to run this locally to test, as I’m not sure what to write in `~/.gu/riff-raff.conf`.

Happy to pair with you to get it running locally. Alternatively we can deploy to CODE. Alternatively, alternatively, we could merge and hope as it looks like a very innocent change and users are internal.

PR #622, opened 2021-03-31
feat: surprise 🎉

akash1810:
# 🃏   A most excellent April Fools 🃏 

On `DOMContentLoaded` rotate the `body` element by a random multiple of 10 between 1 and 360 and prevent elements from being clicked to avoid any accidents!

Also use `localStorage` to track if the April Fools has been seen; if it has, it won't be run again. That is, refreshing the page will return things back to normal.

image

img


philmcmahon (approved):
Excellent performance improvement

PR #623, opened 2021-05-07
Apply stack policies before and after CFN updates

sihil:
## What does this change?
We'd like to better protect users from accidental resource deletion or changes that will replace a resource. For a sub set of resource types this can be destructive so we no longer allow such updates when Riff-Raff is configured to manage the stack policy. 

Riff-Raff can be configured to manage stack policies either by directly setting the `manageStackPolicies` setting in a `riff-raff.yaml` config file or targetted against multiple projects by configuring the `cloudformation:manage-stack-policy` key in the lookup service (Prism data).

## How to test
This will need a permission change (https://github.com/guardian/deploy-tools-platform/pull/471) and setting `manageStackPolicies` to true in another project to try it out (https://github.com/guardian/prism/compare/sihil/riff-raff-stack-policy-management).

If a CFN update is attempted that would replace a load balancer or another resource type on the list then this will fail.

This change sets the stack policy back to "allow all" after making changes. If a team is actually using stack policies then this will overwrite them. For now this is behind a flag whilst we understand this risk better.

### DIFF
--- a/magenta-lib/src/main/scala/magenta/deployment_type/CloudFormation.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/CloudFormation.scala
@@ -65,6 +65,8 @@ object CloudFormation extends DeploymentType with CloudFormationDeploymentTypePa
   val secondsToWaitForChangeSetCreation = Param("secondsToWaitForChangeSetCreation",
     "Number of seconds to wait for the change set to be created").default(15 * 60)
 
+  val manageStackPolicyDefault = false
+  val manageStackPolicyLookupKey = "cloudformation:manage-stack-policy"
   val manageStackPolicyParam = Param[Boolean]("manageStackPolicy",
     s"""Allow RiffRaff to manage stack update policies on your behalf.
       |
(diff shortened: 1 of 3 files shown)


akash1810 (approved):

magenta-lib/src/main/scala/magenta/deployment_type/CloudFormation.scala
```diff
@@ -64,6 +65,29 @@ object CloudFormation extends DeploymentType with CloudFormationDeploymentTypePa
...
+      |policy will be updated again to allow all operations to be carried out (the same as having no policy, but
+      |it is not possible to delete a stack policy).
+      |
+      |This can also be set via the `$manageStackPolicyLookupKey` lookup key which can be used to control this setting
+      |across a larger number of projects. This setting overrides this and if neither exist the default is
+      |$manageStackPolicyDefault.
+      |
+      |The two stack policies are show below.
+      |
+      |${StackPolicy.toMarkdown(StackPolicy.DENY_REPLACE_DELETE_POLICY)}
```
akash1810, line 84: Love this!

akash1810 (approved):

PR #624, opened 2021-05-11
fix: Only set a stack policy when a stack exists

akash1810:
## What does this change?

It is not possible to update a stack policy for a stack that doesn't yet exist, so don't.

### Before
image

#### Initial deploy to create the stack
image

#### Additional deploy to demonstrate stack policy being set
image

## How to test

- Deploy `main` to CODE
- Deploy a project that will create a stack
- Witness the deploy fail
- Deploy to CODE
- Deploy a project that will create a stack
- Witness the stack being created and the deploy succeeding
- Deploy the project again and witness the deploy log setting the stack policy

We're able to continue using RiffRaff to create stacks if they do not exist.

n/a


sihil (approved):
💯

PR #625, opened 2021-05-14
Stream the S3 data rather than buffer entirely in memory

sihil:
## What does this change?
When migrating from AWS SDK v1 to v2 the S3 Upload was changed to buffer the entire file being uploaded from bucket A to B in memory. This created extra memory pressure (and has caused some deploys to fail when they have particularly large artifacts being copied when there is contention with other deploys) so this change switches back to streaming with a 1Mb buffer.

Due to the way that this is now implemented, the get and put requests must be on different threads. In order to facilitate this we add an isolated execution context for IO which is now passed around so tasks can use that when appropriate.

## How to test
Pretty much any deploy will hit this path and so issues will likely become apparent fairly quickly. This has been tested in CODE to deploy itself. On a very small sample it seems to be around 10-15% slower but this might not be the case with further testing.

Riff-Raff memory pressure is significantly decreased when lots of deploys are running simultaneously.

This is a significant change to a frequent path of code and it's possible that there are performance or memory concerns that we've not flushed out. It'll be hard to do this without deploying it.

### DIFF
--- a/magenta-lib/src/main/scala/magenta/Project.scala
+++ b/magenta-lib/src/main/scala/magenta/Project.scala
@@ -9,6 +9,7 @@ import magenta.input._
 import magenta.tasks.Task
 import software.amazon.awssdk.services.sts.StsClient
 
+import scala.concurrent.ExecutionContext
 import scala.collection.immutable
 import scala.math.Ordering.OptionOrdering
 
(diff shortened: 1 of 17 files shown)


jonathonherbert (approved):
LGTM. Good job adding the mocks for that test, looks sensible. 🤞 it performs as expected.

PR #626, opened 2021-05-20
Add an update strategy to the deploy form

sihil:
## What does this change?
PR #623 introduced a more cautious approach for applying cloudformation updates. We want to roll this out to everyone but it would be nice to provide a way to execute a destructive change if they really wanted to without having to go around RiffRaff.

This PR introduces the idea of update strategies. We currently support two strategies: `MostlyHarmless` and `Dangerous`. At the moment this will only be used to decide the policy to apply for a cloudformation update when stack policy management is enabled (which should be the default soon).

When set to `MostlyHarmless` we will use a policy that denies updates and replacements. When set to `Dangerous` we will use an allow all policy.

This appears in the deploy dialog as below with a link to docs that describe this in more detail. It's hidden behind an advance settings reveal so that they are not on display until actually required.

image

We can roll out policy management from #623 without concern that users will be unable to deploy certain artifacts if they have an intentionally destructive change.

Users think that we are safer than we really are. In order to mitigate this we provide a page of documentation that is easily discoverable from the deploy screen.

### DIFF
--- a/riff-raff/app/assets/javascripts/collapsing.coffee
+++ b/riff-raff/app/assets/javascripts/collapsing.coffee
@@ -13,6 +13,7 @@ setupCallbacks = ->
 
 
 $ ->
+  setupCallbacks()
   if (window.autoRefresh)
     window.autoRefresh.postRefresh ->
       setupCallbacks()
--- a/riff-raff/app/controllers/forms/DeployParameterForm.scala
+++ b/riff-raff/app/controllers/forms/DeployParameterForm.scala
@@ -1,12 +1,14 @@
 package controllers.forms
 
+import magenta.Strategy
 import magenta.input.{All, DeploymentKey, DeploymentKeysSelector, DeploymentSelector}
 import play.api.data.Form
 import play.api.data.Forms._
+import play.api.data.format.Formatter
 import utils.Forms._
 
 case class DeployParameterForm(project:String, build:String, stage:String, action: String,
-  selectedKeys: List[DeploymentKey], totalKeyCount: Option[Int]) {
+  selectedKeys: List[DeploymentKey], totalKeyCount: Option[Int], updateStrategy: Strategy) {
 
   def makeSelector: DeploymentSelector = {
     val keysList =
@@ -23,6 +25,8 @@ case class DeployParameterForm(project:String, build:String, stage:String, actio
 }
 
 object DeployParameterForm {
+  val strategyFormat: Formatter[Strategy] = enumeratum.Forms.format(Strategy)
+
   val form = Form[DeployParameterForm](
     mapping(
       "project" -> nonEmptyText,
@@ -30,7 +34,8 @@ object DeployParameterForm {
       "stage" -> text,
       "action" -> nonEmptyText,
       "selectedKeys" -> list(deploymentKey),
-      "totalKeyCount" -> optional(number)
+      "totalKeyCount" -> optional(number),
+      "updateStrategy" -> of[Strategy](strategyFormat)
     )(DeployParameterForm.apply)(DeployParameterForm.unapply)
   )
 }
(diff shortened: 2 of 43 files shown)


riff-raff/app/controllers/forms/DeployParameterForm.scala
```diff
@@ -23,14 +25,17 @@ case class DeployParameterForm(project:String, build:String, stage:String, actio
...
   val form = Form[DeployParameterForm](
     mapping(
       "project" -> nonEmptyText,
       "build" -> nonEmptyText,
       "stage" -> text,
       "action" -> nonEmptyText,
       "selectedKeys" -> list(deploymentKey),
-      "totalKeyCount" -> optional(number)
+      "totalKeyCount" -> optional(number),
+      "updateStrategy" -> of[Strategy](strategyFormat)
```
akash1810, line 38: Ooh, this syntax is new to me! Is this the enumeratum library?

riff-raff/app/views/preview/yaml/showTasks.scala.html
```diff
@@ -13,18 +14,20 @@
...
           case None => {
               <div class="alert alert-info">
                   This preview shows only the subset of deployment tasks that would be executed. To see all tasks, <a href="@routes.PreviewController.preview(
-                      form("project").value.get,
-                      form("build").value.get,
-                      form("stage").value.get,
-                      None
+                      form.value.map(_.project).get,
+                      form.value.map(_.build).get,
+                      form.value.map(_.stage).get,
```
akash1810, line 28: Should we use this syntax across the entire file?

riff-raff/public/docs/riffraff/advanced-settings.md
```diff
@@ -0,0 +1,14 @@
...
+
+**Note: At the time of writing this is only observed for some deploys under some circumstances.**
+
+Specifically, this currently means that when your deploy applies updates to a cloudformation stack, Riff-Raff will apply a stack update policy before executing the update. When you are using the default _MostlyHarmless_ update strategy this policy will prohibit the deletion and replacement of certain resource types. When you are using the _Dangerous_ update strategy this policy will allow any change to take place.
+
+In practice, we expect most deploys to be run using the _MostlyHarmless_ strategy. In the case that this fails with a stack policy error you can review the deployment and then use the _Dangerous_ strategy. It is recommended that you manually check (by looking at the change set) that you really do want the resources in question to be deleted before using the _Dangerous_ strategy to execute destructive actions.
```
akash1810, line 14: It is worth mentioning continuous and scheduled are always `MostlyHarmless`?

riff-raff/app/assets/javascripts/collapsing.coffee
```diff
@@ -13,6 +13,7 @@ setupCallbacks = ->
 
 
 $ ->
+  setupCallbacks()
```
akash1810, line 16: I don't quite understand why this is needed (I also don't think I want to... ☕ 😱 )

akash1810 (approved):
A couple of minor comments, but this looks great overall!

riff-raff/app/controllers/forms/DeployParameterForm.scala
sihil: Not enumeratum perse, but enumeratum is generating a binder for the Form. Much like path bindables, but for forms :)

riff-raff/app/views/preview/yaml/showTasks.scala.html
sihil: I thought I had - have just done the hiddens at the top if that was the one you were referencing?

riff-raff/app/assets/javascripts/collapsing.coffee
sihil: The section below was rerunning it after refreshes, but I am now using it on a page that doesn't have auto-refresh and so the callbacks were never being instantiated. Rewriting all this in react would be fun hey?

PR #628, opened 2021-06-09
Remove ElasticSearch deployment type

jacobwinch:
## What does this change?

This PR removes the ElasticSearch deployment type (which was originally added back in https://github.com/guardian/riff-raff/pull/61). Based on a [GitHub search](https://github.com/search?q=org%3Aguardian+extension%3Ayaml+filename%3Ariff-raff+%22type%3A+elasticsearch%22&type=Code), I believe that this deployment type is no longer used, but I'm struggling to find any old PRs/docs which explain why we moved away from this solution. 

Many teams in the department are now using [elasticsearch-node-rotation](https://github.com/guardian/elasticsearch-node-rotation) in order to keep nodes in their Elasticsearch clusters up to date and I don't think it makes sense to maintain two ways of achieving this goal, particularly if nobody is currently using this older solution.

## How to test

I've confirmed that Riff-Raff can still be deployed to `CODE` and that a standard `autoscaling` deployment type still works as expected.

After this code has been removed:

* There will be a single, recommended way of rotating Elasticsearch nodes
* It will be easier and faster to develop against this codebase (i.e. I won't need to touch this code in the refactor that I was just trying to work on!)

Yes, there is a risk that a project somewhere is still making use of this deployment type. If someone tries to deploy with an `elasticsearch` deployment type after this change has been merged, I'd expect to see [this error](https://github.com/guardian/riff-raff/blob/01f0938b09b15208244f473f88032666c4882577/magenta-lib/src/main/scala/magenta/input/resolver/DeploymentTypeResolver.scala#L12-L13).

If someone with sufficient permissions to see any repo in our org would be able to check [these search results](https://github.com/search?q=org%3Aguardian+extension%3Ayaml+filename%3Ariff-raff+%22type%3A+elasticsearch%22&type=Code), it would be much appreciated! If you replace the word `elasticsearch`, with another valid deployment type (e.g. [`fastly`](https://github.com/search?q=org%3Aguardian+extension%3Ayaml+filename%3Ariff-raff+%22type%3A+fastly%22&type=Code)), this should also prove that the search itself is correct!


jacobwinch:
cc @rtyley @davidfurey @philmcmahon 

Are any of you able to help me with this?

> I'm struggling to find any old PRs/docs which explain why we moved away from this solution.

Or this?

> If someone with sufficient permissions to see any repo in our org would be able to check [these search results](https://github.com/search?q=org%3Aguardian+extension%3Ayaml+filename%3Ariff-raff+%22type%3A+elasticsearch%22&type=Code), it would be much appreciated! If you replace the word `elasticsearch`, with another valid deployment type (e.g. [`fastly`](https://github.com/search?q=org%3Aguardian+extension%3Ayaml+filename%3Ariff-raff+%22type%3A+fastly%22&type=Code)), this should also prove that the search itself is correct!

davidfurey:
I think this deployment serves a purpose beyond the node rotation.  It should probably still be used by, at least, CAPI and Ophan both of which, I think, accidentally stopped using it:
* https://github.com/guardian/content-api/tree/364094a2d5dd14dbc1db0bf166d8d0970bd4638d/elasticsearch
* https://github.com/guardian/ophan/pull/3951/files

Can we discuss this on a call?

davidfurey:
Also relevant: https://github.com/guardian/content-api/pull/191/commits/e30352be71a3126428e3d532d715522da4efcb97

PR #629, opened 2021-06-11
Use mockito-scala and latest version of scalatest

jacobwinch:
## What does this change?

* Removes [`mockito-core`](https://github.com/mockito/mockito) and adds [`mockito-scala`](https://github.com/mockito/mockito-scala), as this has better support for Scala projects (I need to make use of [this feature](https://github.com/mockito/mockito-scala#mocking-scala-object) in a subsequent PR, which is the primary motivation for this change).
* Updates `scalatest` for compatibility and to keep things up to date.

Although this PR touches a lot of files as a result of modularisation changes in an [intermediate version of `scalatest`
](https://www.scalatest.org/release_notes/3.1.0) the only thing that's really changing here is import statements. I've raised this in its own PR to avoid adding noise to the refactor that I'm really trying to work on!

## How to test

* All tests should still pass in CI

* Our dependencies are up to date

* I don't consider this to be a risky change

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -16,8 +16,8 @@ object Dependencies {
     "org.parboiled" %% "parboiled" % "2.1.5",
     "org.typelevel" %% "cats-core" % "1.0.1",
     "com.kailuowang" %% "henkan-convert" % "0.2.10",
-    "org.scalatest" %% "scalatest" % "3.0.3" % Test,
-    "org.mockito" % "mockito-core" % "2.27.0" % Test
+    "org.scalatest" %% "scalatest" % "3.2.9" % Test,
+    "org.mockito" %% "mockito-scala" % "1.16.37" % Test
   )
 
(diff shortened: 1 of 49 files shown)


project/Dependencies.scala
```diff
@@ -16,8 +16,8 @@ object Dependencies {
     "org.parboiled" %% "parboiled" % "2.1.5",
     "org.typelevel" %% "cats-core" % "1.0.1",
     "com.kailuowang" %% "henkan-convert" % "0.2.10",
-    "org.scalatest" %% "scalatest" % "3.0.3" % Test,
-    "org.mockito" % "mockito-core" % "2.27.0" % Test
+    "org.scalatest" %% "scalatest" % "3.2.9" % Test,
```
jacobwinch, line 19: Here's the change to the dependencies

akash1810 (approved):

PR #631, opened 2021-06-16
Refactor ASG search

jacobwinch:
## What does this change?

This PR refactors the way that Riff-Raff identifies which ASG to interact with as part of an `autoscaling` deployment.

Previously, Riff-Raff would search for the relevant ASG (by tags) as part of every individual task (e.g. `WaitForStabilization`) in a deployment action (e.g. `deploy`). Accounts can have many ASGs and the `deploy` action is comprised of many tasks, so this is inefficient. It's also unnecessary, as once we have selected the ASG to deploy to, we should perform all actions against the same ASG.

This PR refactors the code to identify the relevant ASG (still using tags) before starting _any_ tasks. Each individual task can then simply ask for the latest state of the specific ASG it cares about, rather than searching through all ASGs in the account. This allows us to simplify the logic(?) and fail faster if an appropriate ASG cannot be found. It should also make it easier to run multiple `deploy` actions as part of the `autoscaling` deployment type, which we will need in order to support the `cdk` migration.

## How to test

I've deployed this to `CODE` and confirmed that a couple of projects (AMIgo and AMIable) can still be deployed as expected.

* All `autoscaling` deployments should continue working
* It should be easier to add support for deploying to multiple ASGs as part of the same `autoscaling` deploy

Yes, this deployment type is used by many projects in the department, including our highest-profile projects. After merging this we should keep a close eye on the deployment failure rate in case we have introduced any unexpected problems.

### DIFF
--- a/magenta-lib/src/main/scala/magenta/deployment_type/ElasticSearch.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/ElasticSearch.scala
@@ -37,17 +37,16 @@ object ElasticSearch extends DeploymentType {
   ) { (pkg, resources, target) =>
     implicit val keyRing = resources.assembleKeyring(target, pkg)
     val reporter = resources.reporter
-    val parameters = target.parameters
-    val stack = target.stack
+    val asgName: String = AutoScalingGroupLookup.getTargetAsgName(keyRing, target, resources, pkg)
     List(
-      CheckGroupSize(pkg, parameters.stage, stack, target.region),
-      WaitForElasticSearchClusterGreen(pkg, parameters.stage, stack, secondsToWait(pkg, target, reporter) * 1000, target.region),
-      SuspendAlarmNotifications(pkg, parameters.stage, stack, target.region),
-      TagCurrentInstancesWithTerminationTag(pkg, parameters.stage, stack, target.region),
-      DoubleSize(pkg, parameters.stage, stack, target.region),
-      WaitForElasticSearchClusterGreen(pkg, parameters.stage, stack, secondsToWait(pkg, target, reporter) * 1000, target.region),
-      CullElasticSearchInstancesWithTerminationTag(pkg, parameters.stage, stack, secondsToWait(pkg, target, reporter) * 1000, target.region),
-      ResumeAlarmNotifications(pkg, parameters.stage, stack, target.region)
+      CheckGroupSize(asgName, target.region),
+      WaitForElasticSearchClusterGreen(asgName, secondsToWait(pkg, target, reporter) * 1000, target.region),
+      SuspendAlarmNotifications(asgName, target.region),
+      TagCurrentInstancesWithTerminationTag(asgName, target.region),
+      DoubleSize(asgName, target.region),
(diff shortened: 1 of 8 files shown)


magenta-lib/src/main/scala/magenta/deployment_type/ElasticSearch.scala
```diff
@@ -37,17 +37,16 @@ object ElasticSearch extends DeploymentType {
   ) { (pkg, resources, target) =>
```
jacobwinch, line 37: I considered removing this deployment type rather than including it in the refactor (https://github.com/guardian/riff-raff/pull/628), but it's unclear whether we can do this or not yet, so I've just ported the changes across for now.

magenta-lib/src/main/scala/magenta/tasks/ASGTasks.scala
```diff
@@ -162,16 +162,14 @@ case class ResumeAlarmNotifications(pkg: DeploymentPackage, stage: Stage, stack:
...
   def region: Region
-  def pkg: DeploymentPackage
-  def stage: Stage
-  def stack: Stack
 
   def execute(asg: AutoScalingGroup, resources: DeploymentResources, stopFlag: => Boolean, asgClient: AutoScalingClient)
 
   override def execute(resources: DeploymentResources, stopFlag: => Boolean) {
     ASG.withAsgClient(keyRing, region, resources) { asgClient =>
-      val group = ASG.groupForAppAndStage(pkg, stage, stack, asgClient, resources.reporter)
```
jacobwinch, line 174: This is the main change described by:

> Previously, Riff-Raff would search for the relevant ASG (by tags) as part of every individual task (e.g. WaitForStabilization) in a deployment action (e.g. deploy).

> Each individual task can then [now] simply ask for the latest state of the specific ASG it cares about, rather than searching through all ASGs in the account.

magenta-lib/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
```diff
@@ -0,0 +1 @@
+mock-maker-inline
```
jacobwinch, line 1: https://github.com/mockito/mockito-scala#mocking-scala-object

magenta-lib/src/main/scala/magenta/tasks/ASGTasks.scala
akash1810: Does this mean the `riff-raff.yaml` would need to list the ASGs to work on explicitly?

jacobwinch: We still find the ASG using tags (so there are no changes to the `riff-raff.yaml`), it just happens once per `deploy` action ([here](https://github.com/guardian/riff-raff/pull/631/files#diff-c63b6a3e501bff7096061d2ceaf6514397e592cd3440d40a5dc5cf0b96d72f9dR72)) instead of once per task. Happy to chat about this via a call if it's hard to understand (it took me a while to get my head around this code)!

akash1810 (approved):

PR #632, opened 2021-06-17
Add support for dual-ASG deployment (in support of cdk migration)

jacobwinch:
## What does this change?

In order to migrate an application from its current (CloudFormation-defined) infrastructure to new (cdk-defined) infrastructure, Riff-Raff must be able to update two Autoscaling Groups (for the same`App`) as part of a single deployment. Prior to this PR it was impossible to do this. 

This PR allows users to instruct Riff-Raff to update both of their Autoscaling Groups by setting a new parameter in their `riff-raff.yaml`. Riff-Raff will check that it's safe to proceed by ensuring that it finds Autoscaling Groups with appropriate tags. If it finds any other AutoScaling groups for the same `App`/`Stack`/`Stage`, it will still refuse to proceed with the deployment.

## How to test

* I have confirmed that deploys which _do not_ set this parameter are unaffected.
* I have confirmed that deploys which set this parameter are able to update two ASGs (given that these ASGs are tagged correctly).

* It is possible to complete a migration to `cdk`, following [this guide](https://docs.google.com/document/d/17M6TK4uVcMiZAHaTFSo-zahIqrA5eGaKNYU-j4BVUB0/edit).
* It is possible to deploy changes via Riff-Raff during a long-running `cdk` migration.

* Deployments which enable this mode will take longer to complete because the Autoscaling Groups are updated in sequence. However, I think this is acceptable as we only expect users to run with this mode enabled for a short period of time (for more details see the [suggested migration process](https://docs.google.com/document/d/17M6TK4uVcMiZAHaTFSo-zahIqrA5eGaKNYU-j4BVUB0/edit)).
* This deployment type is used by many projects in the department, including our highest-profile projects. After merging this we should keep a close eye on the deployment failure rate in case we have introduced any unexpected problems.

### DIFF
--- a/magenta-lib/src/main/scala/magenta/deployment_type/AutoScaling.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/AutoScaling.scala
@@ -3,14 +3,14 @@ package magenta.deployment_type
 import magenta.{DeployTarget, DeploymentPackage, DeploymentResources, KeyRing}
 import magenta.tasks._
 
-sealed trait CdkTagRequirements
-case object MustBePresent extends CdkTagRequirements
-case object MustNotBePresent extends CdkTagRequirements
+sealed trait MigrationTagRequirements
+case object MustBePresent extends MigrationTagRequirements
+case object MustNotBePresent extends MigrationTagRequirements
 
 object AutoScalingGroupLookup {
-  def getTargetAsgName(keyRing: KeyRing, target: DeployTarget, cdkTagRequirements: Option[CdkTagRequirements], resources: DeploymentResources, pkg: DeploymentPackage) = {
+  def getTargetAsgName(keyRing: KeyRing, target: DeployTarget, migrationTagRequirements: Option[MigrationTagRequirements], resources: DeploymentResources, pkg: DeploymentPackage) = {
     ASG.withAsgClient[String](keyRing, target.region, resources) { asgClient =>
-      ASG.groupForAppAndStage(pkg, target.parameters.stage, target.stack, cdkTagRequirements, asgClient, resources.reporter).autoScalingGroupName()
+      ASG.groupForAppAndStage(pkg, target.parameters.stage, target.stack, migrationTagRequirements, asgClient, resources.reporter).autoScalingGroupName()
     }
   }
 }
@@ -57,7 +57,7 @@ object AutoScaling extends DeploymentType {
     "Whether the uploaded artifacts should be given the PublicRead Canned ACL"
   ).default(false)
 
-  val cdkMigrationInProgress = Param[Boolean]("cdkMigrationInProgress",
+  val asgMigrationInProgress = Param[Boolean]("asgMigrationInProgress",
     "When this is set to true, Riff-Raff will search for two autoscaling groups and deploy to them both"
   ).default(false)
 
@@ -95,7 +95,7 @@ object AutoScaling extends DeploymentType {
         ResumeAlarmNotifications(autoScalingGroupName, target.region)
       )
     }
-    val groupsToUpdate: List[String] = if (cdkMigrationInProgress(pkg, target, reporter)) {
+    val groupsToUpdate: List[String] = if (asgMigrationInProgress(pkg, target, reporter)) {
       List(
         AutoScalingGroupLookup.getTargetAsgName(keyRing, target, Some(MustNotBePresent), resources, pkg),
         AutoScalingGroupLookup.getTargetAsgName(keyRing, target, Some(MustBePresent), resources, pkg)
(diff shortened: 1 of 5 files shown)


magenta-lib/src/main/scala/magenta/deployment_type/AutoScaling.scala
```diff
@@ -3,10 +3,14 @@ package magenta.deployment_type
 import magenta.{DeployTarget, DeploymentPackage, DeploymentResources, KeyRing}
 import magenta.tasks._
 
+sealed trait CdkTagRequirements
+case object MustBePresent extends CdkTagRequirements
+case object MustNotBePresent extends CdkTagRequirements
+
 object AutoScalingGroupLookup {
-  def getTargetAsgName(keyRing: KeyRing, target: DeployTarget, resources: DeploymentResources, pkg: DeploymentPackage) = {
+  def getTargetAsgName(keyRing: KeyRing, target: DeployTarget, cdkTagRequirements: Option[CdkTagRequirements], resources: DeploymentResources, pkg: DeploymentPackage) = {
```
jacobwinch, line 11: We could also use `cdkTagRequirements: Option[Boolean]` here, but I felt that the `sealed trait` helped to make things a little clearer.

akash1810: `Option[Boolean]` has three values too, which would be confusing 😄 .

magenta-lib/src/main/scala/magenta/tasks/AWS.scala
```diff
@@ -240,14 +241,24 @@ object ASG {
...
+  def groupForAppAndStage(pkg: DeploymentPackage, stage: Stage, stack: Stack, cdkTagRequirements: Option[CdkTagRequirements], client: AutoScalingClient, reporter: DeployReporter): AutoScalingGroup = {
     case class ASGMatch(app:App, matches:List[AutoScalingGroup])
 
     implicit class RichAutoscalingGroup(asg: AutoScalingGroup) {
       def hasTag(key: String, value: String): Boolean = asg.tags.asScala.exists { tag =>
         tag.key == key && tag.value == value
       }
-      def matchApp(app: App, stack: Stack): Boolean = hasTag("Stack", stack.name) && hasTag("App", app.name)
+      def meetsCdkTagRequirements(cdkTagRequirements: Option[CdkTagRequirements]) = {
+        val cdkTagIsPresent: Boolean = asg.tags.asScala.exists { tag => tag.key == "gu:cdk:pattern-name" }
```
akash1810, line 252: For extra clarity in the future, might be worth adding some comments around this? Or linking to the PR that added the tag?

magenta-lib/src/test/scala/magenta/deployment_type/AutoScalingTest.scala
```diff
@@ -56,6 +56,54 @@ class AutoScalingTest extends AnyFlatSpec with Matchers with MockitoSugar with A
...
+
+    val app = App("app")
+
+    val p = DeploymentPackage("app", app, data, "autoscaling", S3Path("artifact-bucket", "test/123/app"),
+      deploymentTypes)
+
+    withObjectMocked[AutoScalingGroupLookup.type] {
+      when(AutoScalingGroupLookup.getTargetAsgName(*, *, *, *, *)).thenReturn("testOldCfnAsg", "testNewCdkAsg")
+      val actual = AutoScaling.actionsMap("deploy").taskGenerator(p, DeploymentResources(reporter, lookupEmpty, artifactClient, stsClient, global), DeployTarget(parameters(), stack, region))
+      val expected = List(
```
akash1810, line 73: minor: worth adding a comment to delimit ASG 1 and ASG 2?

akash1810 (commented):
Looks great!

I wonder if there's benefit in having RiffRaff be ignorant to GuCDK? That is, rather than look for a `gu:cdk:pattern-name` tag, should it look for `gu:riffraff:asg-migration` (naming TBD!) as this will mean other migrations can benefit w/out needing to first get onto CDK. For example, the current identity migration.

jacobwinch:
> I wonder if there's benefit in having RiffRaff be ignorant to GuCDK? That is, rather than look for a `gu:cdk:pattern-name` tag, should it look for `gu:riffraff:asg-migration` (naming TBD!)

Yes, perhaps that would be better. The only downside is that we'll need to ask users to add/remove this tag at the appropriate times during the migration.

I'm not that familiar with the Identity scenario, perhaps we can chat about other migration scenarios via a call?

magenta-lib/src/main/scala/magenta/deployment_type/AutoScaling.scala
```diff
@@ -53,6 +57,10 @@ object AutoScaling extends DeploymentType {
     "Whether the uploaded artifacts should be given the PublicRead Canned ACL"
   ).default(false)
 
+  val asgMigrationInProgress = Param[Boolean]("asgMigrationInProgress",
```
akash1810, line 60: minor: Do we want to mention that their tags should be the same, with the new one having the `gu:riffraff:new-asg` tag? Not entirely sure where this string gets used though...

akash1810 (approved):
Lovely!

PR #635, opened 2021-08-10
chore: Follow CI best practice

akash1810:
## What does this change?

[Best practice](https://github.com/guardian/recommendations/blob/master/continuous-integration.md#platforms) has been defined as having a `script/ci` file in the repository to get the benefits of VCS.

This change also makes the TeamCity build more accurate. We're currently using the `sbt` runner in TeamCity which always exits 0, even when there are failing tests! This results in artifacts being always uploaded. This is undesirable! As an example, build [1632 failed](https://teamcity.gutools.co.uk/buildConfiguration/Tools_Riffraff/678317) but is available to be deployed and [did](https://riffraff.gutools.co.uk/deployment/history?projectName=tools%3A%3Ariffraff&stage=PROD&pageSize=20&page=1)!

This requires a change to the TeamCity configuration and other branches would need to rebase in order to continue building.

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -8,7 +8,7 @@ addSbtPlugin("com.typesafe.sbt" % "sbt-less" % "1.1.0")
 addSbtPlugin("com.typesafe.sbt" % "sbt-digest" % "1.1.4")
 
 addSbtPlugin("com.gu" % "sbt-teamcity-test-reporting-plugin" % "1.5")
-addSbtPlugin("com.gu" % "sbt-riffraff-artifact" % "0.9.4")
+addSbtPlugin("com.gu" % "sbt-riffraff-artifact" % "1.1.12")
 
 addSbtPlugin("com.eed3si9n" % "sbt-buildinfo" % "0.6.1")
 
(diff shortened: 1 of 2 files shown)


jacobwinch (approved):

project/plugins.sbt
```diff
@@ -8,8 +8,8 @@ addSbtPlugin("com.typesafe.sbt" % "sbt-less" % "1.1.0")
...
 
 addSbtPlugin("com.gu" % "sbt-teamcity-test-reporting-plugin" % "1.5")
-addSbtPlugin("com.gu" % "sbt-riffraff-artifact" % "0.9.4")
+addSbtPlugin("com.gu" % "sbt-riffraff-artifact" % "1.1.12")
 
 addSbtPlugin("com.eed3si9n" % "sbt-buildinfo" % "0.6.1")
 
-addSbtPlugin("net.virtual-void" % "sbt-dependency-graph" % "0.9.2")
\ No newline at end of file
+addSbtPlugin("net.virtual-void" % "sbt-dependency-graph" % "0.9.2")
```
johnduffell, line 15: fyi if you update to SBT 1.4 or later (or it might be 1.5) then you can remove this unmaintained plugin, as the things needed for snyk are built in.

PR #642, opened 2021-08-12
Favourite projects on the deployments page

tomrf1:
With this change you can add projects as favourites. These will appear as buttons at the top of the form.
They're stored in a local storage item.

### Button to add a favourite is next to the project name input:
Screen Shot 2021-08-12 at 10 15 35

### Favourite buttons appear above (if there are any):
Screen Shot 2021-08-12 at 10 14 59


riff-raff/app/utils/Favourites.scala
```diff
@@ -0,0 +1,35 @@
+package utils
+
+
+import java.util.Base64
+import com.fasterxml.jackson.core.JsonParseException
+import com.fasterxml.jackson.databind.JsonMappingException
+import play.api.libs.json.{ Json => PlayJson }
+import play.api.mvc.Cookie
+
+object Favourites {
```
tomrf1, line 10: shamelessly copied from janus

sihil:
Love the idea. Given that it's local to a machine I wondered if you could use local storage instead of cookies and remove the entirity of the server side code?

tomrf1:
> Love the idea. Given that it's local to a machine I wondered if you could use local storage instead of cookies and remove the entirity of the server side code?

Good point, I had followed janus's example but will have a go at keeping it client-side instead (though I'm missing react!).

riff-raff/app/assets/javascripts/form-autocomplete.coffee
```diff
@@ -20,6 +21,62 @@ updateDeployInfo = () ->
...
+    if selectedProject?
+      deleteFavourite(selectedProject)
+
+  $('.select-favourite-project-button').click (e) ->
+    e.preventDefault()
+    project = e.target.value
+
+    elemProjectInput = $('#projectInput')
+    elemProjectInput.val(project)
+    updateDeployInfo()
```
tomrf1, line 58: I had to put all the favourites stuff in this script because it needs to be able to call `updateDeployInfo`, and it doesn't appear to be possible to import functions from other files

mxdvl: Not sure how coffeeScript works, but just wanted to mention there’s a nice `@guardian/lib` storage tool: https://github.com/guardian/libs/blob/main/src/storage.README.md

sihil:
All of this stuff is ancient and doesn't even use NPM for building or pulling in libraries. Thanks @tomrf1 for getting this done. If you fancy using 10% time to reactify Riff-Raff then that would be lovely 😁

PR #647, opened 2021-09-08
feat: Show if a user is a superuser

akash1810:
## What does this change

There are a few scenarios where only a superuser can perform a task.

Rather than guessing who is a superuser or reading the configuration for the app, add a column to `/auth/list`.

### Before
image

### After
image


jfsoul:
I would've thought it slightly more natural to leave the delete button in the final column and add this as the penultimate?

jfsoul (approved):

akash1810:
> I would've thought it slightly more natural to leave the delete button in the final column and add this as the penultimate?

Good idea! I've made that change now and force pushed.

sihil:
Having looked at this in prod, this might be nicer with tick for superuser and blank for others as it's more visually stricking and easier to find a superuser.

akash1810:
> Having looked at this in prod, this might be nicer with tick for superuser and blank for others as it's more visually stricking and easier to find a superuser.

Good idea! https://github.com/guardian/riff-raff/pull/648

PR #650, opened 2021-09-20
Improve the reporting of ASG discovery

sihil:
## What does this change?
This improves the way that RiffRaff displays tasks that act on AWS autoscaling groups in two ways:
 - the tasks now contain the related ASG name
 - the verbose discovery logic has been improved to exhaustively list the tag requirements

Some housekeeping was done to:
 - switch the ASG helpers to use the paginators available in SDKv2 rather than implementing our own custom recursive function
 - refactor the lookup code to take generic tag requirements and move the ASG specific logic mapping into the AutoScaling deployment class
 - remove the unused and probably dangerous `elasticsearch` deployment type

https://trello.com/c/oj4zjP28/853-improve-asg-identification-in-riffraffs-autoscaling-deploy-type

## How to test
Run a deploy in CODE and verify that there is now an ASG name on relevant tasks and more information in the logs.

Before:
image

After:
image

Deploy logs will now contain much clearer information about which ASG was chosen and how it was chosen.


jacobwinch (approved):
Great work - this should make debugging failed deployments much simpler 💯🚢

PR #652, opened 2021-09-24
Revert "Prevent deletion and replacement of DNS infrastructure"

jacobwinch:
Reverts guardian/riff-raff#651


akash1810 (approved):
Looks like the custom type needs to live in the account first, which is reasonable!

jacobwinch:
> Looks like the custom type needs to live in the account first, which is reasonable!

Yep, makes sense. I only tested deploying one app (which lives in the Deploy Tools account) and didn't consider that this would behave differently in other accounts!

PR #653, opened 2021-09-24
Prevent deletion and replacement of DNS infrastructure

jacobwinch:
~~**This breaks CFN deploys in accounts where the `Guardian::DNS::RecordSet` has not been registered. It should not be merged until the private resource type has been registered with all accounts**~~

This change now detects whether or not 'Guardian::DNS::RecordSet' has been registered  as part of a deploy and acts accordingly.  As such it is ok to deploy.

## What does this change?

In https://github.com/guardian/riff-raff/pull/623 we introduced the concept of `sensitiveResourceTypes`. This is essentially a list of infrastructure resources that we don't want Riff-Raff to delete/replace during CloudFormation updates (unless the user has specifically enabled a more dangerous mode of deployment - see https://github.com/guardian/riff-raff/pull/626 and https://github.com/guardian/riff-raff/pull/627).

This PR adds resource types which manage DNS infrastructure to the list of `sensitiveResourceTypes`. 

This change should prevent users from accidentally deleting DNS records, which is particularly important now as we're about to allow/encourage users to manage NS1 DNS infrastructure via CloudFormation.

In addition, this change detects whether the 'Guardian::DNS::RecordSet' has been registered for the project being deployed and only includes it in the sensitiveResourceTypes list if it is registered.  As such it is backwards compatible and safe for projects that do not yet use this resource.

Logging was added to the riffraff deploy to display the stack policy so that it is easy to determine what resources have been applied during the deploy

**Testing** 
Unit tests ran successfully
Testing in CODE completed successfully.

**Details of testing in CODE**
This version of riffraff was deployed to CODE and then used to deploy one project with the 'Guardian::DNS::Recordset' registered (amigo) and one project without 'Guardian::DNS::Recordset' (frontend). 
Both deployed successfully. 
The logs of the deploy showed Guardian:DNS::Recordset policy was applied during deploy for the project which had the 'Guardian::DNS::Recordset' registered (amigo), but was not applied for the project which did not have this registered (frontend). This is the behaviour we expected and so the test is successful.

### DIFF
--- a/magenta-lib/src/main/scala/magenta/deployment_type/CloudFormation.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/CloudFormation.scala
@@ -1,11 +1,10 @@
 package magenta.deployment_type
 
-import magenta.Strategy.{MostlyHarmless, Dangerous}
-
-import java.net.URLEncoder
-import java.util.UUID
+import magenta.Strategy.{Dangerous, MostlyHarmless}
 import magenta.artifact.S3Path
 import magenta.deployment_type.CloudFormationDeploymentTypeParameters._
+import magenta.tasks.CloudFormation.withCfnClient
+import magenta.tasks.StackPolicy.privateSensitiveResourceTypes
 import magenta.tasks.UpdateCloudFormationTask.LookupByTags
 import magenta.tasks._
 import org.joda.time.DateTime
@@ -82,7 +81,7 @@ object CloudFormation extends DeploymentType with CloudFormationDeploymentTypePa
       |
       |The two stack policies are show below.
       |
-      |${StackPolicy.toMarkdown(StackPolicy.DENY_REPLACE_DELETE_POLICY)}
+      |${StackPolicy.toMarkdown(StackPolicy.DENY_REPLACE_DELETE_POLICY(privateSensitiveResourceTypes))}
       |
       |${StackPolicy.toMarkdown(StackPolicy.ALLOW_ALL_POLICY)}
       |""".stripMargin,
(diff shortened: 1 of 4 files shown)


akash1810 (approved):
🎉

magenta-lib/src/main/scala/magenta/tasks/SetStackPolicyTask.scala
```diff
@@ -1,82 +1,110 @@
...
+  def accountPrivateTypes(client: CloudFormationClient): Set[String] = {
+    val request =  ListTypesRequest.builder().visibility(Visibility.PRIVATE).build()
+    val response = client.listTypesPaginator(request)
+    response.typeSummaries().asScala.map(_.typeName()).toSet
+  }
+
+  /** CFN resource types that have state or are likely to exist in external
+   * config such as DNS or application config.
+   */
+  val sensitiveResourceTypes: Set[String] =
```
akash1810, line 32: minor: I think naming this `allSensitiveResourceTypes` (or similar) would make the tests in `SetStackPolicyTaskTest` a bit easier to read, specifically [this bit](https://github.com/guardian/riff-raff/pull/653/files#diff-340a57d05e31707b6da53a228e9b7c2894479d7b842dbc3bde130de131d8b823R82).

michaelwmcnamara:
Retested latest changes in the manner listed in the PR description above. All tests passed.
Latest changes pass all unit tests 
Deployed projects that have the 'Guardian::DNS::Recordset' registered had this recordset applied.
Deployed projects that did not have this recordset registered did not have it applied.

magenta-lib/src/main/scala/magenta/deployment_type/CloudFormation.scala
```diff
@@ -157,21 +158,27 @@ object CloudFormation extends DeploymentType with CloudFormationDeploymentTypePa
       )
     )
 
+    val policy = target.parameters.updateStrategy match {
```
sihil, line 161: The naming of this val here doesn't make it very clear what the policy is for.

michaelwmcnamara: renamed to updatePolicy

PR #654, opened 2021-09-28
fix: Use safer `resourceStatusAsString` method

akash1810:
## What does this change?

There's been an odd occasion where the CloudFormation log in Riff-Raff shows an event as `null`.

`resourceStatus` is an enum and from [the docs](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/cloudformation/model/StackEvent.html#resourceStatus--):

> If the service returns an enum value that is not available in the current SDK version, resourceStatus will return ResourceStatus.UNKNOWN_TO_SDK_VERSION. The raw value returned by the service is available from resourceStatusAsString().

That is [`resourceStatusAsString`](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/cloudformation/model/StackEvent.html#resourceStatusAsString--) is safer to use.


akash1810:
This is similar to a fix made to Prism https://github.com/guardian/prism/pull/162.

sihil (approved):

PR #656, opened 2021-10-19
Add warning around changing ASG

mxdvl:
## What does this change?

Adds a word of warning:

> Be aware that reducing the number of instances can lead to outages if they are under stress. Always check with the relevant team before performing this action.

## How to test

Navigate to the page: `/docs/howto/fix-a-failed-deploy`


mxdvl:
This is from Github.com, so it’s fresh as a daisy!

PR #659, opened 2021-11-12
Tweaks to favourite project UI

tjmw:
## What does this change?

I ❤️  the new favourite project feature in Riff Raff. This PR makes a couple of small tweaks to the UI:

### 1. Make it harder to favourite an empty project name

I'd like to have the UI only allow you to favourite when you've entered a "real" project name, but I think this incremental improvement is worth adding along the way.

Before:

After:

### 2. Make it harder to favourite duplicate projects

Before:

After:

## How to test

1. Try click the favourite star when the project name field is empty. The button should be disabled and a new favourite should not be added.

2. Favourite a project, then try favouriting the same project again. A duplicate should not be added.

## Accessibility

- [ ] [Tested with screen reader](https://guardian.github.io/source/?path=/docs/docs-06-accessibility--page#screen-readers)
- [ ] [Navigable with keyboard](https://guardian.github.io/source/?path=/docs/docs-06-accessibility--page#keyboard-navigation)
- [ ] [Colour contrast passed](https://guardian.github.io/source/?path=/docs/docs-06-accessibility--page#colour-contrast)
- [ ] [The change doesn't use only colour to convey meaning](https://guardian.github.io/source/?path=/docs/docs-06-accessibility--page#use-of-colour)

### DIFF
--- a/riff-raff/app/assets/javascripts/form-autocomplete.coffee
+++ b/riff-raff/app/assets/javascripts/form-autocomplete.coffee
@@ -85,6 +85,14 @@ $ ->
       source:serverUrl
       minLength:0
 
+    addFavouriteProjectButton = $('#add-favourite-project-button')
+    updateFavouriteButton = ->
+      projectInputIsEmpty = input.val().trim() == ''
+      addFavouriteProjectButton.prop('disabled', projectInputIsEmpty)
+
+    input.on('change', updateFavouriteButton)
+    input.on('keyup', updateFavouriteButton)
+
   $('#projectInput').blur updateDeployInfo
 
   $('#buildInput').each ->
@@ -126,7 +134,7 @@ $ ->
     elemProjectInput = $('#projectInput')
     selectedProject = elemProjectInput.val()
 
-    if selectedProject?
+    if selectedProject
       addFavourite(selectedProject)
 
   renderFavourites()
--- a/riff-raff/app/assets/javascripts/form-autocomplete.coffee
+++ b/riff-raff/app/assets/javascripts/form-autocomplete.coffee
@@ -21,25 +21,34 @@ updateDeployInfo = () ->
       $("[rel='tooltip']").tooltip()
   )
 
+readFavourites = () ->
+  JSON.parse(localStorage.getItem('favouriteProjects'))
+
+writeFavourites = (newFavourites) ->
+  localStorage.setItem('favouriteProjects', JSON.stringify(newFavourites))
+
 addFavourite = (project) ->
-  favourites = JSON.parse(localStorage.getItem('favouriteProjects'))
+  favourites = readFavourites()
   newFavourites =
     if favourites?
-      favourites.push(project)
+      projectAlreadyFavourited = favourites.includes(project)
+      if !projectAlreadyFavourited
+        favourites.push(project)
+
       favourites
     else
       [project]
-  localStorage.setItem('favouriteProjects', JSON.stringify(newFavourites))
+  writeFavourites(newFavourites)
   renderFavourites()
 
 deleteFavourite = (project) ->
-  favourites = JSON.parse(localStorage.getItem('favouriteProjects'))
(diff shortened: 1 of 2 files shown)


riff-raff/app/assets/javascripts/form-autocomplete.coffee
```diff
@@ -85,6 +94,13 @@ $ ->
       source:serverUrl
       minLength:0
 
+    addFavouriteProjectButton = $('#add-favourite-project-button')
+    updateFavouriteButton = ->
+      console.log("fave change")
+      addFavouriteProjectButton.prop('disabled', /^\s*$/.test(input.val()))
+    input.on('change', updateFavouriteButton)
+    input.on('keyup', updateFavouriteButton)
```
akash1810, line 102: What's this doing? My assumption is it's removing the `disabled` attribute?

tjmw: Yep that's correct. We're listening on the keyup event and calling the `updateFavouriteButton` callback when it happens (the callback is defined just above).

tjmw: To clarify further: in the callback we set the `disabled` prop of the button, depending on whether the project input is empty or not.

akash1810: 👍🏽  I'm a little confused by the regex - could we use [`removeAttribute`](https://developer.mozilla.org/en-US/docs/Web/API/Element/removeAttribute)? Or is this because this is coffeescript?

akash1810: > To clarify further: in the callback we set the `disabled` prop of the button, depending on whether the project input is empty or not.

Ah! Would something like this be clearer?

```js
addFavouriteProjectButton.disabled = input.val() === ""
```

mxdvl:
Just chiming in here to say that you can now embed `.mov` files in PRs directly, so it’s just `cmd+shift+5`!

riff-raff/app/assets/javascripts/form-autocomplete.coffee
tjmw: Ah yep we could definitely do that! No this isn't coffeescript specific I don't think, it's using [`prop` from jQuery](https://api.jquery.com/prop/#prop2) - old school! I quite like this approach as it doesn't require two branches (one for removing and one for adding), we take this approach elsewhere in the codebase too. Would this make it clearer do you think?

```suggestion
      projectInputIsEmpty = /^\s*$/.test(input.val())
      addFavouriteProjectButton.prop('disabled', projectInputIsEmpty)
    input.on('change', updateFavouriteButton)
    input.on('keyup', updateFavouriteButton)
```

tjmw: > addFavouriteProjectButton.disabled = input.val() === ""

Oops our last messages overlapped. Yep I like that approach 👍  The regex was a little extra check which meant you couldn't hit space a bunch of times and favourite _that_ project name, but maybe not worth the additional complexity added?

akash1810: > The regex was a little extra check which meant you couldn't hit space a bunch of times and favourite that project name

`input.val().trim()` would do the same thing and is clearer, at least to me, than regex in this instance.

tjmw: Ah yep that's much better, thanks for the suggestion.

tjmw: I've made this change. Setting `.disabled` directly doesn't seem to have the desired effect (I guess that is a CoffeeScript thing) so I've stuck with using `prop`.

akash1810 (approved):

mchv:
Great addition @tjmw 💎

PR #660, opened 2021-11-16
fix: Do not attempt to terminate twice

akash1810:
## What does this change?

Ignore instances already in a terminating phase when deciding which instances to terminate during an EC2 deployment.

This should address this occasional error:

```
Your ASG desired size may have been reset Instance Id not found - No managed instance found for instance ID: i-abcdefgh123456789
```

## How to test

???

Fewer errors and possibly faster deploys as there are fewer instances to affect.

### DIFF
--- a/magenta-lib/src/main/scala/magenta/tasks/ASGTasks.scala
+++ b/magenta-lib/src/main/scala/magenta/tasks/ASGTasks.scala
@@ -123,7 +123,25 @@ case class CullInstancesWithTerminationTag(info: AutoScalingGroupInfo, region: R
   override def execute(asg: AutoScalingGroup, resources: DeploymentResources, stopFlag: => Boolean, asgClient: AutoScalingClient) {
     EC2.withEc2Client(keyRing, region, resources) { ec2Client =>
       ELB.withClient(keyRing, region, resources) { elbClient =>
-        val instancesToKill = asg.instances.asScala.filter(instance => EC2.hasTag(instance, "Magenta", "Terminate", ec2Client))
+        val instancesToKill = asg.instances.asScala
+          .filter(instance => {
+            if(instance.lifecycleState == LifecycleState.UNKNOWN_TO_SDK_VERSION) {
+              logger.warn(s"Instance lifecycle state ${instance.lifecycleStateAsString} isn't recognised in the AWS SDK. Is there a later version of the AWS SDK available?")
+            }
+
+            // See https://docs.aws.amazon.com/autoscaling/ec2/userguide/lifecycle-hooks.html#lifecycle-hooks-overview
+            val terminatingStates = List(
+              LifecycleState.TERMINATING,
+              LifecycleState.TERMINATING_WAIT,
+              LifecycleState.TERMINATING_PROCEED,
+              LifecycleState.TERMINATED
+            ).map(_.toString)
+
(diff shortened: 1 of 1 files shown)


magenta-lib/src/main/scala/magenta/tasks/ASGTasks.scala
```diff
@@ -123,7 +123,25 @@ case class CullInstancesWithTerminationTag(info: AutoScalingGroupInfo, region: R
   override def execute(asg: AutoScalingGroup, resources: DeploymentResources, stopFlag: => Boolean, asgClient: AutoScalingClient) {
     EC2.withEc2Client(keyRing, region, resources) { ec2Client =>
       ELB.withClient(keyRing, region, resources) { elbClient =>
-        val instancesToKill = asg.instances.asScala.filter(instance => EC2.hasTag(instance, "Magenta", "Terminate", ec2Client))
+        val instancesToKill = asg.instances.asScala
+          .filter(instance => {
+            if(instance.lifecycleState == LifecycleState.UNKNOWN_TO_SDK_VERSION) {
+              logger.warn(s"Instance lifecycle state ${instance.lifecycleStateAsString} isn't recognised in the AWS SDK. Is there a later version of the AWS SDK available?")
+            }
```
akash1810, line 130: Taking a similar approach as https://github.com/guardian/prism/pull/162 to avoid possible NPEs.

magenta-lib/src/main/scala/magenta/tasks/ASGTasks.scala
```diff
@@ -123,7 +123,25 @@ case class CullInstancesWithTerminationTag(info: AutoScalingGroupInfo, region: R
...
+
+            // See https://docs.aws.amazon.com/autoscaling/ec2/userguide/lifecycle-hooks.html#lifecycle-hooks-overview
+            val terminatingStates = List(
+              LifecycleState.TERMINATING,
+              LifecycleState.TERMINATING_WAIT,
+              LifecycleState.TERMINATING_PROCEED,
+              LifecycleState.TERMINATED
+            ).map(_.toString)
+
+            val isAlreadyTerminating = terminatingStates.contains(instance.lifecycleStateAsString)
```
jacobwinch, line 140: Is there a reason that we need to work with strings here? Is it not possible to use `lifecycleState` instead?

jacobwinch (approved):
Great work, thanks for improving this 👍

magenta-lib/src/main/scala/magenta/tasks/ASGTasks.scala
akash1810: `lifecycleStateAsString` is safer as there's a potential for [`lifecycleState`](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/autoscaling/model/Instance.html#lifecycleState--) to be `null` (via `LifecycleState.UNKNOWN_TO_SDK_VERSION`) if the SDK version is old.

We've been burned by this in the past (https://github.com/guardian/prism/pull/162, https://github.com/guardian/riff-raff/pull/654) and I'm now cautious by default.

jacobwinch: Thanks for explaining 👍

johnduffell (approved):
nice work, thanks Akash! 👍

PR #661, opened 2021-11-18
Improve asgMigrationInProgress docs

nicl:
## What does this change?

Improves parameter docs.

See also https://github.com/guardian/riff-raff/pull/632/files#r655291289 which was helpful in me figuring this out when I hit this.

### DIFF
--- a/magenta-lib/src/main/scala/magenta/deployment_type/AutoScaling.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/AutoScaling.scala
@@ -79,7 +79,14 @@ object AutoScaling extends DeploymentType {
   ).default(false)
 
   val asgMigrationInProgress = Param[Boolean]("asgMigrationInProgress",
-    "When this is set to true, Riff-Raff will search for two autoscaling groups and deploy to them both"
+    """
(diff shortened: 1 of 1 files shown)


magenta-lib/src/main/scala/magenta/deployment_type/AutoScaling.scala
```diff
@@ -79,7 +79,14 @@ object AutoScaling extends DeploymentType {
...
 
   val asgMigrationInProgress = Param[Boolean]("asgMigrationInProgress",
-    "When this is set to true, Riff-Raff will search for two autoscaling groups and deploy to them both"
+    """
+      |When this is set to true, Riff-Raff will search for two autoscaling groups and deploy to them both.
+      |
+      |Note, for this to work, the following need to hold:
+      |
+      | - both ASGs need matching Stack,Stage,App tags
+      | - only two matching ASGs (not 3 or more)
```
nicl, line 88: Note to self: this should be be updated from 'only two' to 'exactly two...'.

akash1810 (approved):
When your note to self is complete, this LGTM.

PR #662, opened 2021-12-10
Upgrade SBT to 1.5.6 and Play to 2.8 boom

nicl:
## What does this change?

Bump SBT version to 1.5.6 and related changes.


project/plugins.sbt
```diff
@@ -1,15 +1,12 @@
-resolvers += "Typesafe Releases" at "https://repo.typesafe.com/typesafe/releases/"
```
jacobwinch, line 1: Would you mind explaining why this resolver is no longer required? (Just for my own understanding!)

project/plugins.sbt
```diff
@@ -1,15 +1,12 @@
...
-
 // keep in sync with the play version in Dependencies
```
jacobwinch, line 1: Do you think this comment is still relevant? 

I noticed that the Play version here does not match the version of `play-json` in Dependencies (this was also true before this PR!)

jacobwinch (approved):
Great work 👍

nicl:
@jacobwinch sorry yeah, meant to add WIP. I'm going to close this and split into some pre-PRs (to e.g. remove Guardian Management) to make things easier to review and less gigantic.

PR #663, opened 2021-12-15
Remove Guardian Management deps

nicl:
These dependencies (https://github.com/guardian/guardian-management) were last updated 4 years ago, very little used outsideo of RR (it looks like RR and Flexible Content are the two main users from a quick Github search) and have proven very difficult to update to Scala 2.13+.

As such, we are removing them for security and maintenance reasons.

Switch-related code has been partially inlined from the library as part of this to ease the transition.

Metrics, which were done using Guardian Management and served to a JSON endpoint for consumption by Ganglia (which no longer exists) have been removed completely for now.

A fuller switch solution would be useful here, but is beyond the scope of this immediate change.

**UPDATE: tested on CODE successfully.**


magenta-lib/src/main/scala/magenta/deployment_type/SelfDeploy.scala
```diff
@@ -53,26 +53,16 @@ object SelfDeploy extends DeploymentType {
...
-      ChangeSwitch(
-        host,
-        managementProtocol(pkg, target, reporter),
-        managementPort(pkg, target, reporter),
-        switchboardPath(pkg, target, reporter),
-        "shutdown-when-inactive",
-        desiredState=true
-      )
-    }
+    hosts.map(ShutdownTask.apply)
```
nicl, line 65: Guardian Management automatically presented endpoints to toggle switches on a non-standard port. This has been replaced with a custom endpoint (on the standard 9000) to do this.

magenta-lib/src/main/scala/magenta/logging.scala
```diff
@@ -190,4 +187,8 @@ case object Message extends Enum[Message] with PlayJsonEnum[Message] {
   case class Warning(text: String) extends Message
 
   val values = findValues
+}
+
+trait Loggable {
```
nicl, line 192: A basic `Loggable` trait as a drop-in replacement for the same-named trait in Guardian Management.

magenta-lib/src/main/scala/magenta/switches.scala
```diff
@@ -0,0 +1,65 @@
+// Modified version of https://github.com/guardian/guardian-management/blob/c7d99120927958757b35bea68931c4bd8690d64a/management/src/main/scala/com/gu/management/switchables.scala
```
nicl, line 1: Copied from Guardian Management as rewriting switch logic would have involved wider changes. Not all of the switch logic from Guardian Management has been copied though, only the basic switch models.

riff-raff/app/conf/context.scala
```diff
@@ -269,98 +255,3 @@ class Config(configuration: TypesafeConfig, startTime: DateTime) extends Logging
 
   override def toString: String = configuration.toString
 }
-
-class Management(config: Config, shutdownWhenInactive: ShutdownWhenInactive, deployments: Deployments, datastore: DataStore) {
```
nicl, line 273: We collected a load of metrics using Guardian Management that look interesting but aren't actually in use from what I can tell. Open Telemetry/something else is probably the future here.

riff-raff/app/deployment/Deployments.scala
```diff
@@ -36,8 +35,6 @@ class Deployments(deploymentEngine: DeploymentEngine,
       requestedParams.build.projectName, requestedParams.stage.name, requestSource)
     if (restrictionsPreventingDeploy.nonEmpty) {
       Left(Error(s"Unable to queue deploy as restrictions are currently in place: ${restrictionsPreventingDeploy.map(r => s"${r.fullName}: ${r.note}").mkString("; ")}"))
-    } else if (enableQueueingSwitch.isSwitchedOff) {
```
nicl, line 39: (I have not recreated the enableQueueing switch as was not convinced of the value, but we can re-add if it's more useful than I realise.)

akash1810 (commented):
These changes look sensible, great work. I'm a little surprised that we've been collecting all these metrics w/out displaying them anywhere, which leads me to think they might have another, subtle, use other than fancy graphs. I can't quite think of what though. Let's run this on CODE to test.

akash1810:
> UPDATE: tested on CODE successfully.

How exhaustively did we test? Do we need to do each deployment type, for example? (Probably not)

magenta-lib/src/main/scala/magenta/tasks/tasks.scala
```diff
@@ -209,40 +208,35 @@ case class SayHello(host: Host)(implicit val keyRing: KeyRing) extends Task {
...
-  val switchboardUrl = s"$protocol://${host.name}:$port$path"
+case class ShutdownTask(host: Host)(implicit val keyRing: KeyRing)  extends Task {
 
-  // execute this task (should throw on failure)
-  override def execute(resources: DeploymentResources, stopFlag: => Boolean): Unit = {
-    resources.reporter.verbose(s"Changing $switchName to $desiredStateName using $switchboardUrl")
+  val description = "Shutdown Riffraff target(s) as part of self-deploy."
 
+  override def execute(resources: DeploymentResources, stopFlag: => Boolean): Unit = {
+    val shutdownUrl = s"https:://${host.name}:9000/shutdown"
```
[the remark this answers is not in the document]
nicl: @akash1810 this actually gets used when deploying PROD Riffraff from CODE and vice versa, so is potentially across the internet. We could probably figure out if it was targeting itself, but I'm not sure it's worth it really.

nicl:
@akash1810 it definitely wasn't an exhaustive test. I basically clicked through the main nav links to check they worked and also did a self-deploy.

nicl:
@akash1810 I am definitely suspicious that the metrics were used elsewhere too! I'll have another look to see if I can find anything else.

riff-raff/app/deployment/Deployments.scala
```diff
@@ -143,6 +125,22 @@ class Deployments(deploymentEngine: DeploymentEngine,
 
   lazy val enableQueueingSwitch = new DefaultSwitch("enable-deploy-queuing", "Enable riff-raff to queue deploys.  Turning this off will prevent anyone queueing a new deploy, although running deploys will continue.", true)
```
jacobwinch, line 126: Based on [this comment](https://github.com/guardian/riff-raff/pull/663/files#r769648481) I think this switch can be removed now?

riff-raff/app/deployment/Deployments.scala
jacobwinch: I don't remember this being used in my time on DevX so I think it's OK to remove it.

riff-raff/app/deployment/Deployments.scala
```diff
@@ -91,27 +89,11 @@ class Deployments(deploymentEngine: DeploymentEngine,
...
-      } catch {
-        case NonFatal(t) =>
-          log.error(s"Exception thrown whilst processing update with $wrapper", t)
-      }
-    }
-  def init() {}
-  def shutdown() { messagesSubscription.unsubscribe() }
-
-
   val deploysEnabled = new Switch(startAsOn = true)
```
jacobwinch, line 92: I noticed that this is using `akka.util.Switch`, which seems to provide pretty similar functionality to some of the code that is being ported across from Guardian Management - I wonder if we could simplify/make this consistent in the future

magenta-lib/src/main/scala/magenta/deployment_type/SelfDeploy.scala
jacobwinch: I guess this allows us to remove [this ingress rule](https://github.com/guardian/deploy-tools-platform/blob/ea015f04e297291ad7b2066095459c273b8e447e/cloudformation/riffraff/riffraff.template.yaml#L141-L146)?

magenta-lib/src/main/scala/magenta/tasks/tasks.scala
```diff
@@ -209,40 +208,39 @@ case class SayHello(host: Host)(implicit val keyRing: KeyRing) extends Task {
...
-case class ChangeSwitch(host: Host, protocol:String, port: Int, path: String, switchName: String, desiredState: Boolean)(implicit val keyRing: KeyRing) extends Task {
-  val desiredStateName: String = if (desiredState) "ON" else "OFF"
-  val switchboardUrl = s"$protocol://${host.name}:$port$path"
+case class ShutdownTask(host: Host)(implicit val keyRing: KeyRing)  extends Task {
 
-  // execute this task (should throw on failure)
-  override def execute(resources: DeploymentResources, stopFlag: => Boolean): Unit = {
-    resources.reporter.verbose(s"Changing $switchName to $desiredStateName using $switchboardUrl")
+  val description = "Shutdown Riffraff target(s) as part of self-deploy."
+  val shutdownUrl = s"http://${host.name}:9000/shutdown"
```
jacobwinch, line 214: Should this be `/requestShutdown`?

magenta-lib/src/main/scala/magenta/tasks/tasks.scala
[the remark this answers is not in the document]
nicl: Updating this, it turns out that there is actually a manual deployment restriction on cross-stage deployments of Riffraff as, while the code itself supports this behaviour, the security groups do not.

riff-raff/app/deployment/Deployments.scala
nicl: Yeah, I think so. It was surprising to me that both were used. This PR is already quite large(!) but this is a good candidate for a follow-up.

jacobwinch (approved):
👍

PR #664, opened 2021-12-31
fix!: Set `publicReadAcl` to `false` by default

akash1810:
Expands on #665.

## What does this change?

Following the principle of least privilege, set the `publicAcl` to `false` by default. This matches the default in the [`S3Upload` task](https://github.com/guardian/riff-raff/blob/dfc93996016cd37df8d4ede2f66f669556a39172/magenta-lib/src/main/scala/magenta/tasks/tasks.scala#L29).

This also fits with AWS's own [docs](https://docs.aws.amazon.com/AmazonS3/latest/userguide/acl-overview.html):

> A majority of modern use cases in Amazon S3 no longer require the use of ACLs, and we recommend that you disable ACLs except in unusual circumstances where you need to control access for each object individually.

See https://github.com/guardian/riff-raff/pull/141 for a discussion about this - the overall conclusion was to have a default of `false` too.

More `s3-upload` deployment types work first time?

This change would break projects that rely on the default being `true` and have not explicitly set the value in their `riff-raff.yaml`. It's pretty difficult to identify these projects - there are a _lot_ or projects that [explicitly set it to `false`](https://github.com/search?p=1&q=org%3Aguardian+publicReadAcl+filename%3Adeploy.json+filename%3Ariff-raff.yml+filename%3Ariff-raff.yaml&type=Code) though!

I'd suggest we address any issues when they come up.

### DIFF
--- a/magenta-lib/src/main/scala/magenta/deployment_type/S3.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/S3.scala
@@ -35,11 +35,14 @@ object S3 extends DeploymentType {
 
   val publicReadAcl = Param[Boolean]("publicReadAcl",
     """
-      |Whether the uploaded artifacts should be given the PublicRead Canned ACL. (Default is true!) Note that the
-      |AWS default S3 bucket setting is to block all public access, which doesn't play nicely with `publicReadAcl`
-      |when set to true. If public access to the bucket is restricted then you should set `publicReadAcl` to `false`.
+      |Whether the uploaded artifacts should be given the PublicRead Canned ACL.
+      |
+      |If public access to the bucket is restricted (which is the AWS default), `publicReadAcl` should be `false`.
+      |If the bucket content needs to be public, it's better to control this via the bucket policy.
+      |
+      |See https://docs.aws.amazon.com/AmazonS3/latest/userguide/acl-overview.html.
     """.stripMargin
-  ).default(true)
+  ).default(false)
 
   val cacheControl = Param[List[PatternValue]]("cacheControl",
     """
--- a/magenta-lib/src/test/scala/magenta/deployment_type/DeploymentTypeTest.scala
+++ b/magenta-lib/src/test/scala/magenta/deployment_type/DeploymentTypeTest.scala
@@ -89,8 +89,7 @@ class DeploymentTypeTest extends AnyFlatSpec with Matchers with Inside with Mock
         "bucket-1234",
         Seq(sourceS3Package -> "test-stack/CODE/myapp"),
         cacheControlPatterns = List(PatternValue(".*", "no-cache")),
-        surrogateControlPatterns = List(PatternValue(".*", "max-age=3600")),
-        publicReadAcl = true
+        surrogateControlPatterns = List(PatternValue(".*", "max-age=3600"))
       ))
     )
   }


akash1810:
cc @guardian/all as this is an important change

rtyley:
How can teams try this out? If this is deployed to [CODE RiffRaff](https://riffraff.code.dev-gutools.co.uk/), could we try deploying our own projects to see if they still work?

akash1810:
> How can teams try this out? If this is deployed to [CODE RiffRaff](https://riffraff.code.dev-gutools.co.uk/), could we try deploying our own projects to see if they still work?

Good idea! [Done](https://riffraff.code.dev-gutools.co.uk/deployment/view/9dc710a1-d5e4-4a38-8daa-cd39bb234f51).

rtyley:
I've deployed pretty much every Ophan project to PROD using the updated Riff-Raff CODE, and they all completed without any visible fatal error! I was only re-deploying the current builds everywhere, as I assume that's sufficient to prove there's no breakage from this change? No point in deploying a different, new, build number?

image

There was a nice informative message [on deploy of `ophan::Elasticsearch Debian`](https://riffraff.code.dev-gutools.co.uk/deployment/view/c96a84bd-2b1a-42c2-9386-3b71270f8006): _"Parameter publicReadAcl is unnecessarily explicitly set to the default value of false"_

image

...the message is already present without any additional logging from this PR, as it's in this existing code, which is pretty cool:

https://github.com/guardian/riff-raff/blob/0b43d431c3a75f2fa38b21fdb9e210cdc19d86a9/magenta-lib/src/main/scala/magenta/deployment_type/Param.scala#L46-L47

jfsoul:
I'm trying to think of the scenarios in which this would break, and _how_ it would break. The thing that troubles me about switching the default like this (not that I disagree that the change is needed) is that this could cause successful deploys of apps to break production apps e.g:
- Some user facing web app has some static assets deployed to s3 via `aws-s3` deployment type and relies on those assets being made public via ACLs
- It does not set the `publicReadAcl` to `true` explicitly
- We release this change and riff-raff will happily deploy (potentially unsupervised on a schedule) the app
- ACL is changed, static assets aren't reachable in S3, and the web app is potentially very broken

I don't think this will happen everywhere, but there might be a few apps set up this way. I wonder whether a softer approach might be prudent:
1. Either make omitting `publicReadAcl` a warning, or an error and fail the deploy completely **and then release this change later** once we have a bit more confidence or feel for how often this we might hit the above scenario
2. OR we can incorporate the warning/error into this PR and release all at once. I'd rather a failed deploy than a silent breakage.

tomrf1:
I like @jfsoul 's suggestion of a 2-step roll out.
Also it might be helpful to go through [all uses](https://github.com/search?l=YAML&q=org%3Aguardian+aws-s3&type=Code) of `aws-s3` where `publicReadAcl` is unset and confirm it's safe

akash1810:
@jfsoul I agree, to a point. The place where this breaks down is those projects that update PROD using continuous deployment or scheduled deployment. That is, I'd be hesitant about doing mitigation work that only solves a subset of the projects.

jfsoul:
| any UI based warnings we add will never get seen

I agree, and it's why I'd go for the "fail the deploy" option personally.

akash1810:
> > any UI based warnings we add will never get seen
> 
> I agree, and it's why I'd go for the "fail the deploy" option personally.

Oops, totally missed that part! Apologies. I like this solution and will have a go at implementing it.

johnduffell:
> this is an important change that might break some projects

Thanks for proposing the change.  I feel nervous because the last time something went out with this kind of warning on it, it inadvertantly broke support-frontend cloudformation in PROD after the scheduled deploy (took it down and lost some cloudformation parameters), so I definitely support a gentle way of rolling this out!  Having said that, this would be a much easier fix as it just involves shipping an updated parameter.

Just thinking of other ways to reduce the risk - perhaps a search of github or recently deployed riff-raff.yaml deploy files for aws-s3 usage and a check of each one would be worth scanning through?  Or could we use the angammarad tool in some clever way?

akash1810:
> Oops, totally missed that part! Apologies. I like this solution and will have a go at implementing it on this branch.

It was a little clearer to make the change in a different PR https://github.com/guardian/riff-raff/pull/665.

akash1810:
The changes in #665 have meant that we're now explicit about our choice of `publicReadAcl`. I think this is better than having a default as it encourages an audit trail about the choice via pull requests. That is, I think _not_ having a default for this value is best and this PR can be closed.

PR #665, opened 2022-01-04
feat!: Make `publicReadAcl` parameter to `s3-upload` type required

akash1810:
## What does this change?

This is part of a migration effort to changing the default of `publicReadAcl` to `false` (https://github.com/guardian/riff-raff/pull/664).

By marking the parameter as required (aka not optional) an error will be thrown complaining about a missing parameter, meaning:
  - the deploy fails until the parameter is explicitly set
  - previous deploys are not effected, therefore the app still works

BREAKING CHANGE: `publicReadAcl` is now required to be set for `s3-upload` type.

## How to test

Update your `riff-raff.yaml`, removing the `publicReadAcl` parameter and try to deploy it.

OR try to deploy a build from https://github.com/guardian/cdk-static-site/pull/1.

It is easier to set the default of `publicReadAcl` to `false` as we can guarantee projects that are relying on the current default of `true` have explicitly set the value.

## Images

An example of the UX you're greeted with when `publicReadAcl` isn't set.

image

### DIFF
--- a/magenta-lib/src/main/scala/magenta/deployment_type/S3.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/S3.scala
@@ -35,11 +35,15 @@ object S3 extends DeploymentType {
 
   val publicReadAcl = Param[Boolean]("publicReadAcl",
     """
-      |Whether the uploaded artifacts should be given the PublicRead Canned ACL. (Default is true!) Note that the
-      |AWS default S3 bucket setting is to block all public access, which doesn't play nicely with `publicReadAcl`
-      |when set to true. If public access to the bucket is restricted then you should set `publicReadAcl` to `false`.
-    """.stripMargin
-  ).default(true)
+      |Whether the uploaded artifacts should be given the PublicRead Canned ACL.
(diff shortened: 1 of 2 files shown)


jfsoul (approved):
Separately we could go a step further and make `publicReadAcl=true` a warning or otherwise somehow mark this option as deprecated, if we really believe that the bucket policy is the preferred way to make objects public.

akash1810:
> Separately we could go a step further and make `publicReadAcl=true` a warning or otherwise somehow mark this option as deprecated, if we really believe that the bucket policy is the preferred way to make objects public.

The more I think about it, the less convinced we ever want to make a bucket public. IIUC there is one reason to make a bucket totally public: for static sites. However, access to the bucket can/should be controlled via IAM instead such as [this for CloudFront](https://aws.amazon.com/premiumsupport/knowledge-center/cloudfront-serve-static-website/) or [this for Fastly](https://docs.fastly.com/en/guides/amazon-s3#using-an-amazon-s3-private-bucket). That is, more thought needed and for a separate PR.

PR #668, opened 2022-01-06
fix: Use raw YAML to determine Anghammarad targets

akash1810:
## What does this change?

Currently, we validate the `riff-raff.yaml` before identifying targets for an Anghammarad notification. This is fine until we change the rules for what makes a valid `riff-raff.yaml` file!

In this change we remove the validation step and instead use the raw YAML file.

See https://github.com/guardian/riff-raff/pull/665 which is causing deployments to fail if they are missing a parameter.

In the scenario where a deployment fails due to a missing parameter in `riff-raff.yaml`, teams continue to get notified directly.


jacobwinch (approved):
Great work, thanks for improving this 👍

PR #671, opened 2022-01-13
fix: Handle no-op ChangeSets more gracefully

akash1810:
Easiest to review [w/out whitespace](https://github.com/guardian/riff-raff/pull/671/files?diff=split&w=1).

## What does this change?

The `cloud-formation` deployment type in Riff-Raff has a number of steps (well, [tasks](https://github.com/guardian/riff-raff/blob/3bf2aa3529fda6f536a698c712c0186812147477/magenta-lib/src/main/scala/magenta/deployment_type/CloudFormation.scala#L135-L184)):
  1. Set the update policy on the CloudFormation stack to guard against deleting stateful resources
  2. Create a [ChangeSet](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/using-cfn-updating-stacks-changesets.html)
  3. Execute the ChangeSet
  4. Watch the CloudFormation event log to determine when 3 has finished
  5. Revert 1

Currently in step 4, we're only looking at the _first_ page of CloudFormation events and we expect to find an `UPDATE_IN_PROGRESS` or `CREATE_IN_PROGRESS` event in the list. When we don't, the deploy currently [fails](https://github.com/guardian/riff-raff/blob/dacc0dedc6ec23816d213c39e1b7af3e321b772f/magenta-lib/src/main/scala/magenta/tasks/UpdateCloudFormationTask.scala#L347) with `No events found at all for stack...`.

<details>
  <summary>Riff-Raff screenshot</summary>
  
  image
</details>

This happens if a CloudFormation update generates _lots_ of items in the event log, causing the `UPDATE_IN_PROGRESS` or `CREATE_IN_PROGRESS` events off the first page.

In this change we update step 4, exiting early if the change set created in step 2 will not yield changes to the stack. Meaning we only look at the event log when necessary. This logic already happens for [step 3](https://github.com/guardian/riff-raff/blob/0b43d431c3a75f2fa38b21fdb9e210cdc19d86a9/magenta-lib/src/main/scala/magenta/tasks/ChangeSetTasks.scala#L141-L142).

After a discussion with @nicl @jacobwinch and @michaelwmcnamara, we think it would actually be better to check the [`ExecutionStatus`](https://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_DescribeChangeSet.html#:~:text=length%20of%201024.-,ExecutionStatus,-If%20the%20change) of a change set instead of the CloudFormation event log to determine when a change set has finished executing. This also reduces the chance of us observing an event that was _not_ triggered by a deploy (somehow!) but still treating it as though it was part of the deploy. We'll get this added to the backlog.

This change is still valuable though as it results in our skipping of step 4 above all together.

## How to test

1. Deploy `main` to CODE
2. Identify a stack who's CloudFormation event log has `UPDATE_IN_PROGRESS` off the first page.
3. Make a no-op change to a CloudFormation template, e.g. add a `.` to a parameter's description
4. Perform a `cloud-formation` deployment and observe it fail, similar to the above screenshot
5. Deploy this branch to CODE
6. Perform a `cloud-formation` deployment and observe it succeed. Specifically the `CheckUpdateEventsTask` step should succeed with `No changes to perform for ... on stack...` rather than failing with `No events found at all for stack...`.

Fewer deployment failures.

Building of tech dept! As noted in the `TODO`s added in this PR, we'd ideally wrap the `ami-cloudformation-parameter` deployment type in the above ChangeSet logic. This is good as ChangeSet's offer another audit trail but also because `CheckUpdateEventsTask` is shared between the `cloud-formation` type and the `ami-cloudformation-parameter` type. For `CheckUpdateEventsTask` to check if a ChangeSet actually includes any changes, the types within `CheckUpdateEventsTask` need updating. This wasn't immediately trivial. That is, this PR is the pragmatic solution.

### DIFF
--- a/magenta-lib/src/main/scala/magenta/deployment_type/AmiCloudFormationParameter.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/AmiCloudFormationParameter.scala
@@ -14,33 +14,27 @@ object AmiCloudFormationParameter extends DeploymentType with CloudFormationDepl
       |You will need to add this as a dependency to your autoscaling deploy in your riff-raff.yaml to guard against race conditions.
     """.stripMargin
 
-  val update = Action("update",
-    """
-      |Given AMI tags, this will resolve the latest matching AMI and update the AMI parameter
-      | on the provided CloudFormation stack.
-    """.stripMargin
-  ){ (pkg, resources, target) => {
-      implicit val keyRing = resources.assembleKeyring(target, pkg)
-      val reporter = resources.reporter
-      val amiParameterMap: Map[CfnParam, TagCriteria] = getAmiParameterMap(pkg, target, reporter)
-      val cloudFormationStackLookupStrategy = getCloudFormationStackLookupStrategy(pkg, target, reporter)
+  val update = Action("update", "Given AMI tags, this will resolve the latest matching AMI and update the AMI parameter on the provided CloudFormation stack."){ (pkg, resources, target) => {
+    implicit val keyRing = resources.assembleKeyring(target, pkg)
+    val reporter = resources.reporter
+    val amiParameterMap: Map[CfnParam, TagCriteria] = getAmiParameterMap(pkg, target, reporter)
+    val cloudFormationStackLookupStrategy = getCloudFormationStackLookupStrategy(pkg, target, reporter)
 
-      List(
-        UpdateAmiCloudFormationParameterTask(
-          target.region,
-          cloudFormationStackLookupStrategy,
-          amiParameterMap,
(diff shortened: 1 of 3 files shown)


nicl (approved):
Looks good! And clearly documents the motivation. Agree, checking status of the ChangeSet execution is a good longer term approach.

akash1810:
@shtukas this should resolve the issues you were seeing 🤞🏽 .

akash1810:
This is causing new deploys to fail. We've rolled back for now.

image

PR #670, opened 2022-01-12
Attempt fix of scheduled deploys issue

nicl:
[(re https://github.com/guardian/riff-raff/pull/666](https://github.com/guardian/riff-raff/pull/666#issuecomment-1010990152)

The error seen is:

    software.amazon.awssdk.core.exception.SdkClientException Unable to marshall request to JSON: Can not write a field name, expecting a value

And it is kicked off here:

magenta.tasks.UpdateS3Lambda.$anonfun$execute$7(tasks.scala:263)

though ultimately is thrown from the AWS SDK.

It's a bit unclear why this has broken. The surrounded code has not been changed by the upgrades PR. And there are not many similar reports, but an example is:

https://github.com/aws/aws-sdk-java-v2/issues/2875

As the inputs to the call look okay (we log them in the build) for now bumping the library makes sense as a first effort at a fix.

PR #672, opened 2022-01-14
fix: Poll CloudFormation events within `ExecuteChangeSetTask`

akash1810:
## What does this change?

In https://github.com/guardian/riff-raff/pull/671 we made some changes exiting early if a ChangeSet does not yield any changes. We accomplished this by calling `DescribeChangeSet` within `CheckUpdateEventsForChangeSetTask`. This happens  _after_ `ExecuteChangeSetTask` but it is too late! It appears a ChangeSet cannot be described by _name_ once it has been executed; it can only be described using the _ARN_.

<details>
<summary>CLI examples of `DescribeChangeSet` oddities</summary>

This fails:

```console
➜ aws cloudformation describe-change-set --change-set-name deploy-1642157037866 --stack-name REDACTED

An error occurred (ChangeSetNotFound) when calling the DescribeChangeSet operation: ChangeSet [deploy-1642157037866] does not exist
```

Whereas this passes:

```console
➜ aws cloudformation describe-change-set --change-set-name arn:aws:cloudformation:eu-west-1:REDACTED:changeSet/deploy-1642157037866/72832978-90fc-4408-955f-4426d4cf2ef9
{
    "Changes": [
        {
            "Type": "Resource",
            "ResourceChange": {
                "Action": "Modify",
                "LogicalResourceId": "AutoScalingGroupAmiableASGFCD99427",
                "PhysicalResourceId": "amiable-gucdk-PROD-AutoScalingGroupAmiableASGFCD99427-SZXECQFFBHKX",
                "ResourceType": "AWS::AutoScaling::AutoScalingGroup",
                "Replacement": "Conditional",
                "Scope": [
                    "Properties"
                ],
                "Details": [
                    {
                        "Target": {
                            "Attribute": "Properties",
                            "Name": "LaunchConfigurationName",
                            "RequiresRecreation": "Conditionally"
                        },
                        "Evaluation": "Static",
                        "ChangeSource": "ResourceReference",
                        "CausingEntity": "AutoScalingGroupAmiableLaunchConfig09A6D8E3"
                    }
                ]
            }
        },
        {
            "Type": "Resource",
            "ResourceChange": {
                "Action": "Modify",
                "LogicalResourceId": "AutoScalingGroupAmiableLaunchConfig09A6D8E3",
                "PhysicalResourceId": "amiable-gucdk-PROD-AutoScalingGroupAmiableLaunchConfig09A6D8E3-EY0YNXSNHM4Y",
                "ResourceType": "AWS::AutoScaling::LaunchConfiguration",
                "Replacement": "True",
                "Scope": [
                    "Properties"
                ],
                "Details": [
                    {
                        "Target": {
                            "Attribute": "Properties",
                            "Name": "ImageId",
                            "RequiresRecreation": "Always"
                        },
                        "Evaluation": "Static",
                        "ChangeSource": "ParameterReference",
                        "CausingEntity": "AMIAmiable"
                    },
                    {
                        "Target": {
                            "Attribute": "Properties",
                            "Name": "ImageId",
                            "RequiresRecreation": "Always"
                        },
                        "Evaluation": "Dynamic",
                        "ChangeSource": "DirectModification"
                    }
                ]
            }
        }
    ],
    "ChangeSetName": "deploy-1642157037866",
    "ChangeSetId": "REDACTED",
    "StackId": "REDACTED",
    "StackName": "REDACTED",
    "Description": null,
    "CreationTime": "2022-01-14T10:43:59.447000+00:00",
    "ExecutionStatus": "EXECUTE_COMPLETE",
    "Status": "CREATE_COMPLETE",
}
```

</details>

In this change, we perform the CloudFormation event log polling _within_ `ExecuteChangeSetTask` as opposed to after it. This is necessary because we don't (yet) have a mechanism to _conditionally_ perform a `Task`. That is, it is not possible to call `CheckUpdateEventsForChangeSetTask` if and only if `ExecuteChangeSetTask` generated changes to the CloudFormation stack.

It is also not possible to pass data _between_ tasks. We only know the ChangeSet name [within the task orchestrator](https://github.com/guardian/riff-raff/blob/2c394948bc6b0241b58e220a672c8046f15e919a/magenta-lib/src/main/scala/magenta/deployment_type/CloudFormation.scala#L118-L159), therefore `DescribeChangeSet` cannot be performed within `CheckUpdateEventsForChangeSetTask` as it is after `ExecuteChangeSetTask`. The only option is to combine the two (and consequently remove `CheckUpdateEventsForChangeSetTask` as it's now unused).

By performing the event log polling within `ExecuteChangeSetTask`, we're also able to observe the execution status of the ChangeSet. Now, in addition to waiting for the event log to contain a "complete" message, we also wait until the ChangeSet has successfully completed.

## How to test

Given https://github.com/guardian/riff-raff/pull/671 caused deploys to fail (also demonstrating that https://github.com/guardian/riff-raff/pull/671 wasn't tested on CODE 😬 ), we'd expect deploys made with this branch to succeed and the event log to be printed within `ExecuteChangeSetTask`.

**From CODE**
image

As previously [noted](https://github.com/guardian/riff-raff/pull/670#issuecomment-1012950596), we could benefit from integration testing!

### DIFF
--- a/magenta-lib/src/main/scala/magenta/tasks/ChangeSetTasks.scala
+++ b/magenta-lib/src/main/scala/magenta/tasks/ChangeSetTasks.scala
@@ -140,10 +140,6 @@ class ExecuteChangeSetTask(
     response.executionStatusAsString()
   }
 
-  private def isChangeSetExecutionComplete(cfnClient: CloudFormationClient, changeSetArn: String): Boolean = {
-    getChangeSetExecutionStatus(cfnClient, changeSetArn) == "EXECUTE_COMPLETE"
-  }
-
   override def execute(resources: DeploymentResources, stopFlag: => Boolean): Unit = {
     CloudFormation.withCfnClient(keyRing, region, resources) { cfnClient =>
       val (stackName, _, _) = stackLookup.lookup(resources.reporter, cfnClient)
@@ -163,11 +159,19 @@ class ExecuteChangeSetTask(
 
         val request = ExecuteChangeSetRequest.builder().changeSetName(changeSetName).stackName(stackName).build()
         cfnClient.executeChangeSet(request)
-        
-        import magenta.tasks.StackEventPoller.check
 
-        // poll events and wait for completion
-        check(stackName, cfnClient, resources, stopFlag, None, Some(() => isChangeSetExecutionComplete(cfnClient, changeSetArn)))
+        new CloudFormationStackEventPoller(stackName, cfnClient, resources, stopFlag)
+          .check(None, () => {
+            val executionStatus = getChangeSetExecutionStatus(cfnClient, changeSetArn)
+
+            // Any other status is a failure condition, which `CloudFormationStackEventPoller` is handling at the CloudFormation event log level
+            val isComplete = executionStatus == "EXECUTE_COMPLETE"
+
+            if(!isComplete) {
+              resources.reporter.verbose(s"Execution status for change set $changeSetName on stack $stackName is $executionStatus")
+            }
+            isComplete
+          })
       }
     }
   }
--- a/magenta-lib/src/main/scala/magenta/tasks/ChangeSetTasks.scala
+++ b/magenta-lib/src/main/scala/magenta/tasks/ChangeSetTasks.scala
@@ -129,10 +129,19 @@ class CheckChangeSetCreatedTask(
   def description = s"Checking change set ${stackLookup.changeSetName} creation for stack ${stackLookup.strategy}"
 }
 
-class ExecuteChangeSetTask(
-                            region: Region,
-                            stackLookup: CloudFormationStackMetadata,
-)(implicit val keyRing: KeyRing, artifactClient: S3Client) extends Task {
+/**
+  * A task to execute a ChangeSet against a CloudFormation stack.
+  *
+  * If the ChangeSet does not contain any changes, it exits early.
+  *
+  * If there are changes, it waits until:
+  *   - The CloudFormation event log contains `UPDATE_COMPLETE` or `CREATE_COMPLETE`
+  *   - The execution status of the ChangeSet is `EXECUTE_COMPLETE`
+  *
+  * @see [[CloudFormationStackEventPoller]]
+  * @see https://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_DescribeChangeSet.html#API_DescribeChangeSet_ResponseElements
+  */
+class ExecuteChangeSetTask(region: Region, stackLookup: CloudFormationStackMetadata)(implicit val keyRing: KeyRing, artifactClient: S3Client) extends Task {
 
   private def getChangeSetExecutionStatus(cfnClient: CloudFormationClient, changeSetArn: String): String = {
     val request = DescribeChangeSetRequest.builder().changeSetName(changeSetArn).build()
--- a/magenta-lib/src/main/scala/magenta/tasks/ChangeSetTasks.scala
+++ b/magenta-lib/src/main/scala/magenta/tasks/ChangeSetTasks.scala
@@ -4,11 +4,14 @@ import magenta.artifact.S3Path
 import magenta.tasks.CloudFormationParameters.{ExistingParameter, InputParameter, TemplateParameter}
 import magenta.tasks.UpdateCloudFormationTask._
 import magenta.{ApiRoleCredentials, DeployReporter, DeploymentResources, KeyRing, Region}
+import org.joda.time.{DateTime, Duration}
+import software.amazon.awssdk.services.cloudformation.CloudFormationClient
 import software.amazon.awssdk.services.cloudformation.model.ChangeSetStatus._
-import software.amazon.awssdk.services.cloudformation.model.{Change, ChangeSetType, DeleteChangeSetRequest, DescribeChangeSetRequest, ExecuteChangeSetRequest}
+import software.amazon.awssdk.services.cloudformation.model.{Change, ChangeSetType, DeleteChangeSetRequest, DescribeChangeSetRequest, ExecuteChangeSetRequest, ListChangeSetsRequest, StackEvent}
 import software.amazon.awssdk.services.s3.S3Client
 import software.amazon.awssdk.services.sts.StsClient
 
+import scala.annotation.tailrec
 import scala.collection.JavaConverters._
 import scala.util.{Success, Try}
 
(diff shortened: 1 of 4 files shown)


magenta-lib/src/main/scala/magenta/tasks/ChangeSetTasks.scala
```diff
@@ -147,6 +168,19 @@ class ExecuteChangeSetTask(
 
         val request = ExecuteChangeSetRequest.builder().changeSetName(changeSetName).stackName(stackName).build()
         cfnClient.executeChangeSet(request)
+
+        new CloudFormationStackEventPoller(stackName, cfnClient, resources, stopFlag)
```
nicl, line 172: I'm not sure why `CloudFormationStackEventPoller` needs to be a class here. It only exposes a single method,`check`, and only that method relies on the class constructor parameters/state as far as I can tell.

In general, I would say use classes for data, and objects with methods (which is a form of namespacing really) for behaviour. There are exceptions, but typically this is simplest and best.

magenta-lib/src/main/scala/magenta/tasks/UpdateCloudFormationTask.scala
```diff
@@ -336,91 +334,66 @@ class CheckUpdateEventsTask(
...
-      if (describeResponse.changes.isEmpty) {
-        resources.reporter.info(s"No changes to perform for $changeSetName on stack $stackName")
-      } else {
-        super.execute(resources, stopFlag)
-      }
-    }
+  private def updateComplete(stackName: String)(e: StackEvent): Boolean =
+    isStackEvent(stackName)(e) && (e.resourceStatusAsString == "UPDATE_COMPLETE" || e.resourceStatusAsString == "CREATE_COMPLETE")
+
+  private def updateFailed(e: StackEvent): Boolean = {
```
nicl, line 359: Minor, but I recommend defaulting to `private[this]` unless you have a good reason to do otherwise.

nicl (approved):
Some questions, but the overal approach makes sense.

It would be great for us, in a follow-up, to migrate the `AmiCloudFormationParameter` deployment type to use change sets and remove `CheckUpdateEventsTask` altogether now.

magenta-lib/src/main/scala/magenta/tasks/ChangeSetTasks.scala
akash1810: > I'm not sure why `CloudFormationStackEventPoller` needs to be a class here.

As `check` is called recursively, with only `lastSeenEvent` varying between calls, it felt unnecessary to pass the cfn client etc each time. My Scala is quite stale atm - pairing/suggestions welcomed!

magenta-lib/src/main/scala/magenta/tasks/UpdateCloudFormationTask.scala
akash1810: Done, thanks!

magenta-lib/src/main/scala/magenta/tasks/ChangeSetTasks.scala
nicl: @akash1810 here's one way to do it (below). Note: an inner loop is used, though you could just recur here without this. A class (with constructor params) is only serving to fragment the true signature of `check` here (as none of the other class members use the constructor params). It's better, I think, to be explicit and accept that the function signature is a wide one.

```
object CloudFormationStackEventPoller extends Loggable {
  private[this] def reportEvent(reporter: DeployReporter, e: StackEvent): Unit = {
    reporter.info(s"${e.logicalResourceId} (${e.resourceType}): ${e.resourceStatusAsString}")
    if (e.resourceStatusReason != null) reporter.verbose(e.resourceStatusReason)
  }

  private[this] def isStackEvent(stackName: String)(e: StackEvent): Boolean =
    e.resourceType == "AWS::CloudFormation::Stack" && e.logicalResourceId == stackName

  private[this] def updateStart(stackName: String)(e: StackEvent): Boolean =
    isStackEvent(stackName)(e) && (e.resourceStatusAsString == "UPDATE_IN_PROGRESS" || e.resourceStatusAsString == "CREATE_IN_PROGRESS")

  private[this] def updateComplete(stackName: String)(e: StackEvent): Boolean =
    isStackEvent(stackName)(e) && (e.resourceStatusAsString == "UPDATE_COMPLETE" || e.resourceStatusAsString == "CREATE_COMPLETE")

  private[this] def updateFailed(e: StackEvent): Boolean = {
    val failed = e.resourceStatusAsString.contains("FAILED") || e.resourceStatusAsString.contains("ROLLBACK")
    logger.debug(s"${e.resourceStatusAsString} - failed = $failed")
    failed
  }

  private[this] def fail(reporter: DeployReporter, e: StackEvent): Unit = reporter.fail(
    s"""${e.logicalResourceId}(${e.resourceType}}: ${e.resourceStatusAsString}
       |${e.resourceStatusReason}""".stripMargin)

  def check(
    stackName: String,
    cfnClient: CloudFormationClient,
    resources: DeploymentResources,
    stopFlag: => Boolean,
    lastSeenEvent: Option[StackEvent],
    additionalChecks: () => Boolean = () => true): Unit = {

    def loop(lastSeenEvent: Option[StackEvent], additionalChecks: () => Boolean = () => true): Unit = {
      val result = CloudFormation.describeStackEvents(stackName, cfnClient)
      val events = result.stackEvents.asScala

      lastSeenEvent match {
        case None =>
          events.find(updateStart(stackName)) match {
            case None => resources.reporter.fail(s"No events found at all for stack $stackName")
            case Some(e) =>
              val age = new Duration(new DateTime(e.timestamp().toEpochMilli), new DateTime()).getStandardSeconds
              if (age > 30) {
                resources.reporter.verbose("No recent IN_PROGRESS events found (nothing within last 30 seconds)")
              } else {
                reportEvent(resources.reporter, e)
                check(Some(e), additionalChecks)
              }
          }
        case Some(event) =>
          val newEvents = events.takeWhile(_.timestamp.isAfter(event.timestamp))
          newEvents.reverse.foreach(reportEvent(resources.reporter, _))
          newEvents.filter(updateFailed).foreach(fail(resources.reporter, _))

          val complete = newEvents.exists(updateComplete(stackName)) && additionalChecks()
          if (!complete && !stopFlag) {
            Thread.sleep(5000)
            loop(Some(newEvents.headOption.getOrElse(event)), additionalChecks)
          }
      }

      loop(lastSeenEvent, additionalChecks)
    }
  }
}
```

akash1810: Yeah, that is nicer! I've even started a [refactor](https://github.com/guardian/riff-raff/compare/aa-refactor) that looks almost identical!

> In general, I would say use classes for data, and objects with methods (which is a form of namespacing really) for behaviour. There are exceptions, but typically this is simplest and best.

💯

PR #677, opened 2022-02-01
Remove /deployinfo/about and /deployinfo/hosts

jacobwinch:
## What does this change?

This PR deletes `/deployinfo/about` and `/deployinfo/hosts` (following on from a temporary change which hid the latter from users - https://github.com/guardian/riff-raff/pull/674). 

We noticed that a bug with the sorting logic used by `/deployinfo/hosts` caused Riff-Raff to crash, and as we're no longer convinced that these pages are useful we've decided to remove this functionality entirely rather than spending more time investigating the underlying issue.

## How to test

I've deployed this to `CODE` to confirm that the UI changes worked as expected.

There is less code to maintain!

Yes, there is a small risk that some colleagues still find these parts of the Riff-Raff UI useful.


akash1810 (approved):

akash1810:
Worth mentioning that the Prism CLI can be used as an alternative: `prism PROD -f instanceName` for all the PROD instances.

PR #681, opened 2022-03-03
feat: Introduce `templateStagePaths`

akash1810:
## What does this change?

Following https://github.com/guardian/cdk/pull/1123, and inspired by https://github.com/guardian/riff-raff/pull/237, introduce a new way to define a cloudformation template file. More detail/context can be seen in the [ADR/proposal doc](https://github.com/guardian/cdk/blob/main/docs/architecture-decision-records/006-stage-awareness.md).

This will allow us to:
  - produce multiple files (one per stage) with GuCDK
  - simplify the GuCDK codebase
  - enable GuCDK to support any stage name (previously it only supported CODE, PROD, INFRA)

This is a non-breaking change as, when `templateStagePaths` is not defined, we fallback to `templatePath`.

## How to test

- Deploy to CODE
- Deploy a build of https://github.com/guardian/cdk-playground/pull/173

This unblocks https://github.com/guardian/cdk/pull/1134. More detail can be seen in the [ADR/proposal doc](https://github.com/guardian/cdk/blob/main/docs/architecture-decision-records/006-stage-awareness.md).

There is a gap in this feature - if `riff-raff.yaml` does not list the stage being deployed to, Riff-Raff will fail to deploy as the file cannot be found. More correctly, we'll fallback to `templatePath` (which is defaulted to `cloud-formation/cfn.json`) and we'll witness this:

image

This isn't disastrous, however a better version (for a different PR) could be for `riff-raff.yaml` to list the stages an app can be deployed to:

```yaml
stacks:
  - deploy
regions:
  - eu-west-1
 stages: # <-- new
  - INFRA
deployments:
  cloudformation:
    type: cloud-formation
    app: cdk-playground
    parameters:
      templateStagePaths: 
        INFRA: CdkPlayground.template.json
``` 

This has positive consequences:
  - The stage drop-down can be populated from here, rather than Prism, meaning less risk of deploying to an incorrect stage
  - `templateStagePaths` can be validated against `stages`

### DIFF
--- a/magenta-lib/src/main/scala/magenta/deployment_type/CloudFormation.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/CloudFormation.scala
@@ -1,5 +1,6 @@
 package magenta.deployment_type
 
+import magenta.Loggable
 import magenta.Strategy.{Dangerous, MostlyHarmless}
 import magenta.artifact.S3Path
 import magenta.deployment_type.CloudFormationDeploymentTypeParameters._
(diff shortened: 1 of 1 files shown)


nicl (approved):
Nice! Looking forward to testing this out :)

PR #686, opened 2022-04-22
Correct documentation for CFN deployment type

jacobwinch:
## What does this change?

image

The description for `manageStackPolicy` states that the default is `true` but in the `Default` column it is incorrectly listed as having `<no default>`. This PR should fix the docs.

### DIFF
--- a/magenta-lib/src/main/scala/magenta/deployment_type/CloudFormation.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/CloudFormation.scala
@@ -104,7 +104,7 @@ object CloudFormation extends DeploymentType with CloudFormationDeploymentTypePa
       |${StackPolicy.toMarkdown(AllowAllPolicy)}
       |""".stripMargin,
     optional = true
-  )
+  ).default(manageStackPolicyDefault)
 
   val updateStack = Action("updateStack",
     """


akash1810 (approved):
Worth not making this optional too? `Option[Boolean]` is always a funny one, as `None` is usually treated as `True`, so `Option` is a little redundant.

jacobwinch:
> Worth not making this optional too?

It looks like [the default value for a param is always wrapped in an `Option`](https://github.com/guardian/riff-raff/blob/f46d469fd22500b1153eabd25e78301ff88a2966/magenta-lib/src/main/scala/magenta/deployment_type/Param.scala#L29).

The `optional` flag https://github.com/guardian/riff-raff/blob/f46d469fd22500b1153eabd25e78301ff88a2966/magenta-lib/src/main/scala/magenta/deployment_type/Param.scala#L16-L18 is used for something else related to validation that I don't quite understand https://github.com/guardian/riff-raff/blob/f46d469fd22500b1153eabd25e78301ff88a2966/magenta-lib/src/main/scala/magenta/input/resolver/DeploymentTypeResolver.scala#L44-L45 so I'm inclined to leave it for now as I just want to fix the docs without risking modifying the behaviour 😅

PR #688, opened 2022-04-28
Favourite stage button

tomrf1:
On the deploy page I'd like to always have CODE pre-selected for the stage. But perhaps not everyone does, so I've added a button for toggling a "favourite" stage.

### Stage is not favourite (empty star) -
Screen Shot 2022-04-28 at 09 11 41

### Stage is favourite (filled star) -
Screen Shot 2022-04-28 at 09 11 36

### Visiting the deploy page without a favourite stage still means no default value -
Screen Shot 2022-04-28 at 09 10 37


nicl:
Submitting https://github.com/guardian/riff-raff/pull/692 as an alternative approach following chat discussions.

PR #691, opened 2022-04-29
feat: Add `/requestInstanceRotation` endpoint

akash1810:
## What does this change?

There is a requirement to keep the age of our AMIs down. We achieve this in (most) apps via the [ami-cloudformation-parameter](https://github.com/guardian/riff-raff/blob/13fdb406e8ea032f0a3843682abc7be4e84e5aa8/magenta-lib/src/main/scala/magenta/deployment_type/AmiCloudFormationParameter.scala#L9-L15), [auto-scaling](https://github.com/guardian/riff-raff/blob/13fdb406e8ea032f0a3843682abc7be4e84e5aa8/magenta-lib/src/main/scala/magenta/deployment_type/AutoScaling.scala#L93-L104) deployment types, and scheduled deploys.

An `auto-scaling` deployment is not possible with Riff-Raff, instead it [employs](https://github.com/guardian/riff-raff/blob/main/riff-raff/riff-raff.yaml) a [self-deploy](https://github.com/guardian/riff-raff/blob/13fdb406e8ea032f0a3843682abc7be4e84e5aa8/magenta-lib/src/main/scala/magenta/deployment_type/SelfDeploy.scala#L8-L14).

Today, we are manually keeping the age of the AMI down - every so often we will send comms about a few minutes of downtime and terminate the instance. Although this works, it's a chore.

In this change a new `/requestInstanceRotation` endpoint is created (it acts similar to the `/requestShutdown` endpoint). When a request is made, Riff-Raff will:
  - Check there are no current deploys running
  - If there are, it will try again after a delay
  - If there are not:
    - A notification will be sent via [Anghammarad](https://github.com/guardian/anghammarad)
    - A request is made to terminate the instance within an ASG, and the ASG is expected to launch a new one in its place
    - The new instance will be running a [later AMI](https://github.com/guardian/riff-raff/blob/13fdb406e8ea032f0a3843682abc7be4e84e5aa8/riff-raff/riff-raff.yaml#L10-L18)

The intention here is to call this endpoint within a cronjob (in a separate PR), automating the task.

The method of using on the ASG to launch a new instance is fairly fast. Running `aws autoscaling terminate-instance-in-auto-scaling-group --no-should-decrement-desired-capacity` saw downtime of just over 2 minutes.

## How to test

- Deploy to CODE
- Call the endpoint (`ssm cmd -t riff-raff,CODE -p deployTools --cmd "curl --silent -X POST localhost:9000/requestInstanceRotation"`)
- Observe the ASG activity. You should see two events:
   1. `At <TIME> instance <INSTANCE-ID> was taken out of service in response to a user request.`
   2. `At <TIME> an instance was started in response to a difference between desired and actual capacity, increasing the capacity from 0 to 1.`

More automation, and less risk of human error.

[Continuous deployments](https://github.com/guardian/recommendations/blob/main/continuous-deployment.md) that are triggered during the (roughly 2 minute) downtime will _not_ run, meaning users will have to manually deploy their change ([PRout](https://github.com/guardian/prout) can help surface this!). This risk can be somewhat mitigated by scheduling the cronjob to run just outside of typical working hours.

A similar concern applies to scheduled deployments. Any scheduled deployment that is due to start when Riff-Raff is unavailable will _not_ be automatically started when Riff-Raff becomes available again. As this is the current status quo, would suggest this be resolved in future work/PRs.

### DIFF
--- a/riff-raff/app/AppComponents.scala
+++ b/riff-raff/app/AppComponents.scala
@@ -9,7 +9,7 @@ import controllers._
 import deployment.preview.PreviewCoordinator
 import deployment.{DeploymentEngine, Deployments}
 import housekeeping.ArtifactHousekeeping
-import lifecycle.{Lifecycle, RotateInstanceWhenInactive, ShutdownWhenInactive}
+import lifecycle.{Lifecycle, ShutdownWhenInactive, TerminateInstanceWhenInactive}
 import magenta.deployment_type._
 import magenta.tasks.AWS
 import notification.{DeployFailureNotifications, GrafanaAnnotationLogger, HooksClient}
@@ -148,7 +148,7 @@ class AppComponents(context: Context, config: Config, passwordProvider: Password
   val hooksClient = new HooksClient(datastore, hookConfigRepository, wsClient, executionContext)
 
   val shutdownWhenInactive = new ShutdownWhenInactive(deployments)
-  val rotateInstanceWhenInactive = new RotateInstanceWhenInactive(deployments, config)
+  val rotateInstanceWhenInactive = new TerminateInstanceWhenInactive(deployments, config)
 
   val lifecycleSingletons: Seq[Lifecycle] = Seq(
     ScheduledAgent,
--- a/riff-raff/app/AppComponents.scala
+++ b/riff-raff/app/AppComponents.scala
@@ -148,7 +148,7 @@ class AppComponents(context: Context, config: Config, passwordProvider: Password
   val hooksClient = new HooksClient(datastore, hookConfigRepository, wsClient, executionContext)
 
   val shutdownWhenInactive = new ShutdownWhenInactive(deployments)
-  val rotateInstanceWhenInactive = new RotateInstanceWhenInactive(deployments)
+  val rotateInstanceWhenInactive = new RotateInstanceWhenInactive(deployments, config)
 
   val lifecycleSingletons: Seq[Lifecycle] = Seq(
     ScheduledAgent,
--- a/riff-raff/app/AppComponents.scala
+++ b/riff-raff/app/AppComponents.scala
@@ -9,7 +9,7 @@ import controllers._
 import deployment.preview.PreviewCoordinator
 import deployment.{DeploymentEngine, Deployments}
 import housekeeping.ArtifactHousekeeping
-import lifecycle.ShutdownWhenInactive
+import lifecycle.{Lifecycle, RotateInstanceWhenInactive, ShutdownWhenInactive}
 import magenta.deployment_type._
 import magenta.tasks.AWS
 import notification.{DeployFailureNotifications, GrafanaAnnotationLogger, HooksClient}
@@ -148,8 +148,9 @@ class AppComponents(context: Context, config: Config, passwordProvider: Password
   val hooksClient = new HooksClient(datastore, hookConfigRepository, wsClient, executionContext)
 
   val shutdownWhenInactive = new ShutdownWhenInactive(deployments)
+  val rotateInstanceWhenInactive = new RotateInstanceWhenInactive(deployments)
 
-  val lifecycleSingletons = Seq(
+  val lifecycleSingletons: Seq[Lifecycle] = Seq(
     ScheduledAgent,
     deployments,
     builds,
@@ -161,6 +162,7 @@ class AppComponents(context: Context, config: Config, passwordProvider: Password
     artifactHousekeeper,
     scheduledDeployNotifier,
     shutdownWhenInactive,
+    rotateInstanceWhenInactive
   )
 
   log.info(s"Calling init() on Lifecycle singletons: ${lifecycleSingletons.map(_.getClass.getName).mkString(", ")}")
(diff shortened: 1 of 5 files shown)


riff-raff/app/AppComponents.scala
```diff
@@ -148,8 +148,9 @@ class AppComponents(context: Context, config: Config, passwordProvider: Password
   val hooksClient = new HooksClient(datastore, hookConfigRepository, wsClient, executionContext)
 
   val shutdownWhenInactive = new ShutdownWhenInactive(deployments)
+  val rotateInstanceWhenInactive = new RotateInstanceWhenInactive(deployments, config)
```
jfsoul, line 151: A real nitpick, but we're not performing a rotation directly here - that's a side effect performed by the ASG. Not important but you could go with `ShutdownAppWhenInactive` and `TerminateInstanceWhenInactive` as an alternative.

riff-raff/app/conf/context.scala
```diff
@@ -219,6 +219,26 @@ class Config(configuration: TypesafeConfig, startTime: DateTime) extends Logging
     }
   }
 
+  object management {
+    object aws {
+      // TODO read config from more generic location
+      lazy val regionName: String = getStringOpt("scheduledDeployment.aws.region").getOrElse("eu-west-1")
+
+      // For compatibility reasons this need to use the old AWS SDK
+      lazy val snsClient: AmazonSNSAsync = AmazonSNSAsyncClientBuilderV1.standard()
```
jfsoul, line 228: I'm assuming it would be a fair amount of refactoring to use v2 here instead?

riff-raff/app/conf/context.scala
```diff
@@ -219,6 +219,26 @@ class Config(configuration: TypesafeConfig, startTime: DateTime) extends Logging
...
+        .withCredentials(credentialsProviderChainV1(None, None))
+        .withRegion(regionName)
+        .build()
+
+      // TODO read config from more generic location
+      lazy val anghammaradTopicARN: String = getString("scheduledDeployment.anghammaradTopicARN")
+
+      lazy val ec2Client: Ec2Client = Ec2Client.builder()
+        .credentialsProvider(credentialsProviderChain(None, None))
```
jfsoul, line 237: I notice you're not using the v1 chain - is that correct?

jfsoul: Minor, but shall we leave off the `(None, None)` since those are the defaults anyway

riff-raff/app/lifecycle/ShutdownWhenInactive.scala
```diff
@@ -1,43 +1,91 @@
...
+      sourceSystem = "riff-raff",
+      target = List(Stack("deploy"), App("riff-raff"), Stage(config.stage)),
+      actions = List.empty,
+      channel = All,
+      topicArn = config.management.aws.anghammaradTopicARN,
+      client = config.management.aws.snsClient
+    ).recover { case ex => log.error(s"Failed to send notification (via Anghammarad)", ex) }
+
+    val request = TerminateInstancesRequest.builder().instanceIds(instanceId).build()
+    config.management.aws.ec2Client.terminateInstances(request)
```
jfsoul, line 89: Is there not a service layer that we can call here instead of directly interacting with the AWS EC2 client? There's already an EC2 object in `AWS.scala` for example, which could work.

riff-raff/app/lifecycle/ShutdownWhenInactive.scala
```diff
@@ -1,43 +1,91 @@
...
     }
   }
 
-  def attemptShutdown(): Unit = {
+  def attemptAction(): Unit = {
     Future {
-      log.info("Attempting to shutdown: trying to atomically disable deployment")
+      log.info(s"Attempting to $name: trying to atomically disable deployment")
       if (deployments.atomicDisableDeploys) {
         log.info("Deployment disabled, shutting down JVM")
```
jfsoul, line 34: This log message might be misleading for a rotation/termination action?

riff-raff/app/conf/context.scala
akash1810: Good question. I just copy pasted this from [above](https://github.com/guardian/riff-raff/blob/13fdb406e8ea032f0a3843682abc7be4e84e5aa8/riff-raff/app/conf/context.scala#L95-L104). It looks like Anghammarad wants a V1 client, so yes, sadly a fair amount of refactoring would be needed 😞 .

riff-raff/app/conf/context.scala
akash1810: The v2 client is used here, on the assumption that's the direction of travel we're going in and v1 should only be used for compat reasons. `credentialsProviderChain` does not default it's arguments, so `None, None` is still needed.

Would suggest we (separately) refactor to improve naming - `credentialsProviderChainV1` becomes `legacyCredentialsProviderChain`, for example.

riff-raff/app/lifecycle/ShutdownWhenInactive.scala
akash1810: Hadn't thought of that. Upon looking, I'm not sure it provides much value, as there's [no code](https://github.com/guardian/riff-raff/blob/13fdb406e8ea032f0a3843682abc7be4e84e5aa8/magenta-lib/src/main/scala/magenta/tasks/AWS.scala#L323-L352) that can be (easily) reused. `AWS.scala` looks to be aimed towards orchestrating a deployment's steps, rather than Riff-Raff's internals.

riff-raff/app/lifecycle/ShutdownWhenInactive.scala
akash1810: Good point! Will refactor!

riff-raff/app/conf/context.scala
```diff
@@ -219,6 +219,26 @@ class Config(configuration: TypesafeConfig, startTime: DateTime) extends Logging
     }
   }
 
+  object management {
+    object aws {
+      // TODO read config from more generic location
```
jfsoul, line 224: Are these for later? Agree that these aren't perhaps the best place to grab them from long term.

jfsoul (approved):

riff-raff/app/conf/context.scala
akash1810: Yes, will re-arrange config in a follow-up PR. Should have made that obvious, apologies.

riff-raff/app/conf/context.scala
akash1810: https://github.com/guardian/riff-raff/pull/696

riff-raff/app/conf/context.scala
akash1810: https://github.com/guardian/riff-raff/pull/697

PR #692, opened 2022-04-29
Add allowedStages

nicl:
e.g.

```
stacks:
- deploy
regions:
- eu-west-1
allowedStages:
- INFRA
deployments:
...
```

The new field is optional and takes a list of (string) stages.

If present, these stages are used in the UI dropdown when doing a manual deploy. New default behaviour for stage is also added:  first sorting the list, CODE (if allowed) is preferred, then the first non-PROD stage, and failing that the first stage is used.

Deploys (of any kind) fail if allowedStages is non-empty and does not contain the stage being deployed to.

Stage is now disabled initially:
Screenshot 2022-04-29 at 15 25 25

Selecting a build (in this case with a riff-raff.yaml containing `allowedStages`:
Screenshot 2022-04-29 at 15 25 44

Stages are limited to those specified as allowed in the riff-raff.yaml file and 'CODE' is defaulted:
Screenshot 2022-04-29 at 19 03 09

If you attempt a deploy to a disallowed stage, either through hacking the UI or via a misconfigured scheduled/continuous deploy (which are still possible), the deploy fails with a descriptive error message:

Screenshot 2022-05-13 at 15 07 25

### DIFF
--- a/riff-raff/app/assets/javascripts/form-autocomplete.coffee
+++ b/riff-raff/app/assets/javascripts/form-autocomplete.coffee
@@ -16,9 +16,9 @@ updateStageInfo = () ->
     url: url,
     dataType: 'html',
     success: (stageOptions) ->
-      stageInput = document.getElementById('stage')
-      stageInput.innerHTML = stageOptions
-      stageInput.disabled = false
+      stageInput = $('#stage')
+      stageInput.html(stageOptions)
+      stageInput.prop('disabled', false)
   });
 
 updateDeployInfo = () ->
--- a/riff-raff/app/assets/javascripts/form-autocomplete.coffee
+++ b/riff-raff/app/assets/javascripts/form-autocomplete.coffee
@@ -135,8 +135,10 @@ $ ->
       close: (event,ui) ->
         menuOpen = false
         updateBuildInfo( input.val() )
+        updateStageInfo()
       select: (event,ui) ->
         updateBuildInfo( input.val() )
+        updateStageInfo()
       minLength:0
 
   $('#buildInput').on('input keyup',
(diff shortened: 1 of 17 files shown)


nicl:
cc @tomrf1 as you might be interested.

tomrf1 (approved):
great! Very happy with this solution

riff-raff/app/assets/javascripts/form-autocomplete.coffee
```diff
@@ -4,6 +4,23 @@ menuOpen = false
...
+  selectedProject = elemProjectInput.val()
+  selectedBuild = $('#buildInput').val()
+
+  url = jsRoutes.controllers.DeployController.allowedStages(selectedProject, selectedBuild).url
+
+  $.ajax({
+    url: url,
+    dataType: 'html',
+    success: (stageOptions) ->
+      stageInput = document.getElementById('stage')
```
akash1810, line 19: minor: It looks like we use jquery across the repo? Should we use that here too for consistency? (albeit not very "modern").

riff-raff/app/views/deploy/form.scala.html
```diff
@@ -75,9 +75,11 @@
                 helper.options(prismLookup.stages.toList),
                 Symbol("_default") -> "--- Choose a stage ---",
                 Symbol("_label") -> "Stage",
+                Symbol("_help") -> "Nb. you can now constrain available stages by setting 'allowedStages' in your riff-raff.yaml.",
```
akash1810, line 78: I think we should move to making this a preferred setting, and as such include it as a deployment warning (for example, the [naming of the default branch](https://github.com/guardian/riff-raff/pull/599)).

This might be for follow-up PR though after we test it out on some projects.

akash1810 (approved):
LGTM!

riff-raff/app/views/deploy/form.scala.html
nicl: Great shout and yep, agree it is sensible to do as a follow-on.

riff-raff/app/assets/javascripts/form-autocomplete.coffee
nicl: 👍 this makes me sad 😢 but perhaps I should be more grateful for jquery in my life.

PR #693, opened 2022-05-03
feat: Perform instance rotation on a schedule

akash1810:
## What does this change?

Follows #691, and combined with the current update AMI deployment type, this will ensure Riff-Raff is using a recent AMI.

Schedule is:
  - CODE every Wednesday at 8AM
  - PROD every Thursday at 8AM

### DIFF
--- a/riff-raff/bootstrap.sh
+++ b/riff-raff/bootstrap.sh
@@ -61,6 +61,11 @@ APP_BUCKET="deploy-tools-dist"
 USER=$APP
 HOME="/home/$APP"
 
+INSTANCE_ROTATION_SCHEDULE="WED *-*-* 08:00:00 Europe/London"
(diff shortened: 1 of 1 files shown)


riff-raff/bootstrap.sh
```diff
@@ -61,6 +61,11 @@ APP_BUCKET="deploy-tools-dist"
 USER=$APP
 HOME="/home/$APP"
 
+INSTANCE_ROTATION_SCHEDULE="WEB *-*-* 08:00:00 Europe/London"
```
jfsoul, line 64:
```suggestion
INSTANCE_ROTATION_SCHEDULE="WED *-*-* 08:00:00 Europe/London"
```

jfsoul (approved):

PR #696, opened 2022-05-03
refactor: Rename `credentialsProviderChainV1` to `legacyCredentialsProviderChain`

akash1810:
## What does this change?

We're moving away from V1 of the AWS SDK. In this change we rename `credentialsProviderChainV1` and add some documentation to that effect.

Follows from https://github.com/guardian/riff-raff/pull/691#discussion_r863574962.

### DIFF
--- a/riff-raff/app/conf/context.scala
+++ b/riff-raff/app/conf/context.scala
@@ -97,7 +97,7 @@ class Config(configuration: TypesafeConfig, startTime: DateTime) extends Logging
     lazy val regionName = getStringOpt("scheduledDeployment.aws.region").getOrElse("eu-west-1")
     // For compatibility reasons this need to use the old AWS SDK
     lazy val snsClient = AmazonSNSAsyncClientBuilderV1.standard()
-      .withCredentials(credentialsProviderChainV1(None, None))
+      .withCredentials(legacyCredentialsProviderChain(None, None))
       .withRegion(regionName)
       .build()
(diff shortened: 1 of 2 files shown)


riff-raff/app/conf/context.scala
```diff
@@ -240,7 +240,13 @@ class Config(configuration: TypesafeConfig, startTime: DateTime) extends Logging
     }
   }
 
-  def credentialsProviderChainV1(accessKey: Option[String] = None, secretKey: Option[String] = None): AWSCredentialsProviderChainV1 = {
+  /**
+    * Credentials for use with V1 of the AWS SDK.
+    * Considered legacy as, where possible, V2 of the AWS SDK should be used.
+    *
+    * @deprecated Use V2 credentials - see [[credentialsProviderChain]].
```
jacobwinch, line 247: Minor: the comment here is useful but I'm not convinced that a deprecation warning on the credentials provider is the best way to communicate that we're moving away from v1. IMO it will likely lead to noise when compiling. We might be better off with a Trello card for the upgrade?

jacobwinch (approved):

riff-raff/app/conf/context.scala
akash1810: Hadn't considered this - looking at the build log for CI, there are _lots_ of deprecation warnings, and I agree this will easily be lost. Removed.

PR #740, opened 2022-07-28
Delete obsolate files in Google Cloud Platform deploys

NathanielBennett:
## What does this change?

The datatech team would like to refine the GCS deployment type previously implemented here: https://github.com/guardian/riff-raff/pull/604 and here: https://github.com/guardian/riff-raff/pull/574. We use this deployment type to update our CODE and PROD instances of [Cloud Composer](https://www.google.com/url?sa=t&rct=j&q=&esrc=s&source=web&cd=&cad=rja&uact=8&ved=2ahUKEwjvta_i3Zv5AhXZgVwKHWgAC8cQFnoECAcQAQ&url=https%3A%2F%2Fcloud.google.com%2Fcomposer&usg=AOvVaw0BvoAUnha-00_IH8bqYzJs). This is the Google Cloud hosted implementation of [Apache Airflow](https://airflow.apache.org/]. 

For each deploy we upload a set of (mostly) python files to a GCS bucket. These files contain the [Dags](https://airflow.apache.org/docs/apache-airflow/stable/concepts/dags.html) defining the data-pipelines in the datalake, along with library code that supports them ( for example, modules to run [DBT](https://www.getdbt.com/) models, or create [Dataproc](https://cloud.google.com/dataproc) clusters and submit jobs to them, for example). These are managed as part of the [ophan datalake](https://github.com/guardian/ophan-data-lake/tree/main/airflow) repo in the `dags` and `guardian` directories. 

What we'd like is to alter the GCP dployment type so that when we remove dags ( or some other type of asset ) from our cloud composer code it can, optionally, be removed from the deploy bucket. This is what this does. In order to make this more generic, change the GCP configuration type to take a list of sub-folders and a list of file types and  remove any matching files which are no longer needed. This should be an optional configuration: for CODE, we'll want to retain the current functionality. 

I've added test cases for the new functionality and refactored the GCS test cases to extract common code and make each test a little more  readable.

I'll squash the commits on merge with all all encompassing commit message

## How to test

I deployed this branch to RiffRaff CODE and then used this Airflow branch https://github.com/guardian/ophan-data-lake/pull/6890 to check everything worked. You can replicate this by first deploying build #7373 to airflow CODE and then #7375. The first creates a couple small dag(.py) files in `dags/dags` and then the second deletes them. The deletions are `gcs://europe-west2-datatech-airfl-fdf1e4b5-bucket/dags/dags/commercial_dfp_impressions_backfill_test.py` and  gcs://europe-west2-datatech-airfl-fdf1e4b5-bucket/dags/dags/commercial_dfp_line_item_test.py `. The logs show these being deleted under the `uploadStaticFiles` deployment, task: `GCSUpload`.  Note: We'll only really want to do this for PROD, but in the above airflow builds the changes to `riffraff.yaml` have the configuration altered for CODE for testing purposes

See above:

You'll probably need access to the: [Code Composer](https://console.cloud.google.com/storage/browser/europe-west2-datatech-airfl-fdf1e4b5-bucket/dags;tab=objects?project=datatech-platform-code&prefix=&forceOnObjectsSortingFiltering=false) Dags  folder(ask datatech).

Deploy this branch to riffaff code and then with that deploy **Ophan datalake PR to come** to `ophan:datalake:airflow` to CODE
confirm that there are no errors and that **Details to Come** files have been removed from the deploy bucket

There should be no increase in errors for deploys to  `ophan:datalake:airflow` 

## Accessibility

### DIFF
--- a/magenta-lib/src/main/scala/magenta/deployment_type/GCS.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/GCS.scala
@@ -193,11 +193,11 @@ object GcsTargetBucket {
             maybeDirectories: Option[Map[String, List[String]]],
             maybeFileTypes: Option[Map[String, List[String]]]): GcsTargetBucket = {
 
-      def listOrEmpty(maybeMap: Option[Map[String, List[String]]]) =
+      def listOrEmpty(m: Option[Map[String, List[String]]]) =
         (for{
-          map <- maybeMap
-          value <- map.get(stage)
-        } yield value).getOrElse(List.empty)
+          infoByStage <- m
+          info <- infoByStage.get(stage)
+        } yield info).getOrElse(List.empty)
 
 
       val directoriesToPurge = listOrEmpty(maybeDirectories).map( dir => if (prefix.isEmpty) dir else s"$prefix/$dir" )
--- a/magenta-lib/src/main/scala/magenta/deployment_type/GCS.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/GCS.scala
@@ -61,18 +61,18 @@ object GCS extends DeploymentType {
   )
 
   val fileTypesToPruneByStage: Param[Map[String, List[String]]]  = Param[Map[String, List[String]]](
-    name = "directoriesToPruneByStage",
+    name = "fileTypesToPruneByStage",
     documentation =
       """
         |Specify the types of file to remove when they are no longer in the current upload. Use case: Remove obselete dags from an airflow to cloud composer deployment
         |```
-        |directoriesToPruneByStage:
+        |fileTypesToPruneByStage:
         | PROD: [py, json]
         | CODE: [py, json]
         |```
         |Or more likely in this case:
         |```
-        |directoriesToPruneByStage:
+        |fileTypesToPruneByStage:
         | PROD: [py, json]
         |```
         |
--- a/magenta-lib/src/main/scala/magenta/deployment_type/GCS.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/GCS.scala
@@ -24,33 +24,66 @@ object GCS extends DeploymentType {
 
   //required configuration, you cannot upload without setting these
   val bucket = Param[String]("bucket", "GCS bucket to upload package files to (see also `bucketResource`)", optional = true)
-  val bucketByStage: Param[Map[String, Map[String, List[String]]]] = Param[Map[String, Map[String, List[String]]]](
+  val bucketByStage: Param[Map[String, String]] = Param[Map[String, String]](
     name = "bucketByStage",
     documentation =
       """
-        |A dict of stages to buckets in the package along with details of how When the current stage is found in here the bucket
-        |will be used from this dict. If it's not found here then it will fall back to `bucket` if it exists. If the name is defined here
-        | some other details can be included
-        |   -
-        |   - directoriesToPrune: A list of directories within the bucket. In each of these, recursively search for any files that aren't present in the current upload and remove them
+        |A dict of stages to buckets in the package. When the current stage is found in here the bucket
+        |will be used from this dict. If it's not found here then it will fall back to `bucket` if it exists.
         |```
         |bucketByStage:
-        |  PROD:
-        |    name: [prod-bucket]
-        |    dirw
-        |
-        |
+        |  PROD: prod-bucket
         |  CODE: code-bucket
         |```
         |""".stripMargin,
     optional = true
   )
 
+  val directoriesToPruneByStage: Param[Map[String, List[String]]]  = Param[Map[String, List[String]]](
+    name = "directoriesToPruneByStage",
+    documentation =
+      """
+        |A list of directories in the target buckets which will be pruned of any files not in the current upload for the current stage
+        |Typically this is used to remove obsolete dags and associated python assets from a airflow deploy to cloud composer
+        |```
+        |directoriesToPruneByStage:
+        | PROD: [dags/dags, dags/guardian]
+        | CODE: [dags/dags, dags/guardian]
+        |```
+        |Or more likely in this case:
+        |```
+        |directoriesToPruneByStage:
+        | PROD: [dags/dags, dags/guardian]
+        | CODE: [dags/dags, dags/guardian]
+        |```
+        |
+         |""".stripMargin,
+    optional = true
+  )
+
+  val fileTypesToPruneByStage: Param[Map[String, List[String]]]  = Param[Map[String, List[String]]](
+    name = "directoriesToPruneByStage",
+    documentation =
+      """
+        |Specify the types of file to remove when they are no longer in the current upload. Use case: Remove obselete dags from an airflow to cloud composer deployment
+        |```
+        |directoriesToPruneByStage:
+        | PROD: [py, json]
+        | CODE: [py, json]
+        |```
+        |Or more likely in this case:
+        |```
+        |directoriesToPruneByStage:
+        | PROD: [py, json]
+        |```
+        |
+         |""".stripMargin,
+    optional = true
+  )
+
   val publicReadAcl = Param[Boolean]("publicReadAcl",
     "Whether the uploaded artifacts should be given the PublicRead Canned ACL. (Default is true!)"
   ).default(true)
-
-
 
   val cacheControl = Param[List[PatternValue]]("cacheControl",
     """
@@ -86,17 +119,6 @@ object GCS extends DeploymentType {
       |    ]
     """.stripMargin
   )
-
-  val removeObseleteFiles = Param(
-    name = "removeObseleteFiles",
-    documentation =
-      """
-        | This is primarily the data-tech's airflow deployment to cloud-composer.
-        | When DAGs ( or other python library files ) are removed frou our codebase, they remain in the deploy bucket
-        | Set to true will remove these unwanted files from GCS on each deploy
-        | Defaults to false
-        |""".stripMargin
-  ).default(false)
 
   val uploadStaticFiles = Action(
     name = "uploadStaticFiles",
@@ -138,7 +160,6 @@ object GCS extends DeploymentType {
       case (None, None) | (Some(_), Some(_)) => reporter.fail(s"One and only one of bucketByStage or bucket must be provided")
     }
 
-
     val maybeDatum = resourceLookupFor(pathPrefixResource)
     val maybeString = maybeDatum.map(_.value)
     val prefix: String = maybeString.getOrElse(GCSUpload.prefixGenerator(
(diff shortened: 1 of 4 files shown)


magenta-lib/src/main/scala/magenta/deployment_type/GCS.scala
```diff
@@ -125,16 +167,43 @@ object GCS extends DeploymentType {
...
+            stage: String,
+            prefix: String,
+            maybeDirectories: Option[Map[String, List[String]]],
+            maybeFileTypes: Option[Map[String, List[String]]]): GcsTargetBucket = {
+
+      def listOrEmpty(maybeMap: Option[Map[String, List[String]]]) =
+        (for{
+          map <- maybeMap
+          value <- map.get(stage)
+        } yield value).getOrElse(List.empty)
```
franksloan, line 200: Should this be private for now?

magenta-lib/src/main/scala/magenta/tasks/gcp/GCSTasks.scala
```diff
@@ -49,7 +53,7 @@ case class GCSUpload(
   def fileString(quantity: Int) = s"$quantity file${if (quantity != 1) "s" else ""}"
 
   // end-user friendly description of this task
-  def description: String = s"Upload ${fileString(objectMappings.size)} to GCS bucket $bucket using file mapping $paths"
+  def description: String = s"Upload ${fileString(objectMappings.size)} to GCS bucket $gcsTargetBucket using file mapping $paths"
```
franksloan, line 56: Should this include the .name

magenta-lib/src/main/scala/magenta/tasks/gcp/GCSTasks.scala
```diff
@@ -102,6 +115,75 @@ case class GCSUpload(
...
+    def allObjectsInMatchingDirectories(directoriesToPurge: List[String], itemsSoFar: List[StorageObject] = List.empty): List[StorageObject] =
+        directoriesToPurge match {
+          case Nil => itemsSoFar
+          case head :: tail =>
+            val gcsQuery = storage.objects()
+             .list(gcsTargetBucket.name)
+             .setPrefix(head)
+            val allItemsForThisDirectory = getAllObjectsForDirectory(gcsQuery, itemsSoFar)
+            allObjectsInMatchingDirectories(tail, allItemsForThisDirectory )
+        }
```
franksloan, line 144: How does this work if it is a directory (rather than a file) in the storage object? Do we need to cover that case..? maybe not

franksloan (commented):
This looks great, awesome work! I have a few minor questions

magenta-lib/src/main/scala/magenta/tasks/gcp/GCSTasks.scala
NathanielBennett: It's a case class so it has a toString method which enables the whole bucket configuration to appear in the riffraff deploy output: Which is useful. I might make this more explicit:
But for example: `[10:21:48] task GCSUpload Upload 163 files to GCS bucket GcsTargetBucket(europe-west2-datatech-airfl-fdf1e4b5-bucket,List(dags/dags, dags/guardian),List(.py, .json)) using file mapping List((S3Path(riffraff-artifact,ophan:data-lake:airflow/7375/cloud-composer-deployment/),))`

magenta-lib/src/main/scala/magenta/tasks/gcp/GCSTasks.scala
NathanielBennett: Yeah the v1 google api sucks a bit. This does what we want. If the directory is `dags/dags`(Like [Here](https://console.cloud.google.com/storage/browser/europe-west2-datatech-airfl-fdf1e4b5-bucket/dags/dags?project=datatech-platform-code&pageState=(%22StorageObjectListTable%22:(%22f%22:%22%255B%255D%22))&prefix=&forceOnObjectsSortingFiltering=false)), this gives all objects, recursively under the folder in head. I don't think folders are considered as storage options. You can sae this by running this code by itself 

```
object GcpBucketReaderTwo extends App{

  lazy val httpTransport: ApacheHttpTransport = GoogleApacheHttpTransport.newTrustedTransport
  //lazy val jsonFactory = JacksonFactory.getDefaultInstance()
  lazy val jsonFactory = JacksonFactory.getDefaultInstance()
  private val gcsTargetBucket = "europe-west2-datatech-airfl-fdf1e4b5-bucket"


  authorize.map{
    credentials =>
      println("Got credentials")
      val storage = new Storage.Builder(httpTransport, jsonFactory, credentials ).build()

     val results = allObjectsInMatchingDirectories(storage, List("dags/dags", "dags/gssuardian") )
     results.foreach(s => println(s"**${s.getName}**"))
  }

  def allObjectsInMatchingDirectories(storage: Storage, directoriesToPurge: List[String], itemsSoFar: List[StorageObject] = List.empty): List[StorageObject] =
    directoriesToPurge match {
      case Nil => itemsSoFar
      case head :: tail =>
        val gcsQuery = storage.objects()
          .list(gcsTargetBucket)
          .setPrefix(head)
        val allItemsForThisDirectory = getAllObjectsForDirectory(gcsQuery, itemsSoFar)
        allObjectsInMatchingDirectories(storage, tail, allItemsForThisDirectory )
    }



  @tailrec
  def getAllObjectsForDirectory(query: Storage#Objects#List, foundSoFar: List[StorageObject] = List.empty): List[StorageObject] = {

    println(s"Excecusted: ${foundSoFar.size}")
    val results = query.execute
    println("Excecusting")
    val currentPageItems = Option(results.getItems).map(rs => rs.asScala).getOrElse(List.empty)
    val allItemsSoFar = foundSoFar ++ currentPageItems


    val token = results.getNextPageToken
    Option(token) match {
      case None => allItemsSoFar
      case Some(nextPageToken) => getAllObjectsForDirectory(query.setPageToken(nextPageToken), allItemsSoFar)
    }
  }
```

magenta-lib/src/main/scala/magenta/deployment_type/GCS.scala
NathanielBennett: Yup good call!

NathanielBennett: No actually - because it's enclosed in the `apply` method it's already private to there

magenta-lib/src/main/scala/magenta/tasks/gcp/GCSTasks.scala
NathanielBennett: OK there was an issue here which I've fixed in the last commit. I don't think I can delete folders tho

magenta-lib/src/main/scala/magenta/deployment_type/GCS.scala
```diff
@@ -125,16 +167,43 @@ object GCS extends DeploymentType {
...
+
+object GcsTargetBucket {
+
+  def apply(name: String,
+            stage: String,
+            prefix: String,
+            maybeDirectories: Option[Map[String, List[String]]],
+            maybeFileTypes: Option[Map[String, List[String]]]): GcsTargetBucket = {
+
+      def listOrEmpty(maybeMap: Option[Map[String, List[String]]]) =
```
nicl, line 196: I wonder if this can be renamed to be a bit clearer. Something like:

```
      def extractStageInfo(m: Option[Map[String, List[String]]], stage: String) = {
        (for{
          infoByStage <- m
          value <- infoByStage.get(stage)
        } yield value).getOrElse(Nil)
      }
```

So that rather than quite general naming (e.g. 'maybeMap', 'map', 'value', 'listOrEmpty') it's a bit clearer what is going on.

magenta-lib/src/main/scala/magenta/tasks/gcp/GCSTasks.scala
```diff
@@ -79,11 +85,18 @@ case class GCSUpload(
...
           val contentType = Option(transfer.target.getContentType).getOrElse(URLConnection.guessContentTypeFromStream(inputStream))
-          val result      = client.objects().insert(bucket, transfer.target, new InputStreamContent(contentType, inputStream)).execute()
+          val result      = client.objects().insert(gcsTargetBucket.name, transfer.target, new InputStreamContent(contentType, inputStream)).execute()
           logger.debug(s"Put object ${result.getName}: MD5: ${result.getMd5Hash} Metadata: ${result.getMetadata}")
           result
         }
       }
+      currentlyDeployedObjectsToDelete.par.foreach { case storageObjectToDelete =>
+        resources.reporter.verbose(s"Deleting obsolete file from GCP: gcs://${gcsTargetBucket.name}/${storageObjectToDelete.getName}")
+        val errorMessage = s"Could not 0emove obselete object ${storageObjectToDelete.getName}"
```
nicl, line 95: Typo - `0emove` -> `remove` ?

nicl (commented):
Looks good! I don't quite understand the domain so this may not be possible, but I wonder if the deploy model suggested here could be simplified? I.e. it's quite flexible at the moment, but the result is a lot of operational complexity - e.g. looping through all files to see which ones to delete etc.

Could a different approach to how you store/load your DAGS help here?

E.g. if you want to clean after deploys, what about:

* deploying all Dags each time with resources and having RR implement S3 sync with it's `--delete` command? (see [here](https://docs.aws.amazon.com/cli/latest/reference/s3/sync.html))
* alternatively, if deploys are of individual DAGS, storing each dag and resources in a separate directory and syncing at the directory level?

That way, the RR deploy API/semantics could be a lot simpler: just add a `sync (boolean)` flag to the existing GCS deployment type.

NathanielBennett:
@nicl 
I'm advocating to deploy as is as anything else needs to have the `GCS`, `GCSTasks` and `GCP` classes almost completely re-written from scratch. 

The code is based around the  [google storage])https://github.com/googleapis/google-api-java-client-services/tree/main/clients/google-api-services-storage/v1) which doesn't have any sync functionality on the [objects](https://googleapis.dev/java/google-api-services-storage/latest/com/google/api/services/storage/model/StorageObject.html] class. It's true that [gsuti]()https://cloud.google.com/storage/docs/gsutil) does kind of give us what we want, but it's a python-based command line utility, so that would require rewriting the whole GCS depoy type. I don't even know whether this can even be done (and is a *much* more subtantial piece of work. 

Also,  We can't really feasably deploy dag-by-dag as this would would add magnitudes of complexity to our work flow we'd need a separate teamcity build/riff-raff config for each single file [here[(https://github.com/guardian/ophan-data-lake/tree/main/airflow/dags) and a way of identifying which of the [library classes](https://github.com/guardian/ophan-data-lake/tree/main/airflow/python/guardian) needed to be bundled up with each of individual `dag` file . 

This gives us what we need without a having to rewrite the whole deploy type from scratch.

nicl:
@NathanielBennett go for it! Thanks for the info/explanations.

nicl (approved):

PR #788, opened 2022-08-05
feat(autoscaling): Enable deployment to a new ASG

akash1810:
## What does this change?

As part of the pre-flight checks for an `autoscaling` deployment, Riff-Raff will attempt to lookup the ASG, and throw an error if it's unable to locate any. This behaviour is desired when deploying to an _existing_ ASG. However if we want to _create_ a new stack (from nothing) it means we cannot use Riff-Raff for the initial creation and have to resort to going directly to AWS. 😢 

In this change, we change the rules. If an ASG cannot be located _and_ the update strategy is `Dangerous` (see #626), then allow the deployment to continue. This allows us to use Riff-Raff to deploy a new stack from scratch! 🎉 

Note: If an ASG cannot be located and the update strategy is `MostlyHarmless`, we continue to fail, but provide a hint to the `Dangerous` option.

### Preferred solution
This change is more pragmatic than preferred, in that it helps solve the described problem, but not in the best way. A better version would be to improve the [dependency graph](https://github.com/guardian/riff-raff/blob/1b457280690f4d4b74fd0638bb0dedcb4c82c2d4/riff-raff/app/deployment/preview/Preview.scala#L32-L43). IIUC RIff-Raff builds a graph of steps and interdependencies, however it only uses this for ordering and not for validation checks.

An improvement could be to mark all children in the graph as passing their validation if the parent node has passed. That is, in the situation where we want to create a new stack, Riff-Raff should pass the ASG lookup if it's deemed the CloudFormation stack doesn't exist (an ASG deployment is dependant on a CloudFormation deployment).

Whilst this change does start to overload the `Dangerous` mode, it doesn't change it's meaning too much - deploying to an ASG that has not been discovered is `Dangerous` if you were expecting it to be discovered.

## How to test

- Deploy to CODE
- Deploy https://github.com/guardian/backend-101-scala-starter-akash1810 from scratch. It should be successful.

### Today
image

#### Error message with prompt
image

### When deploying with `Dangerous` mode 🎉 
image

### DIFF
--- a/magenta-lib/src/main/scala/magenta/tasks/AWS.scala
+++ b/magenta-lib/src/main/scala/magenta/tasks/AWS.scala
@@ -3,7 +3,7 @@ package magenta.tasks
 import java.nio.ByteBuffer
 import cats.implicits._
 import magenta.deployment_type.{MigrationTagRequirements, MustBePresent, MustNotBePresent}
-import magenta.{ApiRoleCredentials, ApiStaticCredentials, App, DeployReporter, DeploymentPackage, DeploymentResources, KeyRing, Loggable, Region, Stack, Stage, StsDeploymentResources, withResource}
+import magenta.{ApiRoleCredentials, ApiStaticCredentials, App, DeployReporter, DeploymentPackage, DeploymentResources, KeyRing, Loggable, Region, Stack, Stage, Strategy, StsDeploymentResources, withResource}
 import software.amazon.awssdk.auth.credentials.{AwsBasicCredentials, AwsCredentials, AwsCredentialsProvider, AwsCredentialsProviderChain, ProfileCredentialsProvider, StaticCredentialsProvider}
 import software.amazon.awssdk.core.SdkBytes
 import software.amazon.awssdk.core.client.config.ClientOverrideConfiguration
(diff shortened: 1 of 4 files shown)


magenta-lib/src/main/scala/magenta/tasks/AWS.scala
```diff
@@ -257,10 +255,12 @@ object ASG {
...
+      case Nil if strategy == Strategy.Dangerous =>
+        reporter.info(s"Unable to locate an autoscaling group AND living dangerously - we'll assume you're creating a new stack")
+        None
       case Nil =>
-        reporter.fail(s"No autoscaling group found with tags $tagRequirements")
-      case List(singleGroup) =>
-        singleGroup
+        reporter.fail(s"No autoscaling group found with tags $tagRequirements. Creating a new stack? Initially choose the ${Strategy.Dangerous} strategy.")
+      case List(singleGroup) => Some(singleGroup)
       case groupList =>
```
kenoir, line 264: Can this condition now be reached or is it covered by matching `Nil` without qualification above?

kenoir (approved):

magenta-lib/src/main/scala/magenta/tasks/AWS.scala
akash1810: I think so, at least it was before...

kenoir: Looking at this i've again i've misread the code, I see how this works now. It does seem like this function is untested though it may be cumbersome to do so given it contains an `AutoScalingClient` and the tests against this file are sparse anyway.

PR #826, opened 2022-08-31
set timeout requests to requests to gcs

NathanielBennett:
## What does this change?

**Short version**:

Following #740 we need to set a timeout on GCS storage requests to correct connect timeout errors seen deploying airflow to [PROD](https://riffraff.gutools.co.uk/deployment/history?projectName=ophan%3Adata-lake%3Aairflow&stage=PROD&pageSize=20&page=1), although [CODE](https://riffraff.gutools.co.uk/deployment/history?projectName=ophan%3Adata-lake%3Aairflow&stage=CODE&pageSize=20&page=1) is fine.

**Long version**
 
This fixes an intermittent bug we've seen in airflow deploys to PROD since deploying this #740. the request that the [GCS Storage](https://github.com/googleapis/google-api-java-client-services/tree/main/clients/google-api-services-storage/v1) library makes to list current contents of the airflow deploy bucket times out. See failed [Prod Deploys](https://riffraff.gutools.co.uk/deployment/history?stage=PROD&pageSize=20&page=1&projectName=ophan%3Adata-lake%3Aairflow). The stack-trace for each of these failures shows a Connection Timeout error occurring  [here](https://github.com/guardian/riff-raff/blob/main/magenta-lib/src/main/scala/magenta/tasks/gcp/GCSTasks.scala#L125).

To mitigate this, adds a timeout to the the request which I *think* will solve the problem . See caveats below. However, we've seen this issue with this library in one of our pipelines so I'm confident it will work (see [here](https://github.com/guardian/ophan-data-lake/blob/main/etl/commercial/src/main/scala/dfp/job/logscopier/transfer/GcpDfpLogsTransferrer.scala),  which we've resolved in the same [way](
https://github.com/guardian/ophan-data-lake/blob/main/etl/commercial/src/main/scala/dfp/job/logscopier/transfer/GcpDfpLogsTransferrer.scala#L153)

*Preamble to Problem*

 This only seems to occur when deploying from riffraff PROD to airflow PROD. I was able to recreate it once deploying airflow from RiffRaff Code to Airflow CODE ([deploy](https://riffraff.code.dev-gutools.co.uk/deployment/view/6f5b2acb-ddfe-4721-aaf1-211bf78527f5)). This branch has the configuration for deleting files implemented #740 applied to airflow CODE. Usually we will only want to do this in PROD. The error seems to occur far less frequently in CODE than with PROD (code [deploys](https://riffraff.code.dev-gutools.co.uk/deployment/history?stage=CODE&pageSize=20&page=1&projectName=ophan%3Adata-lake%3Aairflow). 

*Crux of problem* 

 if you look at the [Prod Deploys](https://riffraff.code.dev-gutools.co.uk/deployment/history?stage=PROD&pageSize=20&page=1&projectName=ophan%3Adata-lake%3Aairflow) from RiffRaff CODE, there's a long list of failed. This is a separate issue. If you checkout a stack-trace to one of these failures (for [example](https://riffraff.code.dev-gutools.co.uk/deployment/view/61afdfcb-307c-4d64-94c7-d92bc29154e5) you'll see the failure is a runtime exception from [this line(https://github.com/guardian/riff-raff/blob/main/magenta-lib/src/main/scala/magenta/tasks/gcp/GCP.scala#L163) rather than the connection timeout issue that we're seeing on deploys Riff Raff PROD to Airflow PROD( like [here](https://riffraff.gutools.co.uk/deployment/view/8dc6581b-6cec-4238-8e20-7569e3c04677).

 I *think* this is because riff raff CODE doesn't have the necessary creds [here](https://github.com/guardian/riff-raff/blob/main/magenta-lib/src/main/scala/magenta/tasks/gcp/GCP.scala#L29) to deploy airflow PROD ( each instance is in a separate  [GCP Project](https://cloud.google.com/storage/docs/project), so will have its own credentials). To confirm this I temporarily deployed a pre #740 main to riffraff code and checked that the same run-time error occurred when I tried to deploy a pre #740 airflow main branch to PROD from there and saw the [same error](https://riffraff.code.dev-gutools.co.uk/deployment/view/e9423452-34fb-4297-bd4f-68b1995f05ad). I have tested is that this doesn't break CODE deploys. But there doesn't seem to be away of testing that Riff Raff CODE can deploy Airflow to PROD

I'm at a loss as to with one GCS bucket, but not another. They're both in same region,  but the PROD bucket is larger maybe? 

## How to test

See above

Expect airflow PROD deploys to be error free

()

## Accessibility

cc @franksloan @tylr-mk


nicl (approved):
Happy for you to try this! What was the timeout before (the default)?

PR #867, opened 2022-10-03
Test out GHA builds

nicl:
**Note, I have temporarily paused the Teamcity Riffraff build as part of this. We can unpause if we decide to rework this PR for any reason.**

*Note, the `riffRaffUpload` task has been temporarily removed until this is ready to merge to avoid clashing with Teamcity which isn't aware of other builds.*

Switch to GHA.

There are a few motivations here:

* move to GHA with associated benefits (CI in code, etc.)
* see whether as a result builds become less flaky (https://trello.com/c/nBdgDu2h/1432-move-riffraff-build-to-gha has details) - we'll probably only be able to answer this over time but it seems likely given the causes of some of the previous issues concern contention for resources on our Teamcity agents
* explore Scala GHA build times and to what extent larger runners help here (will do in subsequent PR as having issues with our custom runner atm - workflows are hanging).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,4 +25,5 @@ jobs:
           # Ensure we don't overwrite existing (Teamcity) builds.
           # LAST_TEAMCITY_BUILD=TODO
           # export GITHUB_RUN_NUMBER=$(( $GITHUB_RUN_NUMBER + $LAST_TEAMCITY_BUILD ))
+          # Note, lib/test because riffraff/riffRaffUpload only tests the riffraff project.
           sbt clean compile lib/test riffraff/riffRaffUpload
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,9 +9,23 @@ on:
 
 jobs:
   test:
-    runs-on: 4core-ubuntu-latest
+    runs-on: ubuntu-latest
+
+    permissions:
+      # Allow GitHub to request an OIDC JWT ID token, for exchange with `aws-actions/configure-aws-credentials`
+      # See https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-amazon-web-services#updating-your-github-actions-workflow
+      id-token: write
+
+      # Required for `actions/checkout`
+      contents: read
+
     steps:
       - uses: actions/checkout@v3
+
+      - uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-region: eu-west-1
+          role-to-assume: ${{ secrets.GU_RIFF_RAFF_ROLE_ARN }}
 
       - name: Setup JDK
         uses: actions/setup-java@v3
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,45 @@
...
+
+      - name: build + test
+        run: |
+          set -e
+          # Ensure we don't overwrite existing (Teamcity) builds.
+          LAST_TEAMCITY_BUILD=2360
+          export GITHUB_RUN_NUMBER=$(( $GITHUB_RUN_NUMBER + $LAST_TEAMCITY_BUILD ))
+
+          # Note, lib/test because riffraff/riffRaffUpload only tests the riffraff project.
+          # TODO add riffraff/riffRaffUpload as final sbt task below once ready to go live
```
nicl, line 44: Note this. An alternative is to rename the project to avoid clashes, but that has other downsides. For a simple GHA migration like this, a cruder approach seems sufficient.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,45 @@
...
+          sbt clean compile lib/test
```
akash1810, line 45: We [currently](https://github.com/guardian/riff-raff/blob/727e9d07ba0b19e640ff93d529fb8d9339e14084/script/ci) run `sbt clean compile test`. Any reason to move to `lib/test`?

akash1810:
IIUC Riff-Raff has some tests that require Postgres. SBT is [orchestrating](https://github.com/guardian/riff-raff/blob/727e9d07ba0b19e640ff93d529fb8d9339e14084/build.sbt#L97-L104) this at the moment. Not a blocker, but we could consider moving this directly to GHA, e.g. https://github.com/guardian/grid/blob/f61c97f2e1761a33d0ba3f79d59f854b885517cb/.github/workflows/ci.yml#L35-L67

akash1810 (approved):
One comment on the tests being run, else LGTM! 👍🏽

.github/workflows/ci.yml
nicl: @akash1810 see the comment on L43 (a bit obscured by the temporary comment above admittedly). It's an unfortunate effect of the way this project is setup. I wonder if the multi-project lib/riffraff divide here is really serving any purpose now but wasn't keen on modifying that here.

nicl:
@akash1810 thanks. Re Postgres, an advantage of retaining the current SBT management of this is that it also works locally when running e.g. `sbt test`. If we configure it in GHA we'll also need to replicate somehow for local development too (likely via SBT).

akash1810:
> @akash1810 thanks. Re Postgres, an advantage of retaining the current SBT management of this is that it also works locally when running e.g. `sbt test`. If we configure it in GHA we'll also need to replicate somehow for local development too (likely via SBT).

I'm not a huge fan of using sbt for this, and would suggest a [`test` script](https://github.blog/2015-06-30-scripts-to-rule-them-all/) to run locally instead. Whilst it adds a little indirection, it's a more explicit as you can easily read the test script to see the order things are run in. It's arguably easier to add error handling for docker without resorting to [adding sbt libraries](https://github.com/guardian/riff-raff/pull/690).

PR #881, opened 2022-10-11
refactor: Simplify JSON for stack policy

akash1810:
## What does this change?

JSON formatting is hard...

We currently render a JSON array as:

```json
{
    "ResourceTypes": [
        "One",
"Two",
"Three"
    ]
}
```

That is, without indentation.

It should be:

```json
{
    "ResourceTypes": [
        "One",
        "Two",
        "Three"
    ]
}
```

However, there isn't much need to pretty-print this JSON blob as machines are the only consumers. In this change, we render a JSON array as a simple comma separated value:

```json
{
    "ResourceTypes": [
        "One","Two","Three"
    ]
}
```

This is functionally equivalent, and makes the tests simpler to read. That is, this is a no-op.

Also fixes a small typo.

## How to test

See CI - all tests should continue to pass.

### DIFF
--- a/magenta-lib/src/main/scala/magenta/tasks/SetStackPolicyTask.scala
+++ b/magenta-lib/src/main/scala/magenta/tasks/SetStackPolicyTask.scala
@@ -136,7 +136,7 @@ object StackPolicy {
          |      "Condition" : {
          |        "StringEquals" : {
          |          "ResourceType" : [
-         |            ${resourcesToProtect.mkString("\"","\",\n\"", "\"")}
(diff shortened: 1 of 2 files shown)


nicl (approved):
(I think you could actually manually add the formatting (newlines) in the *Test file and avoid the change in the *Task file altogether, but this is also fine.)

akash1810:
> (I think you could actually manually add the formatting (newlines) in the *Test file and avoid the change in the *Task file altogether, but this is also fine.)

I did (https://github.com/guardian/riff-raff/pull/881/commits/197de0fb90ecfbbf641b704ef245930447909c76), and it upset me 😂 .

PR #884, opened 2022-10-12
feat: Add warning when explicit bucket names are present in `riff-raff.yaml`

akash1810:
## What does this change?

Bucket names are deemed private, where private is:

> Things we do not want to be common knowledge, but knowing them does not directly compromise anything.
> Generally this means that knowing that detail would make an exploit (technical or people-wise) easier to achieve, especially when combining a few of them.
> 
> – source: [guardian/recommendations](https://github.com/guardian/recommendations/blob/d63e8dbad70d70193da2f0f7ff79181bc83138eb/github.md#private-information)

In this change, we add a warning when an explicit bucket name is provided, prompting to use `bucketSsmLookup` instead.

By default, `bucketSsmLookup` will use the SSM Parameter "/account/services/artifact.bucket", bu that can be customised via `bucketSsmKey`.

The default SSM Parameter should exist in all accounts now (/soon!), courtesy of https://github.com/guardian/aws-account-setup.

## How to test

TBD.

A gradual migration to removing private bucket names from VCS.

There will be a few projects that gain this new warning.

## Images

TBD.


kenoir (approved):

mchv:
Could we assume `bucketSsmLookup: true` if `bucketSsmLookup` and `bucket` parameter are not provided?

mchv:
I have just implemented suggestion in the comment above in #885

aaronp (commented):
LGTM 👍

davidfurey:
@akash1810 we are coming up to the 1 year anniversary of this warning. Have you reviewed how successful it has been? Would it be much work to get down to zero and remove support for explicit bucket names?

PR #885, opened 2022-10-13
Use bucket by ssm key by default

mchv:
## What does this change?

This enable to omit the `bucketSsmLookup` parameter, which reduce `riff-raff.yaml` verbosity.
This looks to me a detail that is not necessary especially if we expect people to use this over explicit `bucket` parameter.  

This should enable this:

```yaml
stacks:
  - deploy
regions:
  - eu-west-1
allowedStages:
  - PROD
deployments:
  cdk-playground:
    type: autoscaling
    parameters:
```

### DIFF
--- a/magenta-lib/src/test/scala/magenta/deployment_type/LambdaTest.scala
+++ b/magenta-lib/src/test/scala/magenta/deployment_type/LambdaTest.scala
@@ -11,7 +11,7 @@ import org.scalatest.matchers.should.Matchers
 import play.api.libs.json.{JsBoolean, JsString, JsValue, Json}
 import software.amazon.awssdk.services.s3.S3Client
 import software.amazon.awssdk.services.ssm.SsmClient
-import software.amazon.awssdk.services.ssm.model.{GetParameterRequest, GetParameterResponse, Parameter}
+import software.amazon.awssdk.services.ssm.model.{GetParameterRequest, GetParameterResponse, Parameter, SsmException}
 import software.amazon.awssdk.services.sts.StsClient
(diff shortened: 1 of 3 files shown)


akash1810:
Sensible change. Looks like the tests are failing.

magenta-lib/src/test/scala/magenta/deployment_type/LambdaTest.scala
```diff
@@ -119,23 +120,69 @@ class LambdaTest extends AnyFlatSpec with Matchers with MockitoSugar {
...
         "updateLambda").taskGenerator(pkg, DeploymentResources(reporter, lookupEmpty, artifactClient, stsClient, global),
         DeployTarget(parameters(PROD), Stack("some-stack"), region))
     }
-    e.message shouldBe "One and only one of the following must be set: the bucket parameter or bucketSsmLookup=true"
+
+    val ssmKey = BucketParametersDefaults.defaultSsmKeyParamDefault
+    e.message shouldBe s"Explicit bucket name has not been provided and failed to read bucket from SSM parameter: $ssmKey"
+  }
+
+  it should "default to lookup bucket from SSM bucket name is not provided and bucketSsmLookup is false" in {
```
nicl, line 155: 'from SSM [when] bucket name...'

nicl (approved):

kenoir:
This has been tested in https://riffraff.code.dev-gutools.co.uk/login, and can successfully deploy.

Screenshot 2022-10-21 at 09 41 26

PR #893, opened 2022-10-18
WIP: Add ssm bucket lookup support to aws-s3 deploy type

philmcmahon:
## What does this change?
Following from https://github.com/guardian/riff-raff/pull/704 here's another deploy type that uses a bucket which could be fetched from SSM rather than being hard coded. 

It's a bit more complex for this deploy type as the current behaviour supports a 3rd way of getting a bucket name - riffraff deployment resources. These can be seen [here](https://riffraff.gutools.co.uk/deployinfo/data?key=aws-bucket) and represent the old way of not hard coding a bucket name. SSM is superior because (as far as I understand) the only way of updating riffraff deployment resources is via the configuration for riffraff itself.

As part of this PR I've pulled the 'bucket resource' logic into `BucketParameters` which complicates it somewhat but I think it's best for all this logic to be in one place.

Following @akash1810's suggestion, I've made the ssm key a required option for the `aws-s3` deploy type - it is common to use this deploy type to upload files to e.g. a public bucket for static files - so we shouldn't assume it's the normal distribution bucket for that AWS account. 

## How to test
This needs testing on CODE, something for my next 10% time unless anyone wants to grab it 😉 

Fewer hard coded bucket names 🎉

### DIFF
--- a/magenta-lib/src/main/scala/magenta/deployment_type/AutoScaling.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/AutoScaling.scala
@@ -134,9 +134,6 @@ object AutoScaling extends DeploymentType with BucketParameters {
     groupsToUpdate.flatMap(asg => tasksPerAutoScalingGroup(asg))
   }
 
-  // TODO this is copied from `Lambda.scala` and could be DRYed out
-  def withSsm[T](keyRing: KeyRing, region: Region, resources: DeploymentResources): (SsmClient => T) => T = SSM.withSsmClient[T](keyRing, region, resources)
-
   val uploadArtifacts = Action("uploadArtifacts",
     """
       |Uploads the files in the deployment's directory to the specified bucket.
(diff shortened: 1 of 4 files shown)


akash1810:
> It's a bit more complex for this deploy type as the current behaviour supports a 3rd way of getting a bucket name - riffraff deployment resources. These can be seen [here](https://riffraff.gutools.co.uk/deployinfo/data?key=aws-bucket) and represent the old way of not hard coding a bucket name.

I wonder if we should remove this feature? It feels like a reinvention of SSM Parameter lookup. Additionally, there are only [two repos](https://github.com/search?l=YAML&q=org%3Aguardian+aws-bucket&type=Code) using it, with only https://github.com/guardian/reuse-content "live".

philmcmahon:
> > It's a bit more complex for this deploy type as the current behaviour supports a 3rd way of getting a bucket name - riffraff deployment resources. These can be seen [here](https://riffraff.gutools.co.uk/deployinfo/data?key=aws-bucket) and represent the old way of not hard coding a bucket name.
> 
> I wonder if we should remove this feature? It feels like a reinvention of SSM Parameter lookup. Additionally, there are only [two repos](https://github.com/search?l=YAML&q=org%3Aguardian+aws-bucket&type=Code) using it, with only https://github.com/guardian/reuse-content currently "live".

Agreed. I think there's some overlap with the GCP deployment type in the scala code but that could be untangled I'm sure. I'm not sure when I'll get round to this as 10% times are quite few and far between but potentially might be able to do something on friday.

PR #687, opened 2022-04-22
Add VCS URL tag

nicl:
Adds a `vcsUrl` tag to Cloudformation stacks on update to clarify ownership and discoverability.

Tested successfully on CODE:

Screenshot 2022-04-22 at 17 22 01

PR #927, opened 2022-11-21
Warn if unable to determine Github repository for a deploy

nicl:
Riffraff adds a `gu:repo` tag to Cloudformation stacks as part of the cloud-formation deployment type. This information is used by the nascent `services.gutools.co.uk` to help identify owners of stacks and related metadata.

However, in many cases, Riffraff is unable to populate this tag as it is unable to determine the repository for a build. The repository information is taken from the `vcsURL` field in the (build.json) 'manifest' file for a deployment and older build methods (or bespoke/script-based approaches) do not populate this. The latest versions of sbt-riffraff-artifact ([here](https://github.com/guardian/sbt-riffraff-artifact/blob/main/src/main/scala/com/gu/riffraff/artifact/BuildManifest.scala#L25)), node-riffraff-artifact ([here](https://github.com/guardian/node-riffraff-artifact/blob/main/src/manifest.ts#L5)) and, of course, actions-riff-raff do provide this info though.

At the moment this is just a warning. At some point we may need to do additional comms and even fail builds to motivate people here but let's get more data first.

### DIFF
--- a/magenta-lib/src/main/scala/magenta/deployment_type/CloudFormation.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/CloudFormation.scala
@@ -13,11 +13,11 @@ import org.joda.time.DateTime
 import scala.collection.mutable.ListBuffer
 
 trait VcsLookup {
-  def get(projectName: String, buildId: String): String
+  def get(projectName: String, buildId: String): Option[String]
 }
 
 // For testing purposes
-object NoopVcsUrlLookup extends VcsLookup { def get(projectName: String, buildId: String): String = "unknown"}  
+object NoopVcsUrlLookup extends VcsLookup { def get(projectName: String, buildId: String): Option[String] = None}
 
 
 //noinspection TypeAnnotation
@@ -172,8 +172,8 @@ class CloudFormation(vcsUrlLookup: VcsLookup) extends DeploymentType with CloudF
     // The tag name here ('gu:repo') and format ('guardian/:reponame') MUST
     // mirror the tag name and format used by @guardian/cdk.
     val vcsUrl = vcsUrlLookup.get(target.parameters.build.projectName, target.parameters.build.id)
-    val guRepoTag = ("gu:repo" -> vcsUrl)
-    if (vcsUrl == "") {
+    val guRepoTag = ("gu:repo" -> vcsUrl.getOrElse("unknown"))
+    if (vcsUrl.isEmpty) {
       reporter.warning("Unable to detect the Github repository for your build. DevX require this information to provide better reporting for teams as part of the upcoming Service Catalogue. This info is probably missing because you are using an unsupported library (or custom script) to generate your Riffraff bundle. To fix this, we recommend that you use https://github.com/guardian/actions-riff-raff/ or, if you are using https://github.com/guardian/sbt-riffraff-artifact or https://github.com/guardian/node-riffraff-artifact, upgrade to the latest version.")
     }
 
--- a/magenta-lib/src/main/scala/magenta/deployment_type/CloudFormation.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/CloudFormation.scala
@@ -171,8 +171,9 @@ class CloudFormation(vcsUrlLookup: VcsLookup) extends DeploymentType with CloudF
 
     // The tag name here ('gu:repo') and format ('guardian/:reponame') MUST
     // mirror the tag name and format used by @guardian/cdk.
-    val guRepoTag = ("gu:repo" -> vcsUrlLookup.get(target.parameters.build.projectName, target.parameters.build.id))
-    if (guRepoTag._2 == "") {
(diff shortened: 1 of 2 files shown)


magenta-lib/src/main/scala/magenta/deployment_type/CloudFormation.scala
```diff
@@ -172,6 +172,9 @@ class CloudFormation(vcsUrlLookup: VcsLookup) extends DeploymentType with CloudF
     // The tag name here ('gu:repo') and format ('guardian/:reponame') MUST
     // mirror the tag name and format used by @guardian/cdk.
     val guRepoTag = ("gu:repo" -> vcsUrlLookup.get(target.parameters.build.projectName, target.parameters.build.id))
+    if (guRepoTag._2 == "") {
+      reporter.warning("Unable to detect the Github repository for your build (used to provide better reporting to teams). This is probably because you are using an unsupported library (or custom script) to generate your Riffraff bundle. To fix this, we recommend that you use https://github.com/guardian/actions-riff-raff/ instead.")
```
akash1810, line 176: The recommended fix, of changing one's CI provider, is quite heavy here. I'm not sure it's the only one. I think using the latest version of [sbt-riffraff-artifact](https://github.com/guardian/sbt-riffraff-artifact) or [node-riffraff-artifact](https://github.com/guardian/node-riffraff-artifact) are other, more lightweight, options?

magenta-lib/src/main/scala/magenta/deployment_type/CloudFormation.scala
```diff
@@ -172,6 +172,9 @@ class CloudFormation(vcsUrlLookup: VcsLookup) extends DeploymentType with CloudF
     // The tag name here ('gu:repo') and format ('guardian/:reponame') MUST
     // mirror the tag name and format used by @guardian/cdk.
     val guRepoTag = ("gu:repo" -> vcsUrlLookup.get(target.parameters.build.projectName, target.parameters.build.id))
+    if (guRepoTag._2 == "") {
```
akash1810, line 175: Forgot about the `._` syntax in Scala 🙈.

nicl: Yeah, it's horrible :(

nicl: I'll lift it into a variable to make it more readable.

nicl:
@akash1810 thanks for the comments - agree the warning should be clearer and recommend upgrading existing build tools where appropriate.

We can tell from the services.gutools.co.uk data (or just cloudformation-lens) how many stacks are missing this tag at the moment and...it's a lot! 

* 405/1649 have the tag populated with a repo
* 115 have the tag populated with 'unknown'
* the remaining 1129 have no tag

That implies the 1129 either are deployed manually (so won't see a warning) or have not been deployed since the `gu:repo` tag behaviour was added to RR (~6 months ago [here](https://github.com/guardian/riff-raff/pull/687)).

So it's a big hill to climb and I'd be keen to get started asap. But the initial impact is likely to be on those 115 repos with 'unknown'.

magenta-lib/src/main/scala/magenta/deployment_type/CloudFormation.scala
```diff
@@ -171,7 +171,11 @@ class CloudFormation(vcsUrlLookup: VcsLookup) extends DeploymentType with CloudF
 
     // The tag name here ('gu:repo') and format ('guardian/:reponame') MUST
     // mirror the tag name and format used by @guardian/cdk.
-    val guRepoTag = ("gu:repo" -> vcsUrlLookup.get(target.parameters.build.projectName, target.parameters.build.id))
+    val vcsUrl = vcsUrlLookup.get(target.parameters.build.projectName, target.parameters.build.id)
+    val guRepoTag = ("gu:repo" -> vcsUrl)
+    if (vcsUrl == "") {
+      reporter.warning("Unable to detect the Github repository for your build. DevX require this information to provide better reporting for teams as part of the upcoming Service Catalogue. This info is probably missing because you are using an unsupported library (or custom script) to generate your Riffraff bundle. To fix this, we recommend that you use https://github.com/guardian/actions-riff-raff/ or, if you are using https://github.com/guardian/sbt-riffraff-artifact or https://github.com/guardian/node-riffraff-artifact, upgrade to the latest version.")
```
akash1810, line 177: Interested to see how this renders as it's quite a long message. If we place a break before the remediation steps, might it make it more obvious?

akash1810 (approved):
Minor comment, which we can fix separately if needed. Otherwise 👍🏽 . Do we need to send comms?

magenta-lib/src/main/scala/magenta/deployment_type/CloudFormation.scala
akash1810: Alternative could be to link to an external resource, similar to the [default branch warning](https://github.com/guardian/riff-raff/blob/06cee9eaa23648d197689ae3b0351fc417e326f6/riff-raff/app/deployment/actors/DeployGroupRunner.scala#L179).

PR #934, opened 2022-11-28
Add standard tags as part of Cloudformation deployments

nicl:
The standard tags are: App, Stack, Stage.

These tags are required by services.gutools.co.uk to group related stacks (those which belong to the same 'App' and 'Stack' but differ by 'Stage'.

Many stacks already have these tags, but non-CDK ones may not.

Nb. this change will force stack updates for some stacks. This shouldn't be a problem but it may confuse teams without some advance comms. The plan is to do a joint email covering this and also https://github.com/guardian/riff-raff/pull/927 at the same time.

### DIFF
--- a/magenta-lib/src/main/scala/magenta/deployment_type/CloudFormation.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/CloudFormation.scala
@@ -177,13 +177,23 @@ class CloudFormation(vcsUrlLookup: VcsLookup) extends DeploymentType with CloudF
       reporter.warning("Unable to detect the Github repository for your build. DevX require this information to provide better reporting for teams as part of the upcoming Service Catalogue. This info is probably missing because you are using an unsupported library (or custom script) to generate your Riffraff bundle. To fix this, we recommend that you use https://github.com/guardian/actions-riff-raff/ or, if you are using https://github.com/guardian/sbt-riffraff-artifact or https://github.com/guardian/node-riffraff-artifact, upgrade to the latest version.")
     }
 
+    val defaultTags = Map(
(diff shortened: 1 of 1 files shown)


kenoir (approved):

magenta-lib/src/main/scala/magenta/deployment_type/CloudFormation.scala
```diff
@@ -177,13 +177,23 @@ class CloudFormation(vcsUrlLookup: VcsLookup) extends DeploymentType with CloudF
       reporter.warning("Unable to detect the Github repository for your build. DevX require this information to provide better reporting for teams as part of the upcoming Service Catalogue. This info is probably missing because you are using an unsupported library (or custom script) to generate your Riffraff bundle. To fix this, we recommend that you use https://github.com/guardian/actions-riff-raff/ or, if you are using https://github.com/guardian/sbt-riffraff-artifact or https://github.com/guardian/node-riffraff-artifact, upgrade to the latest version.")
     }
 
+    val defaultTags = Map(
+      "Stack" -> target.stack.name,
+      "Stage" -> target.parameters.stage.name,
+      "App" -> pkg.app.name,
```
akash1810, line 183: I don't think `pkg.app.name` is the correct value here, as for the stack `cmp-monitoring-CODE` in the frontend account, it's App tag has value `cmp-monitoring-cdk-eu-west-1`, suggesting `pkg.app.name` is the [package name](https://github.com/guardian/consent-management-platform/blob/843f57e3c40f51a29a6d1a5bd8766a03b9c3852a/riff-raff.yaml#L6). Maybe we should move to making the `app` deployment property required (currently optional)?

PR #937, opened 2022-12-01
Record build tool

nicl:
Adds the buildTool as a tag on Cloudformation deploys. This will help us to track usage of build tools for services. The particular focus is @guardian/actions-riff-raff but the solution is general.

Note, as part of this, the reporter error for missing vcsUrl has been removed - looking more closely at the code, I believe deploys should already fail if this is not present so the warning here is redundant.

See also: https://github.com/guardian/actions-riff-raff/pull/23.

### DIFF
--- a/magenta-lib/src/main/scala/magenta/deployment_type/CloudFormation.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/CloudFormation.scala
@@ -4,24 +4,18 @@ import magenta.Loggable
 import magenta.Strategy.{Dangerous, MostlyHarmless}
 import magenta.artifact.S3Path
 import magenta.deployment_type.CloudFormationDeploymentTypeParameters._
-import magenta.tasks.CloudFormation.withCfnClient
-import magenta.tasks.StackPolicy.{accountPrivateTypes, allSensitiveResourceTypes}
 import magenta.tasks.UpdateCloudFormationTask.LookupByTags
 import magenta.tasks._
 import org.joda.time.DateTime
 
-import scala.collection.mutable.ListBuffer
-
-trait VcsLookup {
-  def get(projectName: String, buildId: String): Option[String]
+trait BuildTags {
+  // Returns tags for a build, which should be added to the Cloudformation
+  // stack. Tags are named with a `gu:` prefix.
+  def get(projectName: String, buildId: String): Map[String, String]
 }
 
-// For testing purposes
-object NoopVcsUrlLookup extends VcsLookup { def get(projectName: String, buildId: String): Option[String] = None}
-
-
 //noinspection TypeAnnotation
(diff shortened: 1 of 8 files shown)


riff-raff/app/AppComponents.scala
```diff
@@ -85,15 +83,22 @@ class AppComponents(context: Context, config: Config, passwordProvider: Password
...
+    def get(projectName: String, buildId: String): Map[String, String] = {
+      val build = builds.build(projectName, buildId)
+      val repoUrl = build.flatMap(b => VCSInfo.normalise(b.vcsURL))
+      val buildTool = build.flatMap(b => b.buildTool)
+      val default = "unknown"
+
+      Map(
+        "gu:repo" -> buildTool.getOrElse(default),
+        "gu:build-tool" -> repoUrl.getOrElse(default),
+      )
```
kenoir, line 96:
```suggestion
      Map(
        "gu:repo" -> repoUrl.getOrElse(default),
        "gu:build-tool" -> buildTool.getOrElse(default),
      )
```

magenta-lib/src/main/scala/magenta/deployment_type/CloudFormation.scala
```diff
@@ -12,16 +12,24 @@ import org.joda.time.DateTime
 
 import scala.collection.mutable.ListBuffer
 
-trait VcsLookup {
-  def get(projectName: String, buildId: String): Option[String]
+//trait VcsLookup {
```
kenoir, line 15: Comment can be removed?

riff-raff/test/ci/ContinuousDeploymentTest.scala
```diff
@@ -43,9 +43,9 @@ class ContinuousDeploymentTest extends AnyFlatSpec with Matchers {
   val contDeployConfigs = Seq(tdProdEnabled, tdCodeDisabled, td2ProdDisabled, td2QaEnabled)
   val contDeployBranchConfigs = Seq(tdProdEnabled, tdCodeDisabled, td2ProdDisabled, td2QaBranchEnabled, td2ProdBranchEnabled)
 
-  val tdB71 = S3Build(45397, "tools::deploy", "45397", "branch", "71", new DateTime(2013,1,25,14,42,47), "", "")
-  val td2B392 = S3Build(45400, "tools::deploy2", "45400", "branch", "392", new DateTime(2013,1,25,15,34,47), "", "")
-  val otherBranch = S3Build(45401, "tools::deploy2", "45401", "other", "393", new DateTime(2013,1,25,15,34,47), "", "")
+  val tdB71 = S3Build(45397, "tools::deploy", "45397", "branch", "71", new DateTime(2013,1,25,14,42,47), "", "", None)
```
kenoir, line 46: Not blocking, but perhaps:
- Using named parameters though more verbose is a good standard practise (to avoid the kind of issue like the switched params above).
- Filling the extra fields out here with fake data for build tool & repo would make these tests more useful.

kenoir (approved):

PR #941, opened 2022-12-06
Extend aws-s3 deployment type to support bucket SSM parameters

nicl:
Builds on https://github.com/guardian/riff-raff/pull/942, which provides an alternative to S3 resource-based parameters, to do two things:

1) support SSM parameters for S3 deployments
2) remove support for resource parameters (`bucketResource`, `pathPrefixResource`)

Note, *only `pathPrefixResource` has been removed here* but `bucketResource` will be removed as a follow-up once the affected team have migrated. Resource parameters are virtually unused and are better served through the more general SSM parameters (`bucketSsmKey`, or `bucketSsmKeyStageParam`).

This PR is inspired by https://github.com/guardian/riff-raff/pull/893, which it now supercedes. Thanks to @philmcmahon for the hard work on that.

### DIFF
--- a/magenta-lib/src/main/scala/magenta/deployment_type/S3.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/S3.scala
@@ -2,9 +2,9 @@ package magenta.deployment_type
 
 import magenta.Datum
 import magenta.deployment_type.param_reads.PatternValue
-import magenta.tasks.S3Upload
+import magenta.tasks.{S3Upload, S3 => S3Tasks, SSM}
 
-object S3 extends DeploymentType {
+object S3 extends DeploymentType with BucketParameters {
   val name = "aws-s3"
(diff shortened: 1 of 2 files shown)


magenta-lib/src/main/scala/magenta/deployment_type/S3.scala
```diff
@@ -187,20 +179,26 @@ object S3 extends DeploymentType {
...
-          stack =
-            if (prefixStack(pkg, target, reporter)) Some(target.stack)
-            else None,
-          stage =
-            if (prefixStage(pkg, target, reporter))
-              Some(target.parameters.stage)
-            else None,
-          packageOrAppName = maybePackageOrAppName
-        )
+      S3Upload.prefixGenerator(
```
kenoir, line 182: After reviewing with @nicl, he noticed this block can be removed.

kenoir (approved):

PR #942, opened 2022-12-06
Introduce bucketSsmKeyStageParam

nicl:
To support using a different SSM path per stage. Especially useful to allow us to remove Deployment Resources, which are used in part for this reason.

Once merged, we can migrate the few uses of S3-related deployment resources, and merge https://github.com/guardian/riff-raff/pull/941.

### DIFF
--- a/magenta-lib/src/main/scala/magenta/deployment_type/BucketParameters.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/BucketParameters.scala
@@ -28,28 +28,59 @@ trait BucketParameters {
     """The SSM key used to lookup the bucket name for uploading distribution artifacts.""".stripMargin
   ).default(defaultSsmKeyParamDefault)
 
+  val bucketSsmKeyStageParam = Param[Map[String, String]](
+    "bucketSsmKeyStageParam",
+    """
+      |Like bucketSsmKeyParam, but with the ability to configure by stage:
+      |
+      |```yaml
+      |  CODE: some-ssm-path-for-code
+      |  PROD: some-ssm-path-for-prod
+      |```
+      |""".stripMargin
+  ).default(Map.empty)
+
   def getTargetBucketFromConfig(
       pkg: DeploymentPackage,
       target: DeployTarget,
       reporter: DeployReporter
   ): Bucket = {
-    val bucketSsmLookup = bucketSsmLookupParam(pkg, target, reporter)
-    val maybeExplicitBucket = bucketParam.get(pkg)
 
-    val bucket = (bucketSsmLookup, maybeExplicitBucket) match {
-      // Default to looking up target bucket from SSM
-      case (_, None) =>
-        BucketBySsmKey(bucketSsmKeyParam(pkg, target, reporter))
-      case (false, Some(explicitBucket)) =>
+    def bySsm(): Bucket = {
+      val stage = target.parameters.stage.name
+      val ssmKeyByStage = bucketSsmKeyStageParam(pkg, target, reporter)
+      val stageKey = ssmKeyByStage.get(stage)
+
+      if (ssmKeyByStage.nonEmpty && stageKey.isEmpty) {
+        reporter.fail(
+          s"Unable to determine bucket to deploy to: bucketSsmKeyStageParam is set but no mapping was found for stage '$stage'."
+        )
+      }
+
+      BucketBySsmKey(
+        stageKey.getOrElse(bucketSsmKeyParam(pkg, target, reporter))
+      )
+    }
+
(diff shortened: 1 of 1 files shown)


magenta-lib/src/main/scala/magenta/deployment_type/BucketParameters.scala
```diff
@@ -26,20 +26,36 @@ trait BucketParameters {
...
-      // Default to looking up target bucket from SSM
-      case (_, None) =>
-        BucketBySsmKey(bucketSsmKeyParam(pkg, target, reporter))
-      case (false, Some(explicitBucket)) =>
-        reporter.warning("Explicit bucket name in riff-raff.yaml. Prefer to use bucketSsmLookup=true, removing private information from VCS.")
-        BucketByName(explicitBucket)
-      case (true, Some(explicitBucket)) =>
-        reporter.fail(s"Bucket name provided ($explicitBucket) & bucketSsmLookup=true, please choose one or omit both to default to SSM lookup.")
+
+    val bucket = bucketSsmLookup match {
```
kenoir, line 42: A bit nitpicky: I prefer using an if/else over match on a boolean.

magenta-lib/src/main/scala/magenta/deployment_type/BucketParameters.scala
```diff
@@ -26,20 +26,36 @@ trait BucketParameters {
...
+      case true =>
+        val stage = target.parameters.stage.name
+        val ssmKeyByStage = bucketSsmKeyStageParam(pkg, target, reporter)
+        val stageKey = ssmKeyByStage.get(stage)
+        if (ssmKeyByStage.nonEmpty && stageKey.isEmpty) {
```
kenoir, line 47: It took me a little while to parse the logic here, perhaps this is a good place for a case statement?

```scala
(bucketSsmLookup, bucketParam.get(pkg), ssmKeyByStage, ssmKeyByStage.get(stage)) match {
  // If bucketSsmLookup false and no bucketParam, use default
  (false, None, _,_) => BucketBySsmKey(defaultSsmKeyParamDefault)

  // If bucketSsmLookup false and bucketParam, use bucketParam
  (false, Some(ssmKeyName), _, _) => BucketByName(ssmKeyName)

  // If bucketSsmLookup present, no stage, and bucketParam specified, use bucketParam
  (true, Some(ssmKeyName), ssmKeyByStage, _) if ssmKeyByStage.isEmpty => 
    BucketByName(ssmKeyName)

  // If bucketParam AND ssmKeyByStage specified, fail
  (true, Some(_), ssmKeyByStage, _) if ssmKeyByStage.nonEmpty => 
    reporter.fail(s"Unable to determine bucket to deploy to, please specify only one on of bucketSsmKeyParam & bucketSsmKeyStageParam")

  // If ssmKeyByStage is not empty, but there is no valid stage, fail
  (true, None, ssmKeyByStage, None) if ssmKeyByStage.nonEmpty => 
    reporter.fail(s"Unable to determine bucket to deploy to: bucketSsmKeyStageParam is set but no mapping was found for stage '$stage'.")

  // If ssmKeyByStage is not empty, and there is a valid stage, use that
  (true, None, ssmKeyByStage, Some(stageKey) if ssmKeyByStage.nonEmpty => 
    BucketBySsmKey(stageKey)
}

kenoir (approved):

magenta-lib/src/main/scala/magenta/deployment_type/BucketParameters.scala
nicl: Okay think you're right that pattern matching can help improve confidence that we're handling each case. I've pulled out some of the SSM stuff to simplify hopefully:

```    val bucket = (bucketSsmLookup, explicitBucket) match {
      case (true, Some(name)) =>
        reporter.fail(
          s"Bucket name provided ($name) & bucketSsmLookup=true, please choose one or omit both to default to SSM lookup."
        )
      case (true, None) =>
        bySsm()
      case (false, Some(name)) =>
        reporter.warning(
          "Explicit bucket name in riff-raff.yaml. Prefer to use bucketSsmLookup=true, removing private information from VCS."
        )
        BucketByName(name)
      case (false, None) =>
        BucketBySsmKey(defaultSsmKeyParamDefault)
    }```

kenoir (approved):

akash1810:
> To support using a different SSM path per stage.

What are the use-cases for this? For application artifacts, I think having a single bucket in the account is best as it keeps things simple.

One possible use-case I can think of is for S3 deployments, but not sure if teams currently use separate buckets or a single one, sharded by stage? The [recommendations](https://github.com/guardian/recommendations/discussions/74) suggest using a bucket per stage, but not sure if that's a reality.

PR #630, opened 2021-06-15
Add `braze-publish` deployment type (plus hidden `lambda-invoke` deployment type)

rhystmills:
Co-Authored-By: @twrichards 

The primary motivation for this PR was to facilitate version control and automated deployment of Braze content blocks, making use of [`guardian/braze-version-control-service`](https://github.com/guardian/braze-version-control-service).

## What does this change?
As per the branch name, we initially intended to implement a generic `lambda-invoke` deployment type, but there were concerns around this being used too broadly. Instead, we have kept the `lambda-invoke` deployment type hidden, but extended it in the form of the `braze-publish` deployment type (which adds a more specific description and suppresses user-entered parameters in favour of hard-coded references to the `publish-lambda` - see https://github.com/guardian/braze-version-control-service/pull/3). 

Some example usages of this new `braze-publish` deployment type can be seen in:
- https://github.com/guardian/braze-version-control/blob/main/riff-raff.yaml
- https://github.com/guardian/braze-version-control-TEST/blob/main/riff-raff.yaml

To implement the generic `lambda-invoke` deployment type, we abstracted common behaviour (such as being able to look up lambdas by function name) from `Lambda.scala` into `LambdaDeploymentType.scala` which is extended by both:
- `LambdaDeploy` (existing behaviour but new file name)
- `LambdaInvoke` (new)

The generic `lambda-invoke` deployment type works by taking the artifact files and building the JSON input payload for the lambda from their filenames (keys) and the text content (values) - see the documentation screenshot below for an example. We then invoke the lambda synchronously, which relies on the `lambda:InvokeFunction` IAM permissions added in https://github.com/guardian/deploy-tools-platform/pull/538. The console/log output from the lambda invocation is given to the riff-raff reporter as verbose output, and the lambda response payload as info output. 

When the DevX team are comfortable in opening up the `lambda-invoke` deployment type to the department it should be as simple as uncommenting: https://github.com/guardian/riff-raff/blob/5dd93b0abeb8d1d3bba8162bdab535677b3e656f/riff-raff/app/AppComponents.scala#L89

## How to test
1. Deploy this branch to `riff-raff` CODE.
2. Deploy `braze-version-control-TEST` and observe the deployment detail, e.g. https://riffraff.code.dev-gutools.co.uk/deployment/view/37d754a2-8920-477b-93cd-d84df7b728a0?verbose=1

## How can we measure success?
[...]

PR #945, opened 2022-12-08
Add Lambda Layer deployment type

kenoir:
## What does this change?

In order to deploy new versions of lambda layers via riff-raff in a sensible manner we should add support for an aws-lambda-layer deployment type to riff-raff.

Currently this is possible using the aws-s3 deployment type, but that mechanism results in complete replacement of the layer. This is unsuitable as if layers are being used to support devx functionality in our estate we must have a mechanism of safely updating those layers.

The deployment mechanism in riff-raff should use the [PublishLayerVersion](https://docs.aws.amazon.com/lambda/latest/dg/API_PublishLayerVersion.html) API in AWS to update layers creating new versions without replacing the layer (and consequently creating new ARNs each time).

Currently this deployment type will publish a new version every time it is deployed, and will not clean up old versions. I think this behaviour is acceptable, but we may want to modify it if we find this deployment type gets more usage.

Screenshot 2022-12-14 at 15 41 58

It is safe to manually delete older versions without impacting the lambdas that use those old versions, they will continue to function but you will not be able to create new lambdas that make use of those deleted layers.

## How do we know this works?

This has been tested in code, and you can see a successful deployment of https://github.com/guardian/ssm-to-env/pull/4:
- https://riffraff.code.dev-gutools.co.uk/deployment/view/1447ba5e-1a50-4e6f-99ec-98aacebffa2b

### DIFF
--- a/magenta-lib/src/main/scala/magenta/deployment_type/LambdaLayer.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/LambdaLayer.scala
@@ -60,7 +60,7 @@ object LambdaLayer extends DeploymentType with BucketParameters {
 
     layerNameParam.get(pkg) match {
       case Some(name) => UpdateLambdaLayer(
-        layer = LambdaLayerName(s"$name$stage"),
+        layer = LambdaLayerName(name),
         fileName = fileNameParam(pkg, target, reporter),
         region = target.region,
         s3Bucket = bucket
--- a/magenta-lib/src/main/scala/magenta/deployment_type/LambdaLayer.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/LambdaLayer.scala
@@ -1,0 +1,152 @@
+package magenta.deployment_type
+
+import magenta.artifact.S3Path
+import magenta.tasks.{SSM, UpdateS3LambdaLayer}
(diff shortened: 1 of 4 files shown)


magenta-lib/src/main/scala/magenta/deployment_type/LambdaLayer.scala
```diff
@@ -0,0 +1,152 @@
...
+
+import magenta.artifact.S3Path
+import magenta.tasks.{SSM, UpdateS3LambdaLayer}
+import magenta.{DeployReporter, DeployTarget, DeploymentPackage, DeploymentResources, KeyRing, Region, tasks}
+import software.amazon.awssdk.services.s3.S3Client
+import software.amazon.awssdk.services.ssm.SsmClient
+
+object LambdaLayer extends LambdaLayer
+
+trait LambdaLayer extends DeploymentType with BucketParameters {
```
kenoir, line 11: Why can't this just extend object?

nicl (approved):
LGTM! 🚢

twrichards:
@kenoir could this PR make use of the refactors in https://github.com/guardian/riff-raff/pull/630 now that we've just marked that as ready for review again

PR #948, opened 2022-12-12
Revert "Revert "Introduce bucketSsmKeyStageParam""

nicl:
**~~TODO: in the process of writing some explicit tests for the BucketParameters trait.~~ DONE**

* [Original PR](https://github.com/guardian/riff-raff/pull/942) aimed to introduce stage support for `bucketSsmKeyParam`
* [Revert](https://github.com/guardian/riff-raff/pull/947) as the above broke the default behaviour

This PR restores the original PR while fixing the (accidental) breaking change that the original PR introduced - see the extra commit that ensures we default to SSM lookup even when `bucketSsmLookup=false`.

I have some quibbles with the clarity of this behaviour going forward, but it's not clear that we can change it now that the behaviour exists.

### DIFF
--- a/magenta-lib/src/main/scala/magenta/deployment_type/BucketParameters.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/BucketParameters.scala
@@ -28,28 +28,59 @@ trait BucketParameters {
     """The SSM key used to lookup the bucket name for uploading distribution artifacts.""".stripMargin
   ).default(defaultSsmKeyParamDefault)
 
+  val bucketSsmKeyStageParam = Param[Map[String, String]](
+    "bucketSsmKeyStageParam",
+    """
+      |Like bucketSsmKeyParam, but with the ability to configure by stage:
+      |
+      |```yaml
+      |  CODE: some-ssm-path-for-code
+      |  PROD: some-ssm-path-for-prod
+      |```
+      |""".stripMargin
+  ).default(Map.empty)
+
   def getTargetBucketFromConfig(
       pkg: DeploymentPackage,
       target: DeployTarget,
       reporter: DeployReporter
   ): Bucket = {
-    val bucketSsmLookup = bucketSsmLookupParam(pkg, target, reporter)
-    val maybeExplicitBucket = bucketParam.get(pkg)
 
-    val bucket = (bucketSsmLookup, maybeExplicitBucket) match {
-      // Default to looking up target bucket from SSM
-      case (_, None) =>
-        BucketBySsmKey(bucketSsmKeyParam(pkg, target, reporter))
-      case (false, Some(explicitBucket)) =>
+    def bySsm(): Bucket = {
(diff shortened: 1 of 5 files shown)


magenta-lib/src/main/scala/magenta/deployment_type/BucketParameters.scala
```diff
@@ -28,28 +28,59 @@ trait BucketParameters {
...
     val bucketSsmLookup = bucketSsmLookupParam(pkg, target, reporter)
-    val maybeExplicitBucket = bucketParam.get(pkg)
+    val explicitBucket = bucketParam.get(pkg)
 
-    val bucket = (bucketSsmLookup, maybeExplicitBucket) match {
-      // Default to looking up target bucket from SSM
-      case (_, None) =>
-        BucketBySsmKey(bucketSsmKeyParam(pkg, target, reporter))
-      case (false, Some(explicitBucket)) =>
+    // The behaviour here is *very* counter-intuitive; even if bucketSsmLookup=false
```
kenoir, line 68: I think what's missing here is the default value of `bucketSsmLookup` is `false` and the default behaviour we want is to use `bucketSsmLookup` - so we are confusingly using the signal that `bucketSsmLookup` hasn't been changed to perform that default behaviour!

So should the default value of `bucketSsmLookup` be `true` and then we can remove some of the confusing logic here.

kenoir (approved):

akash1810:
To double check my understanding of this, what are the use-cases this solves? For application artifacts, I think having a single bucket in the account is best as it keeps things simple.

One possible use-case I can think of is for S3 deployments, but not sure if teams currently use separate buckets or a single one, sharded by stage? The [recommendations](https://github.com/guardian/recommendations/discussions/74) suggest using a bucket per stage, but not sure if that's a reality.

magenta-lib/src/main/scala/magenta/deployment_type/BucketParameters.scala
kenoir: Or after talking with @nicl ... maybe we can just remove the param?

nicl:
@akash1810 yeah it's a good question. The immediate motivation is so we can remove the S3-related deployment resources in Riffraff. These are currently used to vary bucket by stage for a couple of projects. Generally speaking teams shouldn't use this but there are exceptions - e.g. static sites - where this is useful.

magenta-lib/src/main/scala/magenta/deployment_type/BucketParameters.scala
nicl: 👍 will do this in a follow-up once I've confirmed it is okay. But I think we should be safe to remove the `bucketSsmLookup` parameter completely.

Note, 'remove' here means ignoring it and reporting a warning whenever it is set. We can't remove it unfortunately as RR errors on unrecognised fields in a riff-raff.yaml file.

PR #950, opened 2022-12-12
Disable flaky tests

nicl:
These tests frequently (perhaps 50% of runs or more) fail with:

scala.reflect.internal.Symbols$CyclicReference: illegal cyclic reference involving class DescribeInstanceHealthResponse

A full test run with failures can be found [here](https://github.com/guardian/riff-raff/actions/runs/3676832331/jobs/6218284172).

We should fix this, but in the meantime let's ensure our tests are reliable.

### DIFF
--- a/magenta-lib/src/test/scala/magenta/tasks/ASGTest.scala
+++ b/magenta-lib/src/test/scala/magenta/tasks/ASGTest.scala
@@ -266,7 +266,12 @@ class ASGTest extends AnyFlatSpec with Matchers with MockitoSugar {
     group shouldBe Some(desiredGroup)
   }
 
-  it should "wait for instances in ELB to stabilise if there is one" in {
+  // TODO: these tests regularly fail with `An illegal reflective access
+  // operation has occurred`. This appears to be caused by Mockito when mocking
+  // the `ApplicationELBClient` and/or `ClassicELBClient` classes. We should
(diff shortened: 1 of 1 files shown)


magenta-lib/src/test/scala/magenta/tasks/ASGTest.scala
```diff
@@ -266,133 +266,138 @@ class ASGTest extends AnyFlatSpec with Matchers with MockitoSugar {
...
-      updatedGroup,
-      ELB.Client(classicELBClient, appELBClient)
-    ) shouldBe Right(())
-  }
+  // TODO: these tests regularly fail with `An illegal reflective access
+  // operation has occurred`. This appears to be caused by Mockito when mocking
+  // the `ApplicationELBClient` and/or `ClassicELBClient` classes. We should
+  // investigate this further and re-enable these tests once fixed.œ
+
+//  it should "wait for instances in ELB to stabilise if there is one" in {
```
akash1810, line 274: Could we use the `@Ignore` annotation? It'll mean we'll be reminded to fix the test in the log (both in CI and locally).

nicl: 👍 agreed, much better than just commenting out.

akash1810 (approved):

PR #951, opened 2022-12-13
fix: Use correct values for `gu:repo` and `gu:build-tool`

akash1810:
## What does this change?

Uses the correct value for the `gu:repo` and `gu:build-tool` tags.

## How to test

TBD.


nicl (approved):
🎂

PR #952, opened 2022-12-15
Remove bucketResource

nicl:
**Note: blocked by/depends on https://github.com/guardian/reuse-content/pull/40.**

Following on from https://github.com/guardian/riff-raff/pull/941 we can now delete this parameter too.

There is only a single user of this parameter, and the purpose (varying bucket by stage) is better achieved with the new [bucketSsmKeyStageParam](https://github.com/guardian/riff-raff/blob/23443fc9fe1fd8da5619cb74aef9be3d41376218/magenta-lib/src/main/scala/magenta/deployment_type/BucketParameters.scala#L31).


kenoir (approved):
👏 much simpler.

PR #972, opened 2023-01-16
fix: Add `AWS::EFS::FileSystem` to set of protected resources

akash1810:
## What does this change?

Configure Riff-Raff to protect EFS, as it holds state.

Use-cases for EFS include persisting data across instance rotations, resulting in little/no service disruption. When an instance rotates into service, it'll mount the EFS volume and not need to rebuild any state, as it has it.

Examples include:
  - Grid FTP using EFS to hold transferred images
  - TeamCity server using EFS to hold configuration, logs, caches, etc

It's worth noting that, as ever, "Dangerous" mode will allow you to delete/replace these resources.

See https://docs.aws.amazon.com/efs/latest/ug/how-it-works.html.


twrichards (approved):
🙏

PR #980, opened 2023-01-23
feat(aws-s3): Allow customisation of upload path to S3

akash1810:
## What does this change?

Allow for the upload path to S3 to be customised.

By default, create a directory for each prefix, e.g. `/<stack>/<stage>/files`.

Within Interactives, there is a desire to upload to `/<stack>-<stage>/files`, so introduce a new optional parameter `prefixDelimiter` to support this. Specifically, [`interactive-atom-maker`](https://github.com/guardian/interactive-atom-maker) is [configured](https://github.com/guardian/interactive-atom-maker/blob/983da014e0afe623f0c211f77a92f10a371fed2b/src/main/scala/app/Config.scala#L52) to read from `atoms-STAGE` (TBD!)

See https://github.com/guardian/csti-wip-interactive-template.

### DIFF
--- a/magenta-lib/src/main/scala/magenta/deployment_type/AutoScaling.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/AutoScaling.scala
@@ -264,6 +264,7 @@ object AutoScaling extends DeploymentType with BucketParameters {
     }
 
     val prefix = S3Upload.prefixGenerator(
+      delimiter = "/",
       stack =
         if (prefixStack(pkg, target, reporter)) Some(target.stack) else None,
       stage =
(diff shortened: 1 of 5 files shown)


kenoir:
I am presuming there is some hidden complexity here around updating `interactive-atom-maker` to read from a `/` only delimited path?

akash1810:
> I am presuming there is some hidden complexity here around updating `interactive-atom-maker` to read from a `/` only delimited path?

Yes. Currently we publish new interactive atoms manually on one's machine 😱 . CSTI are proposing a lightweight CI/CD process to remove this manual step. The complexity lies in needing to support both workflows, as the proposal is for new atoms. We could update interactive-atom-maker to watch a list of S3 paths, which might be better.

In any case, a discussion with CSTI yesterday suggested this isn't needed right now as whilst they are in a POC stage, they're happy to hard-code things their side.

Converting to draft.

akash1810:
This is superseded by https://github.com/guardian/riff-raff/pull/1015. Closing.

PR #991, opened 2023-01-28
fix: Observe `Encrypted` when set in `amiParametersToTags`

akash1810:
## Background
In https://github.com/guardian/riff-raff/pull/479 the parameter `amiEncrypted` was added to the `riff-raff.yaml` file, with the following behaviour:
  - Default of `false`
  - When `false`, Riff-Raff locates AMIs without the `Encrypted` tag, or where the tag is set to `false`
  - When `true`, Riff-Raff locates AMIs with the `Encrypted` set to `true`

The value of `amiEncrypted` is applied to _all_ AMIs in the stack.

The `amiParametersToTags` parameter allows one to describe the characteristics of multiple AMIs in the stack, independently. However, it's not currently possible to set encryption independently across these AMIs.

## What does this change?
In this change, we update the `getLatestAmi` logic to support the use-case where a stack has multiple AMIs, and where each can be encrypted, or not.

For example:
```yaml
deployments:
  update-cloudformation:
    type: "cloud-formation"
    parameters:
      amiParametersToTags:
        elasticsearchAmi:
          BuiltBy: "amigo"
          Recipe: "elk-elasticsearch"
          AmigoStage: "PROD"
          Encrypted: "true" # <-- this
         kibanaAmi:
           BuiltBy: "amigo"
           Recipe: "elk-kibana"
           AmigoStage: "PROD"
           Encrypted: "false" # <-- this   
```

> **Note**
> It could be argued that we should not support unencrypted workflows anymore! If so, a lot of this logic can be removed.
> 
> This'll require an amount of careful consideration so that we don't accidentally break deployments. That is, this PR is valuable today.

Behaviours should include:
1. `amiEncrypted` set to `true` should result in an encrypted AMI being used, as before
2. `amiEncrypted` set to `false` should result in an unencrypted AMI being used, as before
3. `amiParametersToTags/Encrypted` set to `"true"` should result in an encrypted AMI being used
4. `amiParametersToTags/Encrypted` set to `"false"` should result in an encrypted AMI being used
5. Absence of `amiEncrypted` and `amiParametersToTags/Encrypted` should result in an unencrypted AMI being used, as before

I've raised some PRs in [cdk-playground](https://github.com/guardian/cdk-playground) to help test this change:
- https://github.com/guardian/cdk-playground/pull/327
- https://github.com/guardian/cdk-playground/pull/328

With these PRs we can explicitly test behaviours 1, 3, and 5 (and implicitly test 2, and 4):
1. Deploy this branch to Riff-Raff CODE
2. Using Riff-Raff CODE, deploy `main` of `devx::cdk-playground`
3. Observe the AMI in use _does not_ have an `Encrypted` tag set to `true`
4. Using Riff-Raff CODE, deploy one of the `cdk-playground` PRs above
5. Observe the AMI in use _does_ have an `Encrypted` tag set to `true`
6. Using Riff-Raff CODE, deploy `main` of `devx::cdk-playground`
7. Observe the AMI in use _does not_ have an `Encrypted` tag set to `true`
8. Using Riff-Raff CODE, deploy the other `cdk-playground` PR above
9. Observe the AMI in use _does_ have an `Encrypted` tag set to `true`

#### Find the value of the AMI Parameter
```sh
aws cloudformation describe-stacks \
  --stack-name playground-PROD-cdk-playground \
  jq -r '.Stacks[].Parameters[] | select(.ParameterKey =="AMICdkplayground") | .ParameterValue'
```

#### Find an AMI's tags
```sh
aws ec2 describe-images \
  --image-ids ami-00000000000000000 \
  jq -r '.Images[].Tags[]'
```

FYI unencrypted AMIs live in the DevX account. Tags are not copied when sharing AMIs across accounts, so when using the `developerPlayground` account profile, the unencrypted AMI will not have any tags.

### DIFF
--- a/magenta-lib/src/main/scala/magenta/deployment_type/CloudFormationDeploymentTypeParameters.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/CloudFormationDeploymentTypeParameters.scala
@@ -157,24 +157,22 @@ trait CloudFormationDeploymentTypeParameters {
       reporter: DeployReporter,
       lookup: Lookup
   ): CfnParam => String => String => Map[String, String] => Option[String] = {
-    amiIdCfnParam =>
+    amiCfnParam =>
       { accountNumber =>
         {
-          (amiParametersToTags.get(pkg), amiTags.get(pkg)) match {
-            case (Some(params), _) =>
-              val useEncryptedAmi = params
-                .getOrElse(amiIdCfnParam, Map.empty)
+          val useEncryptedAmi: Boolean = amiParametersToTags.get(pkg) match {
+            case Some(params) =>
+              params
+                .getOrElse(amiCfnParam, Map.empty)
                 .get("Encrypted")
                 .contains("true")
+            case _ => amiEncrypted(pkg, target, reporter)
+          }
 
-              if (useEncryptedAmi) {
-                lookup.getLatestAmi(Some(accountNumber), _ => true)
-              } else {
-                lookup.getLatestAmi(None, unencryptedTagFilter)
-              }
-            case (None, Some(_)) if amiEncrypted(pkg, target, reporter) =>
-              lookup.getLatestAmi(Some(accountNumber), _ => true)
-            case _ => lookup.getLatestAmi(None, unencryptedTagFilter)
+          if (useEncryptedAmi) {
+            lookup.getLatestAmi(Some(accountNumber), _ => true)
+          } else {
+            lookup.getLatestAmi(None, unencryptedTagFilter)
           }
         }
       }
--- a/magenta-lib/src/main/scala/magenta/deployment_type/CloudFormationDeploymentTypeParameters.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/CloudFormationDeploymentTypeParameters.scala
@@ -5,7 +5,7 @@ import magenta.tasks.UpdateCloudFormationTask.{
   LookupByName,
   LookupByTags
 }
-import magenta.{DeploymentPackage, DeployReporter, DeployTarget, Lookup}
+import magenta.{DeployReporter, DeployTarget, DeploymentPackage, Lookup}
 
 object CloudFormationDeploymentTypeParameters {
   type TagCriteria = Map[String, String]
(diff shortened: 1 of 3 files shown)


magenta-lib/src/main/scala/magenta/tasks/UpdateCloudFormationTask.scala
```diff
@@ -396,7 +400,9 @@ case class UpdateAmiCloudFormationParameterTask(
     region: Region,
     cloudFormationStackLookupStrategy: CloudFormationStackLookupStrategy,
     amiParameterMap: Map[CfnParam, TagCriteria],
-    latestImage: String => String => Map[String, String] => Option[String],
+    latestImage: CfnParam => String => String => Map[String, String] => Option[
+      String
+    ],
```
akash1810, line 405: Not a fan of ScalaFmt's opinion here, but 🤷🏽 .

magenta-lib/src/main/scala/magenta/deployment_type/CloudFormationDeploymentTypeParameters.scala
```diff
@@ -156,12 +156,25 @@ trait CloudFormationDeploymentTypeParameters {
...
+  ): CfnParam => String => String => Map[String, String] => Option[String] = {
+    amiCfnParam =>
+      { accountNumber =>
+        {
+          val useEncryptedAmi: Boolean = amiParametersToTags.get(pkg) match {
+            case Some(params) =>
+              params
+                .getOrElse(amiCfnParam, Map.empty)
+                .get("Encrypted")
+                .contains("true")
```
twrichards, line 168: do we need a toLowercase in here first for a safe comparison to the string `"true"`
```suggestion
                .map(_.toLowercase)
                .contains("true")
```

twrichards (approved):
great change/fix. the code LGTM and it looks like you got testing covered pretty comprehensively elsewhere. 👏

magenta-lib/src/main/scala/magenta/deployment_type/CloudFormationDeploymentTypeParameters.scala
akash1810: I'm tempted to not do this, as [we don't elsewhere](https://github.com/guardian/riff-raff/blob/main/magenta-lib/src/main/scala/magenta/deployment_type/CloudFormationDeploymentTypeParameters.scala#L14-L22), and this tag shouldn't be manually set, so should be reliably cased. WDYT?

twrichards: yeah fair enough, although what about using https://www.scala-lang.org/files/archive/api/2.13.x/scala/collection/StringOps.html#toBooleanOption:Option[Boolean] then pattern match or something?

akash1810: Oooh! IIUC that'll become:

```scala
params
  .getOrElse(amiCfnParam, Map.empty)
  .getOrElse("Encrypted", "false") // default to `"false"` if no `Encrypted` tag
  .toBooleanOption.getOrElse(false) // default to `false` if failed to parse the tag's string value
```

This form doesn't read as easy to me, though I might be missing a more terse implementation.

kenoir (approved):
I can parse the logic here and understand it's been well tested.

magenta-lib/src/main/scala/magenta/deployment_type/CloudFormationDeploymentTypeParameters.scala
twrichards: or 
```
params
  .getOrElse(amiCfnParam, Map.empty)
  .get("Encrypted")
  .flatMap(_.toBooleanOption)
  .getOrElse(false)
```

agreed its certainly no more readable than what you had

akash1810: `toBooleanOption` made fixing a regression really nice, thanks! https://github.com/guardian/riff-raff/pull/1013

PR #1015, opened 2023-02-14
feat: Add `prefixStagePaths` parameter to `aws-s3` deployments

akash1810:
> **Note**
> Easiest to review [commit by commit](https://github.com/guardian/riff-raff/pull/1015/commits) as this change extracts some existing behaviour into a `trait` - [this specific change](https://github.com/guardian/riff-raff/pull/1015/commits/6b71973b00b5f2b04cb1c34e3381422a365e214f) should be considered a no-op.

## Background
In the `aws-s3` deployment type, the path within S3 that we upload to is controlled by the following parameters:
  - `prefixPackage`
  - `prefixStack`
  - `prefixApp`
  - `prefixStage`

This results in the path being the same regardless of the stage being deployed.

For example, for the file `image.jpg` and a _shared_ bucket the upload path will be something like `<bucket>/<STAGE>/image.jpg`. If the bucket was not shared across stages ([as recommended](https://github.com/guardian/recommendations/blob/main/AWS.md#s3)) the upload path will be something like `<bucket>/image.jpg`.

## What does this change?
Introduces a new `prefixStagePaths` parameter for the `aws-s3` deployment type.

This is to support services such as https://github.com/guardian/interactive-atom-maker. interactive-atom-maker is [configured](https://github.com/guardian/interactive-atom-maker/blob/983da014e0afe623f0c211f77a92f10a371fed2b/src/main/scala/app/Config.scala#L52) in a slightly non-standard way - CODE files live in `<bucket>/atoms-CODE/image.jpg`, PROD files live in `<bucket>/atoms/image.jpg`[^1]. The behaviour described above doesn't quite work here.

This change allows us to do:
```yaml
prefixStagePaths:
  - CODE: atoms-CODE
  - PROD: atoms
```

When set, `prefixStagePaths` takes precedence over all other `prefix<App|Stage|etc...>` options.

## How to test
Unit tests have been added to check:
- When `prefixStagePaths` is not set, we maintain existing behaviours of `prefixApp`, `prefixStage`, etc.
- What happens when `prefixStagePaths` is set
- When `prefixStagePaths` is set, describing stages of `CODE` and `PROD`, what happens when deploying stage of `UAT` (it should fail)
- When `prefixStagePaths` and `prefix*` are both set, what wins

For a practical demonstration of this new parameter, see https://github.com/guardian/csti-interactives/pull/53. To test, deploy this branch to CODE, then deploy a build from https://github.com/guardian/csti-interactives/pull/53, and observe the paths which Riff-Raff uploads files to.

https://github.com/guardian/csti-interactives (specifically https://github.com/guardian/csti-interactives/pull/53) is able to use Riff-Raff for continuous delivery.

[^1]: We _could_ update how interactive-atom-maker works, however this would involve updating all the articles that reference the files in S3, which is non-trivial.

### DIFF
--- a/magenta-lib/src/main/scala/magenta/deployment_type/S3ObjectPrefixParameters.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/S3ObjectPrefixParameters.scala
@@ -76,7 +76,7 @@ trait S3ObjectPrefixParameters {
         packageOrAppName = maybePackageOrAppName
       )
     } else {
-      prefixFromStagePaths.lift.apply(target.parameters.stage.name) match {
+      prefixFromStagePaths.get(target.parameters.stage.name) match {
         case Some(prefix) => prefix
         case _ =>
           reporter.fail(
--- a/magenta-lib/src/main/scala/magenta/deployment_type/S3ObjectPrefixParameters.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/S3ObjectPrefixParameters.scala
@@ -1,0 +1,53 @@
+package magenta.deployment_type
+
+import magenta.{DeployReporter, DeployTarget, DeploymentPackage}
+import magenta.tasks.S3Upload
+
+trait S3ObjectPrefixParameters {
+  this: DeploymentType =>
+
+  val prefixStage: Param[Boolean] =
+    Param("prefixStage", "Prefix the S3 bucket key with the target stage")
+      .default(true)
+
+  val prefixPackage: Param[Boolean] =
+    Param("prefixPackage", "Prefix the S3 bucket key with the package name")
+      .default(true)
+
+  val prefixStack: Param[Boolean] =
+    Param("prefixStack", "Prefix the S3 bucket key with the target stack")
+      .default(true)
+
+  val prefixApp: Param[Boolean] = Param[Boolean](
+    name = "prefixApp",
+    documentation = """
+        |Whether to prefix `app` to the S3 location instead of `package`.
+        |
+        |When `true` `prefixPackage` will be ignored and `app` will be used over `package`, useful if `package` and `app` don't align.
+        |""".stripMargin
+  ).default(false)
+
+  def getPrefix(
+      pkg: DeploymentPackage,
+      target: DeployTarget,
+      reporter: DeployReporter
+  ): String = {
+    val maybePackageOrAppName: Option[String] = (
+      prefixPackage(pkg, target, reporter),
+      prefixApp(pkg, target, reporter)
+    ) match {
+      case (_, true)      => Some(pkg.app.name)
+      case (true, false)  => Some(pkg.name)
+      case (false, false) => None
+    }
+
+    S3Upload.prefixGenerator(
+      stack =
+        if (prefixStack(pkg, target, reporter)) Some(target.stack) else None,
+      stage =
+        if (prefixStage(pkg, target, reporter)) Some(target.parameters.stage)
+        else None,
+      packageOrAppName = maybePackageOrAppName
+    )
+  }
+}
--- a/magenta-lib/src/main/scala/magenta/deployment_type/S3ObjectPrefixParameters.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/S3ObjectPrefixParameters.scala
@@ -83,13 +83,13 @@ trait S3ObjectPrefixParameters {
             s"""
                |Unable to locate prefix for stage ${target.parameters.stage.name}.
                |
-               |prefixFromStagePaths is set to:
+               |prefixStagePaths is set to:
                |
                |$prefixFromStagePaths
                |
                |To resolve, either:
                |  - Deploy to a known stage
-               |  - Update prefixFromStagePaths, adding a value for ${target.parameters.stage.name}
+               |  - Update prefixStagePaths, adding a value for ${target.parameters.stage.name}
                |""".stripMargin
           )
       }
--- a/magenta-lib/src/main/scala/magenta/deployment_type/S3ObjectPrefixParameters.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/S3ObjectPrefixParameters.scala
@@ -30,6 +30,15 @@ trait S3ObjectPrefixParameters {
   val prefixStagePaths: Param[Map[String, String]] = Param[Map[String, String]](
     "prefixStagePaths",
     documentation = """
+        |This option allows full control over the paths used for each stage, and will override all other prefix options.
+        |
+        |This is a useful escape hatch if you find yourself in a scenario where you are deploying to a single bucket and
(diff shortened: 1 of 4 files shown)


twrichards:
lol, I think I've just raised a very similar feature in https://github.com/guardian/riff-raff/pull/1033 without seeing this

twrichards:
When might this PR get finished / merged? I'd like to use it for https://github.com/guardian/actions-static-site/pull/19#issuecomment-1446010643

Happy to help. Is it just tests which need adding?

magenta-lib/src/test/scala/magenta/deployment_type/S3ObjectPrefixParametersTest.scala
```diff
@@ -32,6 +24,19 @@ class S3ObjectPrefixParametersTest extends AnyFlatSpec with Matchers {
...
     Region("eu-west-1")
   )
+
+  val targetStageCODE = DeployTarget(
+    fixtures.parameters(stage = Stage("CODE")),
+    Stack("interactives"),
+    Region("eu-west-1")
+  )
+
+  val targetStageUnknown = DeployTarget(
```
akash1810, line 34: This `val` is only used in one test, should we co-locate it with that test?

magenta-lib/src/test/scala/magenta/deployment_type/S3ObjectPrefixParametersTest.scala
```diff
@@ -62,15 +67,77 @@ class S3ObjectPrefixParametersTest extends AnyFlatSpec with Matchers {
...
+         |Unable to locate prefix for stage ${targetStageUnknown.parameters.stage.name}.
+         |
+         |prefixFromStagePaths is set to:
+         |
+         |$prefixFromStagePaths
+         |
+         |To resolve, either:
+         |  - Deploy to a known stage
+         |  - Update prefixFromStagePaths, adding a value for ${targetStageUnknown.parameters.stage.name}
+         |""".stripMargin
```
akash1810, line 139: When used as a test assertion, I'd recommend using the raw string over an interpolated one as it is more explicit/clear what the expected message is, thus easier to sense check the wording.

magenta-lib/src/test/scala/magenta/deployment_type/S3ObjectPrefixParametersTest.scala
ob6160: good shout, will do!

magenta-lib/src/test/scala/magenta/deployment_type/S3ObjectPrefixParametersTest.scala
ob6160: makes sense 👍  I usually go for concrete values in tests, but slipped my mind this time

magenta-lib/src/main/scala/magenta/deployment_type/S3ObjectPrefixParameters.scala
```diff
@@ -0,0 +1,89 @@
...
+  val prefixStagePaths: Param[Map[String, String]] = Param[Map[String, String]](
+    "prefixStagePaths",
+    documentation = """
+        |```yaml
+        |prefixStagePaths:
+        |  CODE: atoms-CODE
+        |  PROD: atoms
+        |```
+        |""".stripMargin
+  ).default(Map.empty)
```
akash1810, line 39: TODO add more documentation, describing how/when to use this parameter.

magenta-lib/src/main/scala/magenta/deployment_type/S3ObjectPrefixParameters.scala
```diff
@@ -0,0 +1,98 @@
...
+          if (prefixStack(pkg, target, reporter)) Some(target.stack)
+          else None,
+        stage =
+          if (prefixStage(pkg, target, reporter))
+            Some(target.parameters.stage)
+          else None,
+        packageOrAppName = maybePackageOrAppName
+      )
+    } else {
+      prefixFromStagePaths.lift.apply(target.parameters.stage.name) match {
```
kenoir, line 79:
```suggestion
      prefixFromStagePaths.get(target.parameters.stage.name) match {
```

magenta-lib/src/main/scala/magenta/deployment_type/S3ObjectPrefixParameters.scala
```diff
@@ -0,0 +1,98 @@
...
+      case (_, true)      => Some(pkg.app.name)
+      case (true, false)  => Some(pkg.name)
+      case (false, false) => None
+    }
+
+    val prefixFromStagePaths: Map[String, String] =
+      prefixStagePaths(pkg, target, reporter)
+
+    if (prefixFromStagePaths.isEmpty) {
+      S3Upload.prefixGenerator(
```
kenoir, line 68: It feels a little strange that some of the logic for generating a prefix is in `S3Upload.prefixGenerator` and some of it here, should this be consolidated?

magenta-lib/src/test/scala/magenta/deployment_type/S3ObjectPrefixParametersTest.scala
```diff
@@ -0,0 +1,163 @@
...
+    extends DeploymentType
+    with S3ObjectPrefixParameters {
+  def defaultActions = Nil
+
+  def documentation = "n/a"
+
+  def name = "n/a"
+}
+
+class S3ObjectPrefixParametersTest extends AnyFlatSpec with Matchers {
```
kenoir, line 30: 👏 well tested, very nice.

magenta-lib/src/main/scala/magenta/deployment_type/S3ObjectPrefixParameters.scala
akash1810: `S3Upload.prefixGenerator` is [used](https://github.com/search?q=repo%3Aguardian%2Friff-raff+%22S3Upload.prefixGenerator%22&type=code) in a few places. Suggest we further refactor/DRY out in a follow-up PR to keep the diff in this PR to a minimum.

magenta-lib/src/main/scala/magenta/deployment_type/S3ObjectPrefixParameters.scala
ob6160: what's the difference between `lift.apply` vs `get` here?

akash1810: In this case, they are functionally equivalent (as demonstrated in [this Scastie](https://scastie.scala-lang.org/1JMH56AQTbyqn0jYb4krHQ)). `lift` is a [partial function](https://alvinalexander.com/scala/how-to-define-use-partial-functions-in-scala-syntax-examples/), Using `get` is more obvious/readable/maintainable here.

kenoir (approved):

PR #1033, opened 2023-02-27
add `explicitS3Location` and `explicitS3LocationByStage` params to `aws-s3` deployment type

twrichards:
… to support varying S3 location by Stage (and to avoid hacks which force S3 locations by setting app to the value and using with `prefixApp: true`, `prefixStack: false` and `prefixStage: false`).


twrichards:
closing, as just realised there's already https://github.com/guardian/riff-raff/pull/1033 in-flight which is nicer

PR #1037, opened 2023-03-03
ci: Use custom `sbt` script that configures memory

akash1810:
## What does this change?
This is an attempt to resolve the OOM errors we sometimes see in CI. The [`sbt` script in this repo](https://github.com/guardian/riff-raff/blob/1e1ca44746c4876fea1d58d36689775ca165dd2f/sbt) customises the amount of memory available.

Example of OOM seen in CI:
```log
[error] Caught java.lang.OutOfMemoryError: Metaspace
[error] To best utilize classloader caching and to prevent file handle leaks, we recommend running sbt without a MaxMetaspaceSize limit. 
[error] 
java.lang.OutOfMemoryError: Metaspace
[error] java.lang.OutOfMemoryError: Metaspace
```

A more stable build?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,4 +41,4 @@ jobs:
           export GITHUB_RUN_NUMBER=$(( $GITHUB_RUN_NUMBER + $LAST_TEAMCITY_BUILD ))
 
           # Note, lib/test because riffraff/riffRaffUpload only tests the riffraff project.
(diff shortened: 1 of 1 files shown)


kenoir:
Suggest we test this by re-running the build a few times:
Screenshot 2023-03-03 at 10 00 46

kenoir:
Pretty good track record:
Screenshot 2023-03-03 at 10 52 37

PR #1082, opened 2023-03-29
skip scheduled deploys where last deploy was 'partial' to prevent unexpected/problematic behaviour

twrichards:
Recently, when @guardian/digital-cms were working on https://github.com/guardian/editorial-tools-production-monitoring/pull/316, they needed to do partial deploys (in order to test on secondary instances in PROD, but leave primary instances unaffected), meanwhile riff-raff performed the scheduled deploy for this project, **crucially a full deploy**, which in turn broke primary instances, this was quickly dealt with but perhaps if someone forgets that they did a partial deploy, then the current behaviour of riff-raff could cause some quite pernicious problems.

So now, if last deploy for a project was 'partial' then the scheduled deployed is skipped but users are notified (just like when previous deploy failed, as most likely the partial deploy was temporary).

### DIFF
--- a/riff-raff/app/schedule/DeployJob.scala
+++ b/riff-raff/app/schedule/DeployJob.scala
@@ -42,6 +42,9 @@ class DeployJob extends Job with Logging {
 
     attemptToStartDeploy match {
       case Left(error: ScheduledDeployNotificationError) =>
+        log.info(
+          s"Scheduled deploy failed to start due to $error. A notification will be sent..."
+        )
         scheduledDeployNotifier.scheduledDeployFailureNotification(error)
       case Left(anotherError) =>
         log.warn(
--- a/riff-raff/app/schedule/DeployJob.scala
+++ b/riff-raff/app/schedule/DeployJob.scala
@@ -3,6 +3,7 @@ package schedule
 import controllers.Logging
 import deployment._
 import magenta.Strategy.MostlyHarmless
+import magenta.input.DeploymentKeysSelector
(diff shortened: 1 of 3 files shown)


riff-raff/app/schedule/DeployJob.scala
```diff
@@ -61,7 +62,11 @@ object DeployJob extends Logging with LogAndSquashBehaviour {
     lastDeploy.state match {
       case RunState.Completed =>
         val params = extractDeployParameters(lastDeploy)
-        if (scheduledDeploysEnabled) {
+        if (
+          lastDeploy.parameters.selector.isInstanceOf[DeploymentKeysSelector]
+        ) {
+          Left(SkippedDueToPreviousPartialDeploy(lastDeploy))
```
nicl, line 68: Not caused by this PR, but would it be okay to add a log line for the bit this hits on lines 44-45? To indicate a notification is sent for a failed scheduled deploy.

I.e. at this bit:

```
      case Left(error: ScheduledDeployNotificationError) =>
        scheduledDeployNotifier.scheduledDeployFailureNotification(error)
```

above. Shout if not obvious which bit I mean.

nicl (approved):
Nice!

riff-raff/app/schedule/DeployJob.scala
twrichards: done in https://github.com/guardian/riff-raff/commit/f6fa8dff5ee9982ffa3d52d0a5a31e4dde4f3505

PR #1083, opened 2023-03-29
Don't expose branch regexs for CD, just names

tjmw:
## What does this change?

Looking at projects with CD configured, it seems that the only type of configuration we use is to specify a branch name, which gets converted to a regex (by adding start and end anchors). Let's simplify the UI and make it clear you can enter a branch name, and expose that name back to users, rather than the constructed regex. I've gotten confused in the past and added the `^` and `$` manually, which isn't necessary (I see a few instances of `^^main$$` in the CD configurations, so I don't think I'm the only one!)

#### Before

Displays branches as regexes:

Screenshot 2023-03-31 at 12 41 08

#### After

Displays branch names:

Screenshot 2023-03-31 at 12 39 10

#### Before

Asks the user for a branch regex:

Screenshot 2023-03-31 at 12 41 18

#### After

Asks the user for a branch name:

Screenshot 2023-03-31 at 12 39 43

## Accessibility

### DIFF
--- a/riff-raff/app/views/continuousDeployment/list.scala.html
+++ b/riff-raff/app/views/continuousDeployment/list.scala.html
@@ -36,7 +36,7 @@
                 <tr @if(config.trigger == Disabled){ class="danger" }>
                     <td>@utils.DateFormats.Short.print(config.lastEdited) by <span class="label label-default">@config.user</span></td>
                     <td>@config.projectName</td>
-                    <td>@config.branchRE.toString()</td>
+                    <td>@config.branchMatcher.getOrElse("*")</td>
                     <td>@config.stage</td>
                     <td>@config.trigger match {
                         case SuccessfulBuild => { Build succeeds }
--- a/riff-raff/app/views/continuousDeployment/list.scala.html
+++ b/riff-raff/app/views/continuousDeployment/list.scala.html
@@ -24,7 +24,7 @@
                 <tr>
                     <th>Last edited</th>
                     <th>Project Name</th>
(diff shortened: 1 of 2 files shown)


akash1810:
Is this worth some before and after screenshots?

tjmw:
> Is this worth some before and after screenshots?

Great idea @akash1810 👍 I'll grab some.

tjmw:
@akash1810 I've added some screenshots.

riff-raff/app/views/continuousDeployment/list.scala.html
```diff
@@ -36,7 +36,7 @@ <h4 class="alert-heading">Change Freeze</h4>
                 <tr @if(config.trigger == Disabled){ class="danger" }>
                     <td>@utils.DateFormats.Short.print(config.lastEdited) by <span class="label label-default">@config.user</span></td>
                     <td>@config.projectName</td>
-                    <td>@config.branchRE.toString()</td>
+                    <td>@config.branchMatcher.getOrElse("*")</td>
```
akash1810, line 39: `*` as a default feels a bit risky; presumably it means _any_ branch will be continuously deployed?

tjmw: But yes I agree that this seems risky! Perhaps in a future PR this could become a required field?

tjmw:
@akash1810 thanks for bumping this! I didn't merge before as I was about to go on holiday. I'll look into the CI failure and hopefully get this merged :-)

PR #1134, opened 2023-05-02
Add new Compute@Edge deployment type

marsavar:
TBC

### DIFF
--- a/magenta-lib/src/main/scala/magenta/deployment_type/FastlyCompute.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/FastlyCompute.scala
@@ -1,0 +1,37 @@
+package magenta.deployment_type
+
+import magenta.KeyRing
+import magenta.tasks.UpdateFastlyPackage
+import software.amazon.awssdk.services.s3.S3Client
+
+object FastlyCompute extends DeploymentType {
+  val name = "fastly-compute"
+
+  val documentation: String =
+    """
+      |Deploy a [Compute@Edge](https://www.fastly.com/products/edge-compute) package via the Fastly API.
+    """.stripMargin
+
+  val deploy: Action = Action(
+    "deploy",
+    """
+      |Undertakes the following using the Fastly API:
+      |
+      | - Clone the currently active version
+      | - Uploads the package
+      |
+      |Note that `your-service` must match the deployment resource name
+      | ```
+      |  your-service:
+      |   type: fastly-compute-edge
+      |```
+    """.stripMargin
+  ) { (pkg, resources, target) =>
+    implicit val keyRing: KeyRing = resources.assembleKeyring(target, pkg)
+    implicit val artifactClient: S3Client = resources.artifactClient
+    resources.reporter.verbose(s"Keyring is $keyRing")
+    List(UpdateFastlyPackage(pkg.s3Package)(keyRing, artifactClient))
+  }
+
+  def defaultActions: List[Action] = List(deploy)
+}
--- a/magenta-lib/src/main/scala/magenta/tasks/FastlyComputeTasks.scala
+++ b/magenta-lib/src/main/scala/magenta/tasks/FastlyComputeTasks.scala
@@ -1,0 +1,126 @@
+package magenta.tasks
+
+import java.util.concurrent.Executors
+import com.gu.fastly.api.FastlyApiClient
+import magenta._
+import magenta.artifact.S3Path
+import play.api.libs.json.Json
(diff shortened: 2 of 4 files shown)


marsavar:
Apologies for the many commits, will squash and merge when approved :-)

philmcmahon (commented):
This looks excellent, nice work Mario

hanging out on the edge

magenta-lib/src/main/scala/magenta/deployment_type/FastlyCompute.scala
```diff
@@ -0,0 +1,102 @@
...
+      |
+      |Example:
+      |
+      |```yaml
+      |your-service:
+      |  type: fastly-compute
+      |```
+      |Note that `your-service` must match the App name in Deployment Resources under `credentials:fastly`.
+      |It will appear there once it has been added to `prism-data.json`, which lives in the `prism-data`
+      |bucket within the `deployTools` account.
```
akash1810, line 30: This level of detail is excellent, though given it's not available on other deployment types, I wonder if it should be here, as, for example, it's disclosing bucket names.

akash1810 (approved):
Wonderful! Thanks for doing this.

Added a couple of non-blocking comments, but overall this is brilliant!

Wonder if it's worth adding support in https://github.com/guardian/cdk/tree/main/src/experimental/riff-raff-yaml-file?

magenta-lib/src/main/scala/magenta/deployment_type/FastlyCompute.scala
marsavar: That's a good point, and I thought about it before adding it to the docs. However, I wouldn't have figured this out if @philmcmahon hadn't told me about it - I couldn't find it in the Riff-Raff documentation. 
I'm just trying to think what would be a  good compromise: we don't want to disclose private bucket names, but at the same time we don't want people to read the docs and think "How do I add my service to Deployment Resources"? 
Do you think it's acceptable to write something along the lines of "If you're not sure how to do this, contact the DevX team"?

magenta-lib/src/main/scala/magenta/tasks/FastlyComputeTasks.scala
```diff
@@ -0,0 +1,185 @@
...
+        // overridden in order to convert binary to Latin-1, hence ISO-8859-1.
+        // This will allow us to send a valid HTTP request to the Fastly API
+        val bufferedSource =
+          scala.io.Source.fromInputStream(stream)(ISO8859)
+        val bytes =
+          bufferedSource.mkString.getBytes(StandardCharsets.ISO_8859_1)
+        Files
+          .write(
+            Files
+              .createTempFile(fileName.replace(".tar.gz", ""), ".tar.gz"),
```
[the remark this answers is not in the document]
marsavar: `fileName` is a string which includes the extension (eg `package-name.tar.gz`), and `createTempFile`'s constructor takes a prefix and a suffix. If I don't remove it, the file will be called `package-name.tar.gz.tar.gz`. 
I don't think I need to do this at all though, since both `prefix` and `suffix` may be `null`, but it just looks nicer in the logs...
image

magenta-lib/src/main/scala/magenta/deployment_type/FastlyCompute.scala
akash1810: I wonder if https://github.com/guardian/deploy-tools-platform/blob/main/cloudformation/riffraff/riff-raff-roles.md is the place for this information?

magenta-lib/src/main/scala/magenta/tasks/FastlyComputeTasks.scala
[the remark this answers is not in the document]
akash1810: Ah! Might be worth adding this as an inline comment here?

magenta-lib/src/main/scala/magenta/deployment_type/FastlyCompute.scala
marsavar: It could well be, but I still worry about discoverability. I had no idea that repo existed until today 😨 
I'll just remove the reference to private buckets for now - I think the "Add your App to Deployment Resources" is probably good enough in the sense that it provides at the very least some direction.

marsavar: @philmcmahon pointed out I could link to that file - I had some reservations about linking to a private repo from a public one but imho it's better than not saying anything!

PR #1141, opened 2023-05-10
Warn if an AMI cannot be resolved during cloudformation deploy

davidfurey:
## What does this change?

Displays a warning if an AMI cannot be resolved during a cloudformation deploy. For now this is a warning to see how often this happened, but I think we should upgrade this to an error

## How to test

Deploy to CODE. Attempt to deploy a project that has a missing AMI.


jacobwinch (approved):

davidfurey:
Tested on CODE:
Screenshot 2023-05-12 at 14 05 55

PR #1148, opened 2023-05-16
Fix false positives in `fastly` and `fastly-compute` deployment types

marsavar:
## What does this change?

Fixes an issue with `fastly` and `fastly-compute` deployment types that incorrectly deploy successfully when no API token is present in the keyring.

This is a problem because you might deploy your VCL or Compute@Edge application and think everything has gone smoothly, when in reality nothing has happened because the `execute` function wasn't triggered.

In its current implementation, `foreach` is called on the optional `FastlyApiClientProvider.get(keyring)` function, meaning that `execute` will only run if it is a `Some` value. No implementation is defined when it's `None`.

## How to test

Deployed to CODE Riff-Raff.

### Scenario 1: Fastly API key not present in the parameter store (should fail)

image

### Scenario 2: Invalid or expired Fastly API key present in the parameter store (should fail)
image

### Scenario 3: Valid Fastly API key present in the parameter store (should succeed)
image


magenta-lib/src/main/scala/magenta/tasks/FastlyComputeTasks.scala
```diff
@@ -27,34 +27,49 @@ case class FastlyComputeTasks(s3Package: S3Path)(implicit
...
-      uploadPackage(
-        nextVersionNumber,
-        s3Package,
-        client,
-        resources.reporter,
-        stopFlag
-      )
+    FastlyApiClientProvider.get(keyRing) match {
+      case None =>
+        resources.reporter.fail("Failed to fetch Fastly API credentials")
```
nicl, line 32: Might be worth expanding this slightly to give a clearer sense of where credentials should exist.

nicl (approved):
Nice!

Small code style advice (non-blocking though): I'd probably favour pulling the `case Some(client) => ... ` body out into a function to improve readability - as easier to see the actual patterns being matched. And have the function return a `Try[..]` rather than a large try/catch - again as it makes it easier to scan quickly the actual flow. But this is somewhat subjective.

marsavar:
This was closed by mistake - reopening it and refactored it with Nic's suggestion

marsavar:
> This PR is stale because it has been open 30 days with no activity. Unless a comment is added or the “stale” label removed, this will be closed in 3 days

🥹

magenta-lib/src/main/scala/magenta/tasks/FastlyTasks.scala
```diff
@@ -30,43 +31,73 @@ case class UpdateFastlyConfig(s3Package: S3Path)(implicit
...
-
-      resources.reporter
-        .info(s"Fastly version $nextVersionNumber is now active")
+  override def execute(
+      resources: DeploymentResources,
+      stopFlag: => Boolean
+  ): Unit = {
+    FastlyApiClientProvider.get(keyRing) match {
+      case None =>
+        resources.reporter.fail("Failed to fetch Fastly API credentials")
```
davidfurey, line 90:
```suggestion
        resources.reporter.fail("Failed to fetch Fastly API credentials from keyring")
```

magenta-lib/src/main/scala/magenta/tasks/FastlyTasks.scala
```diff
@@ -30,43 +31,73 @@ case class UpdateFastlyConfig(s3Package: S3Path)(implicit
...
+      case Some(client) =>
+        _execute(client, resources, stopFlag) match {
+          case Success(nextVersionNumber) =>
+            resources.reporter
+              .info(s"Fastly version $nextVersionNumber is now active")
+          case Failure(err) =>
+            resources.reporter.fail(
+              s"$err. Your API key may be invalid or it may have expired"
+            )
+        }
```
davidfurey, line 100: I think this might be slightly easier to read:
```suggestion
    override def execute(
      resources: DeploymentResources,
      stopFlag: => Boolean
  ): Unit = {
    FastlyApiClientProvider
      .get(keyRing)
      .map(_execute(_, resources, stopFlag)) match {
      case Some(Success(nextVersionNumber)) =>
        resources.reporter
          .info(s"Fastly version $nextVersionNumber is now active")
      case Some(Failure(err)) =>
        resources.reporter.fail(
          s"$err. Your API key may be invalid or it may have expired"
        )
      case None =>
        resources.reporter.fail("Failed to fetch Fastly API credentials")
    }
  }
```

magenta-lib/src/main/scala/magenta/tasks/FastlyTasks.scala
```diff
@@ -30,43 +31,73 @@ case class UpdateFastlyConfig(s3Package: S3Path)(implicit
   implicit val ec: ExecutionContextExecutorService =
     ExecutionContext.fromExecutorService(Executors.newFixedThreadPool(10))
 
-  override def execute(
+  private def _execute(
```
davidfurey, line 34: Is there a more meaningful name for this method? Perhaps `executeWithCredentials`?

magenta-lib/src/main/scala/magenta/tasks/FastlyTasks.scala
```diff
@@ -30,43 +31,73 @@ case class UpdateFastlyConfig(s3Package: S3Path)(implicit
...
-        stopFlag
       )
-
-      uploadNewVclFilesTo(
-        nextVersionNumber,
-        s3Package,
-        client,
-        resources.reporter,
-        stopFlag
+      _ <- Try(
```
davidfurey, line 46: These might be slightly easier to read if the `Try` wrapper is added in the private methods rather than here

magenta-lib/src/main/scala/magenta/tasks/FastlyComputeTasks.scala
```diff
@@ -23,38 +24,69 @@ case class FastlyComputeTasks(s3Package: S3Path)(implicit
...
-        stopFlag
-      )
-
-      activateVersion(nextVersionNumber, client, resources.reporter, stopFlag)
-
-      resources.reporter
-        .info(
-          s"Fastly Compute@Edge service ${client.serviceId} - version $nextVersionNumber is now active"
+    FastlyApiClientProvider.get(keyRing) match {
+      case None =>
```
davidfurey, line 74: Similar flattening of pattern match could be done as for FastlyTask.scala

davidfurey (commented):
I've made a few suggestions, but this already looks like a good change that would be great to get merged.

magenta-lib/src/main/scala/magenta/tasks/FastlyTasks.scala
marsavar: I ended up removing the for comprehension and just wrapped the entire block in a `Try`. I think it achieves the same effect with fewer code changes =)

magenta-lib/src/main/scala/magenta/tasks/FastlyComputeTasks.scala
marsavar: Thanks - yes, I'm not a big fan of nested pattern matching either. Have refactored accordingly :)

magenta-lib/src/main/scala/magenta/tasks/FastlyTasks.scala
davidfurey: Great, much less distracting to read as well

davidfurey (approved):
Looks good to me. Might be nice to have a 👍 from a current member of devX?

magenta-lib/src/main/scala/magenta/tasks/FastlyComputeTasks.scala
```diff
@@ -49,12 +51,40 @@ case class FastlyComputeTasks(s3Package: S3Path)(implicit
...
+          "Failed to fetch Fastly API credentials from keyring"
+        )
+      case Some((client, Success(nextVersionNumber))) =>
+        resources.reporter
+          .info(
+            s"Fastly Compute@Edge service ${client.serviceId} - version $nextVersionNumber is now active"
+          )
+      case Some((_, Failure(err))) =>
+        resources.reporter.fail(
+          s"$err. Your API key may be invalid or it may have expired"
```
jacobwinch, line 85: Minor: 

IIUC, this case is going to catch a number of potential errors (network problems, Fastly API downtime/errors) as well as the problems that we've already mentioned in the error message. 

If we find reliability problems with this deployment type in the future, it might be worth spending a bit of extra time to extract the specific error (perhaps by adding some verbose logging, or a different return type to [`activateVersion`](https://github.com/guardian/riff-raff/blob/1522cee3f148bb279099f686fa62afcf18383f4f/magenta-lib/src/main/scala/magenta/tasks/FastlyComputeTasks.scala#L141-L153)), but I don't think that needs to block this PR.

jacobwinch (approved):
Great work; thanks for improving this and documenting the testing so clearly 👍

magenta-lib/src/main/scala/magenta/tasks/FastlyComputeTasks.scala
jacobwinch: Thinking about this some more, I think there is another potential false positive here. IIUC calling [`activateVersion`](https://github.com/guardian/riff-raff/blob/1522cee3f148bb279099f686fa62afcf18383f4f/magenta-lib/src/main/scala/magenta/tasks/FastlyComputeTasks.scala#L141-L153) can result in a successful `Future` (i.e. we got _a_ `Response` from the Fastly API), where the response/status code indicates that the request was unsuccessful from Fastly's perspective (i.e. we got a 500 back and did not actually activate the new version). In this case we will still match on:  ```case Some((client, Success(nextVersionNumber)))```.

Happy to pair on fixing that via a separate PR if helpful.

PR #1174, opened 2023-06-08
Use info instead of error alert style when project never deployed

georgeblahblah:
## What does this change?

- Changes the visual style of the `Riff Raff has not been used to deploy {{projectName}} to {{stage}}` alert from error to info, as I don't think this is an error state 

## Screenshots

| Before | After |
|--------|--------|
| image | image |

### DIFF
--- a/riff-raff/app/views/deploy/deployHistory.scala.html
+++ b/riff-raff/app/views/deploy/deployHistory.scala.html
@@ -16,9 +16,9 @@
     <h4>Last deploys of <em>@projectName</em> @maybeStage.map{ stage => to @stage }</h4>
     @snippets.recordTable(records, None, allColumns = false, popOutLink = true)
 } else {
-    <div class="alert alert-danger" role="alert">
+    <div class="alert alert-warning" role="alert">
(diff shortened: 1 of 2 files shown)


nicl (approved):
Happy to try this and see how people respond, though worth getting one or two more thoughts before merging I think. I agree that the red can be offputting and unnecessary sometimes, but at other times it is helpful to realise you are possibly deploying a project to a Stage that it shouldn't be deployed too. (`allowedStages` is now a good solution to this though that teams should be using.)

jacobwinch:
> I agree that the red can be offputting and unnecessary sometimes, but at other times it is helpful to realise you are possibly deploying a project to a Stage that it shouldn't be deployed too.

Yes, I agree. I think it depends on whether we think that it's more common for people to see this screen when a) they are intentionally deploying a project to a new stage for the first time (e.g. when creating a new project) or b) they have chosen the wrong stage from the long drop-down and are about to deploy a duplicate version of their stack accidentally(!).

I'm also happy to try this out as I think we now prefer usage of `allowedStages` to avoid scenario b!

PR #1207, opened 2023-07-17
Add S3 object tags to Lambda artifacts

jacobwinch:
## What does this change?

This PR updates Riff-Raff so that it [tags](https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-tagging.html) Lambda artifacts when it uploads them to S3. 

Lambda artifacts are only needed temporarily[^1] and adding these tags will allow us to more easily cleanup unused Lambda artifacts (using lifecycle rules) in the future:

> Object tags enable fine-grained object lifecycle management in which you can specify a tag-based filter, in addition to a key name prefix, in a lifecycle rule.

The main motivation for adding this capability now is the fact that some teams are planning to start using Lambda versioning. This will mean that we have a lot more Lambda artifacts in our distribution buckets (see https://github.com/guardian/cdk/pull/1925 for more detail).

This change relies on an additional IAM permission which has been added via a separate PR: https://github.com/guardian/riff-raff/pull/1207.

## How to test

Unit tests have been added (and updated) where relevant.

I have [deployed this change to `CODE`](https://riffraff.code.dev-gutools.co.uk/deployment/view/6c7fa932-d693-4567-b561-d37873f23b60) and [successfully deployed a project with uses the `lambda` and `autoscaling` deployment types](https://riffraff.code.dev-gutools.co.uk/deployment/view/bc709355-8482-4415-bdcc-8be244e800eb). 

The Lambda artifact has the expected tag:

image

The EC2 artifact has no tags (as expected):

image

We will be able to implement appropriate retention policies for these artifacts once this PR has been merged.

This PR doesn't introduce any particular risks but implementing the lifecycle rules which delete stuff based on these tags will. I have contacted AWS to confirm that there is no need to retain Lambda code in our own S3 buckets, they said:

> It is safe to delete the deployment package from your S3 bucket once the Lambda function code is successfully update after uploading from your S3 bucket and it will not impact your Lambda function's invocations.

[^1]: AWS manages Lambda code via some [internal storage](https://docs.aws.amazon.com/lambda/latest/operatorguide/code-storage.html), so we only need to temporarily store files in our own distribution buckets whilst the deployment takes place.

### DIFF
--- a/magenta-lib/src/main/scala/magenta/tasks/tasks.scala
+++ b/magenta-lib/src/main/scala/magenta/tasks/tasks.scala
@@ -42,7 +42,7 @@ case class S3Upload(
     surrogateControlPatterns: List[PatternValue] = Nil,
     extensionToMimeType: Map[String, String] = Map.empty,
     publicReadAcl: Boolean = false,
-    lambdaArtifact: Boolean = false,
+    allowDeletionByLifecycleRule: Boolean = false,
     detailedLoggingThreshold: Int = 10
 )(implicit
     val keyRing: KeyRing,
@@ -72,7 +72,7 @@ case class S3Upload(
       surrogateControlLookup(target.key),
       contentTypeLookup(target.key),
       publicReadAcl = publicReadAcl,
-      lambdaArtifact = lambdaArtifact
+      allowDeletionByLifecycleRule = allowDeletionByLifecycleRule
     )
   }
 
@@ -234,7 +234,7 @@ case class PutReq(
     surrogateControl: Option[String],
     contentType: Option[String],
     publicReadAcl: Boolean,
-    lambdaArtifact: Boolean
+    allowDeletionByLifecycleRule: Boolean
 ) {
   import collection.JavaConverters._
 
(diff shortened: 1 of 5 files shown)


magenta-lib/src/main/scala/magenta/tasks/tasks.scala
```diff
@@ -42,6 +42,7 @@ case class S3Upload(
     surrogateControlPatterns: List[PatternValue] = Nil,
     extensionToMimeType: Map[String, String] = Map.empty,
     publicReadAcl: Boolean = false,
+    lambdaArtifact: Boolean = false,
```
akash1810, line 45: The enclosing function is used by multiple deployment types; I wonder if we should be more generic in the naming here? For example `allowDeletionByLifecycleRule`. Maybe the tag could be generic too?

akash1810 (approved):
Minor comments that don't necessarily block.

magenta-lib/src/main/scala/magenta/tasks/tasks.scala
jacobwinch: Thanks for the suggestion, I think that's a good idea! https://github.com/guardian/riff-raff/pull/1207/commits/d002ee4f2feccc753b7c6c2c2b31f92f5e952767

PR #1212, opened 2023-07-21
Order builds by start time, not by build ID

undergroundquizscene:
With Github Actions, newer builds can have smaller IDs. Therefore Riff Raff should order the builds by their start time for CI: recently there was a bug where continuous deployment deployed an older build twice and didn’t deploy the newest build at all.

The implicit ordering on `CIBuild` was the cause: it’s used in `ContinuousDeployment.buildCandidates` by `GreatestSoFar` to order the builds produced by the poller, and currently orders by build ID. This change uses the build start time instead.

I’ve also changed the ordering used in `Builds.successfulBuilds` to match: I don’t think this is currently used anywhere, but it’s good to be consistent in case it’s used in the future.

Writing a meaningful test of `ContinuousDeployment.buildCandidates` to test the ordering seems like a lot of effort, so I haven’t done it. Hopefully the more basic test added here is useful.

### DIFF
--- a/riff-raff/app/ci/Builds.scala
+++ b/riff-raff/app/ci/Builds.scala
@@ -4,6 +4,7 @@ import akka.agent.Agent
 import ci.Context._
 import controllers.Logging
 import lifecycle.Lifecycle
+import org.joda.time.DateTime
 import rx.lang.scala.Subscription
 
 class Builds(ciBuildPoller: CIBuildPoller) extends Lifecycle with Logging {
@@ -33,7 +34,9 @@ class Builds(ciBuildPoller: CIBuildPoller) extends Lifecycle with Logging {
   val buildsAgent = Agent[Set[CIBuild]](BoundedSet(100000))
   val jobsAgent = Agent[Set[Job]](Set())
   def successfulBuilds(jobName: String): List[CIBuild] =
-    all.filter(_.jobName == jobName).sortBy(-_.id)
+    all
+      .filter(_.jobName == jobName)
+      .sortBy(_.startTime)(Ordering[DateTime].reverse)
   def getLastSuccessful(jobName: String): Option[String] =
     successfulBuilds(jobName).headOption.map { latestBuild =>
       latestBuild.number
--- a/riff-raff/app/ci/CIBuild.scala
+++ b/riff-raff/app/ci/CIBuild.scala
@@ -25,7 +25,8 @@ trait CIBuild {
 }
 
 object CIBuild {
-  implicit val ord = Ordering.by[CIBuild, Long](_.id)
+  implicit val ord: Ordering[CIBuild] =
+    Ordering.by[CIBuild, DateTime](_.startTime)
 }
 
 trait Job {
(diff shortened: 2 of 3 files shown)


nicl (commented):
@undergroundquizscene can you expand on why GHA should be providing lower IDs for a project? For individual projects, teams should avoid this as it is confusing. And we provide specific mechanisms to avoid this in our tooling - e.g.

https://github.com/guardian/actions-riff-raff#buildnumberoffset

undergroundquizscene:
> @undergroundquizscene can you expand on why GHA should be providing lower IDs for a project? For individual projects, teams should avoid this as it is confusing. And we provide specific mechanisms to avoid this in our tooling - e.g.
> 
> https://github.com/guardian/actions-riff-raff#buildnumberoffset

This was something we ran into when two PRs were merged shortly after one another – it’s not related to different build numbers on GHA and TeamCity.

On support-frontend, we merged two PRs shortly after one another, giving commits [46a68ee](https://github.com/guardian/support-frontend/commit/46a68eef6257ba272397950f9cac40f9c4843db4) and [e8d21af](https://github.com/guardian/support-frontend/commit/e8d21af6cda0cab4aebfd63db9cffeef6b4a8a5f) – you can see this is the order we merged them in by seeing the former is a parent of the latter on main. However, the former commit got GHA run number [3842](https://github.com/guardian/support-frontend/actions/runs/5375815121/job/14547341610), while the latter got number [3841](https://github.com/guardian/support-frontend/actions/runs/5375802754/job/14547311024): the newer commit had a lower GHA run number and therefore a lower Riff Raff build number. I’m guessing this is because the GHA run numbers are assigned when the job is picked up by an agent, not when it’s queued.

nicl:
ah thanks @undergroundquizscene . I wonder if this is a GHA bug or intended behaviour on their part. I've raised a support ticket (https://support.github.com/ticket/personal/0/2272918) to ask about this. Is it okay if we hold off on this until we hear back. I can't see anything about this in their docs unfortunately.

undergroundquizscene:
Yup, no problem at all :)

nicl:
@undergroundquizscene hey Github responded and it turns out that the way to achieve this is via the concurrency config:

https://docs.github.com/en/enterprise-cloud@latest/actions/using-workflows/workflow-syntax-for-github-actions#concurrency

I think we should probably do some comms on that but I think that's a better solution than this, as non-numerical ordering in RR makes things like rollback quite confusing and would break a long-standing convention on build orders.

So I'm going to close this PR but please do re-open/raise with DevX Ops if you think this is worth revisiting. (Today is my last day so I won't be around to discuss anywmore 😢 .)

undergroundquizscene:
> @undergroundquizscene hey Github responded and it turns out that the way to achieve this is via the concurrency config:
> 
> https://docs.github.com/en/enterprise-cloud@latest/actions/using-workflows/workflow-syntax-for-github-actions#concurrency
>
> I think we should probably do some comms on that but I think that's a better solution than this, as non-numerical ordering in RR makes things like rollback quite confusing and would break a long-standing convention on build orders.

Ah ok, perfect! Nice to be able to preserve the build ordering, it’s a good property to be able to rely on.

> So I'm going to close this PR but please do re-open/raise with DevX Ops if you think this is worth revisiting. (Today is my last day so I won't be around to discuss anywmore 😢 .)

I’m sure there won’t be a need, but will do if there is :) (It’s a sad day for the Guardian!)

PR #1223, opened 2023-08-04
Improve logging in CullInstancesWithTerminationTag task

jacobwinch:
## What does this change?

We have had several reports related to Riff-Raff failing to clean-up old instances as part of a deployment. Although these deployments still succeed, they cause the subsequent deployment to fail due to insufficient capacity (i.e. the max size is too low for Riff-Raff to double the desired capacity, as the desired capacity is higher than it should be).

Here's an example of a failed deployment: https://riffraff.gutools.co.uk/deployment/view/1707660b-32d4-410e-816e-12bb2a2b3677

And the previous deployment which actually introduces the problem: https://riffraff.gutools.co.uk/deployment/view/87d1725b-257b-4f6e-b1d8-7b86a19a7bb9

We can see that the instance that doesn't get cleaned up correctly (`i-07e2a1f0f8dfca39a`) is being tagged as expected:

```
[16:43:59] task TagCurrentInstancesWithTerminationTag Tag existing instances of the auto-scaling group <redacted> for termination
  [16:43:59] Looked up group matching tags List(TagMatch(Stage,PROD), TagMatch(Stack,ophan), TagMatch(App,big-loader-es7-arm)); identified <redacted>
  [16:43:59] Tagging i-0200e10535b928f35, i-024adc8a7f116afae, i-05937d14e1b6788f8, i-07e2a1f0f8dfca39a, i-09085cd95700fe523, i-09f4319d753e6a9eb
```

However, for some reason Riff-Raff seems to ignore `i-07e2a1f0f8dfca39a` when it gets to the `CullInstancesWithTerminationTag` task:

```
[16:46:53] task CullInstancesWithTerminationTag Terminate instances in <redacted> with the termination tag for this deploy
  [16:46:53] Looked up group matching tags List(TagMatch(Stage,PROD), TagMatch(Stack,ophan), TagMatch(App,big-loader-es7-arm)); identified <redacted>
  [16:46:54] Culling instances: i-05937d14e1b6788f8, i-09f4319d753e6a9eb, i-0200e10535b928f35, i-09085cd95700fe523, i-024adc8a7f116afae
```

I don't understand why `i-07e2a1f0f8dfca39a` is being ignored in the `CullInstancesWithTerminationTag` task, so this PR adds a lot more logging to try to help us to debug if it happens again.

## How to test

I've simulated the problem by manually changing the value of the `Magenta` tag part way through a deployment.

This deployment is now marked with a warning:

image

The [verbose logs](https://riffraff.code.dev-gutools.co.uk/deployment/view/f5b80f17-6066-487f-a18d-2177138d6ced?verbose=1) show more details than before if you view details for the `[10:01:40] task CullInstancesWithTerminationTag` section.

It should be easier to debug the aforementioned problem the next time it's reported.

There's a small amount of refactoring here so I've [tested an autoscaling deployment](https://riffraff.code.dev-gutools.co.uk/deployment/view/75431c0f-0e72-446e-abd7-1ec35c15f230) to check that it still works as expected.


magenta-lib/src/main/scala/magenta/tasks/ASGTasks.scala
```diff
@@ -244,6 +248,31 @@ case class CullInstancesWithTerminationTag(
...
+        val instancesToRetain = allInstances.diff(instancesToKill).toList
+        resources.reporter.verbose(
+          s"Decided to keep the following instances: ${instancesToRetain.map(_.instanceId).mkString(", ")}"
+        )
+
+        if (instancesToRetain.size != instancesToKill.size) {
+          resources.reporter.warning(
+            s"Terminating ${instancesToKill.size} instances and retaining ${instancesToRetain.size} instances"
+          )
+          logger.warn(
```
jacobwinch, line 261: We should get this log line in ELK which will make it easier to ascertain frequency of the problem (and whether services other than `big-loader` are affected)

jorgeazevedo (approved):

PR #1229, opened 2023-08-08
ci: Adopt `guardian/actions-riff-raff` in favour of deprecated `sbt-riffraff-artifact`

akash1810:
## What does this change?
Replaces the deprecated [sbt-riffraff-artifact](https://github.com/guardian/sbt-riffraff-artifact) with [guardian/actions-riff-raff](https://github.com/guardian/actions-riff-raff).

Some related changes are also made, including:
- [Setting up caching](https://github.com/actions/setup-java#caching-packages-dependencies)
- Using the native `sbt` that comes with [actions/setup-java](https://github.com/actions/setup-java)

## How to test
This change should be a no-op in terms of the application. Therefore to test:
- Deploy this branch to CODE
- Observe the app still working

Indeed, I have [done just this](https://riffraff.code.dev-gutools.co.uk/deployment/view/b2a39629-32d1-4288-8b21-f5abcaf351a9).

One less repository using a deprecated library.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,11 +39,16 @@ jobs:
           # The GitHub runner has 7GB of RAM, lets give 4GB to Java.
           # See https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources.
           JAVA_OPTS: -Xmx4g
-        run: |
-          set -e
-          # Ensure we don't overwrite existing (Teamcity) builds.
-          LAST_TEAMCITY_BUILD=2360
-          export GITHUB_RUN_NUMBER=$(( $GITHUB_RUN_NUMBER + $LAST_TEAMCITY_BUILD ))
+        run: sbt clean scalafmtCheckAll scalafmtSbtCheck compile lib/test riffraff/test Universal/packageZipTarball
 
-          # Note, lib/test because riffraff/riffRaffUpload only tests the riffraff project.
-          sbt clean scalafmtCheckAll scalafmtSbtCheck compile lib/test riffraff/riffRaffUpload
+      - uses: guardian/actions-riff-raff@v2
+        with:
+          # Ensure we don't overwrite existing (Teamcity) builds.
+          buildNumberOffset: 2360
+
+          projectName: tools::riffraff
+          configPath: riff-raff/riff-raff.yaml
+          contentDirectories: |
+            riff-raff:
+              - riff-raff/bootstrap.sh
+              - riff-raff/target/universal/riff-raff.tgz
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,6 +34,10 @@ jobs:
           java-version: 11
 
       - name: build + test
+        env:
+          # The GitHub runner has 7GB of RAM, lets give 4GB to Java.
+          # See https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources.
+          JAVA_OPTS: -Xmx4g
         run: |
           set -e
           # Ensure we don't overwrite existing (Teamcity) builds.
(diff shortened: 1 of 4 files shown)


.sbtopts
```diff
@@ -1,4 +1,3 @@
 -J-Xms2048m
 -J-Xmx2048m
 -J-XX:ReservedCodeCacheSize=256m
--J-XX:MaxMetaspaceSize=512M
```
akash1810, line 4: This line is dropped after seeing this in the build log:

```log
[error] Caught java.lang.OutOfMemoryError: Metaspace
[error] To best utilize classloader caching and to prevent file handle leaks, we recommend running sbt without a MaxMetaspaceSize limit.
[error]
java.lang.OutOfMemoryError: Metaspace
```

build.sbt
```diff
@@ -45,10 +45,13 @@ lazy val lib = project
     )
   )
 
+def env(propName: String): String =
+  sys.env.get(propName).filter(_.trim.nonEmpty).getOrElse("DEV")
```
akash1810, line 49: `"DEV"` will only be reached when an env var is not found. In reality this will never happen, but we want to return a `String` type here.

.github/workflows/ci.yml
```diff
@@ -32,13 +32,23 @@ jobs:
...
       - name: build + test
-        run: |
-          set -e
-          # Ensure we don't overwrite existing (Teamcity) builds.
-          LAST_TEAMCITY_BUILD=2360
-          export GITHUB_RUN_NUMBER=$(( $GITHUB_RUN_NUMBER + $LAST_TEAMCITY_BUILD ))
+        env:
+          # The GitHub runner has 7GB of RAM, lets give 4GB to Java.
+          # See https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources.
+          JAVA_OPTS: -Xmx4g
```
akash1810, line 41: Previously, we have been using the [sbt file in the repository](https://github.com/guardian/riff-raff/blob/4b06be7fc4566e81d951adf87837a30282725513/sbt) which set the JVM flags `-Xmx768M -XX:+UseCompressedOops -XX:MaxPermSize=384m`.

For simplicity, we just set `Xmx` (maximum heap size) now as `MaxPermSize` is deprecated and `UseCompressedOops` is defaulted to on already.

The value of `Xmx` is different (768M vs 4g) to take advantage of the GitHub runner resources; we have 7GB available, so might as well use (most of) it!

.github/workflows/ci.yml
```diff
@@ -32,13 +32,23 @@ jobs:
         with:
           distribution: corretto
           java-version: 11
+          cache: sbt
```
akash1810, line 35: [Caching](https://github.com/actions/setup-java#caching-packages-dependencies) should improve build times.

.github/workflows/ci.yml
```diff
@@ -32,13 +32,23 @@ jobs:
...
-        run: |
-          set -e
-          # Ensure we don't overwrite existing (Teamcity) builds.
-          LAST_TEAMCITY_BUILD=2360
-          export GITHUB_RUN_NUMBER=$(( $GITHUB_RUN_NUMBER + $LAST_TEAMCITY_BUILD ))
+        env:
+          # The GitHub runner has 7GB of RAM, lets give 4GB to Java.
+          # See https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources.
+          JAVA_OPTS: -Xmx4g
+        run: sbt clean scalafmtCheckAll scalafmtSbtCheck compile lib/test riffraff/test Universal/packageZipTarball
```
jacobwinch, line 42: Is there a particular reason that we need to run `lib/test riffraff/test` (rather than just `test`)?

jacobwinch (approved):

.github/workflows/ci.yml
akash1810: `sbt test` only runs the tests from the `riffraff` project. I'm not entirely sure why - it'll be something in the `build.sbt`, but can't quite see it.

PR #1260, opened 2023-10-05
Clarify commit message about exposed bucket names

JuliaBrigitte:
## What does this change?

Just changed wording of error message

## How to test

ci passing should suffice


AshCorr (approved):
👍

PR #1264, opened 2023-10-09
fail compile on match exhaustiveness warnings

twrichards:
Looks like warnings as errors was turned off in 2017 https://github.com/guardian/riff-raff/commit/38ec8bd7104740529c5238a2b5634c0defddb1a0

This reinstates that but just for match exhaustiveness warnings.

Tested and would've caught the issue in #1082 which needed to be fixed in #1262 (yep 180 PRs later).

FIXES required (before this will compile):

- [x] `app/utils/Retriable.scala:13:58`: match may not be exhaustive. 
`[error] It would fail on the following inputs: Cons(), Empty`
- [x] `conf/routes:114:1`: match may not be exhaustive. 
`[error] It would fail on the following input: (_, _)`

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -19,7 +19,8 @@ val commonSettings = Seq(
   scalaVersion := "2.13.12",
   scalacOptions ++= Seq(
     "-feature",
-    "-language:postfixOps,reflectiveCalls,implicitConversions"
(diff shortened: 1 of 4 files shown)


akash1810 (approved):
💎  thanks!

PR #1278, opened 2023-11-08
use a correct url to the guardian

shtukas:
At the moment url `https://www.grauniad.co.uk` is not correctly resolving. This change move to using the non `s` url to the guardian.com. Note that we should revert to the `s` url if one day we resolve the DNS problems. @mxdvl pointed this out here: https://github.com/guardian/domains-platform/issues/89

### DIFF
--- a/riff-raff/app/views/page.scala.html
+++ b/riff-raff/app/views/page.scala.html
@@ -49,7 +49,7 @@
             <p>
                 <a href="https://github.com/guardian/deploy">
                     <i>Riff Raff</i>
-                </a> is part of <a href="https://www.grauniad.co.uk">The Grauniad's</a> deployment toolset.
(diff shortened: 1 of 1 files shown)


mxdvl (changes requested):
We should fix the root cause instead: https://github.com/guardian/domains-platform/issues/89

shtukas:
@mxdvl Fair enough 😄

akash1810:
Agree with @mxdvl. Given we can be `https`, we definitely should.

shtukas:
@akash1810 Let me close the PR then, but for argument's sake, we could also redirect them to `https`, right ?

akash1810:
> we could also redirect them to `https`, right ?

I think we should favour configuring HSTS here - https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security.

PR #1291, opened 2023-12-12
Bump logback-classic to 1.4.14

AshCorr:
## What does this change?

Bumps logback-classic to 1.4.14. Resolving 4 high vulnerabilities.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -36,7 +36,7 @@ object Dependencies {
   val magentaLibDeps =
     commonDeps ++ jacksonOverrides ++ akkaSerializationJacksonOverrides ++ Seq(
       "com.squareup.okhttp3" % "okhttp" % "4.12.0",
-      "ch.qos.logback" % "logback-classic" % "1.4.8", // scala-steward:off
+      "ch.qos.logback" % "logback-classic" % "1.4.14", // scala-steward:off
       "software.amazon.awssdk" % "core" % Versions.aws,
       "software.amazon.awssdk" % "autoscaling" % Versions.aws,
       "software.amazon.awssdk" % "s3" % Versions.aws,


project/Dependencies.scala
```diff
@@ -36,7 +36,7 @@ object Dependencies {
   val magentaLibDeps =
     commonDeps ++ jacksonOverrides ++ akkaSerializationJacksonOverrides ++ Seq(
       "com.squareup.okhttp3" % "okhttp" % "4.12.0",
-      "ch.qos.logback" % "logback-classic" % "1.4.8", // scala-steward:off
+      "ch.qos.logback" % "logback-classic" % "1.4.14", // scala-steward:off
```
akash1810, line 39: We previously experienced an issue updating this dependency in this repository - https://github.com/guardian/riff-raff/pull/1232. Have we successfully deployed to CODE?

AshCorr: Haven't tried this in code. The error does seem unrelated though

`java.lang.AssertionError: assertion failed: no symbol could be loaded from interface software.amazon.awssdk.services.s3.model.ListObjectsV2Response$Builder in class ListObjectsV2Response with name Builder and classloader sbt.internal.LayeredClassLoader@6286d22d`

AshCorr: You are absolutely right this does break CODE. Unfortunately I don't think we can just dump logback and switch to devx-logs... As far as I understand Logback is pretty core to Play logging, and you would still need to use it regardless of logging to kinesis or logging to stdout 🤔

AshCorr:
This version of logback-classic is incompatible with our Play version. We probably need to do a bit of work to try and get riff-raff to Play 3.0 (or atleast a newer version of play) before we can fix this vuln.

project/Dependencies.scala
akash1810: https://github.com/guardian/riff-raff/pull/1292 switches to devx-logs, it might not resolve the reported vulnerabilities, but it's less code in the repository 😄.

PR #1301, opened 2024-01-03
Play 3

AshCorr:
## What does this change?

Bumps riff-raff to Play 3.x and various other dependencies that were blocked by Play 2.x

This resolves all high vulnerabilities for riff-raff which could not previously be done by Play 2's dependency on logback. 

## How to test

Tests run fine, runs locally. Tested in CODE.

Play 3.x is a major update and could break riff-raff. We should inform teams before we release this change.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -7,19 +7,18 @@ object Dependencies {
     val aws = "2.21.41"
     val jackson = "2.15.2"
     val awsRds = "1.12.629"
-    val enumeratumPlay = "1.7.3"
+    val enumeratumPlay = "1.8.0"
   }
 
   // https://github.com/orgs/playframework/discussions/11222
+  // We no longer have any vulnerabilities through Jackson but still need to define jackson dependencies.
+  // Jackson does not like having different versions of its packages installed.
   private val jacksonOverrides = Seq(
     "com.fasterxml.jackson.core" % "jackson-core",
     "com.fasterxml.jackson.core" % "jackson-annotations",
+    "com.fasterxml.jackson.core" % "jackson-databind",
     "com.fasterxml.jackson.datatype" % "jackson-datatype-jdk8",
     "com.fasterxml.jackson.datatype" % "jackson-datatype-jsr310",
-    "com.fasterxml.jackson.core" % "jackson-databind"
-  ).map(_ % Versions.jackson)
-
-  private val akkaSerializationJacksonOverrides = Seq(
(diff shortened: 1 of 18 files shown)


project/Dependencies.scala
```diff
@@ -7,19 +7,18 @@ object Dependencies {
...
   private val jacksonOverrides = Seq(
     "com.fasterxml.jackson.core" % "jackson-core",
     "com.fasterxml.jackson.core" % "jackson-annotations",
+    "com.fasterxml.jackson.core" % "jackson-databind",
     "com.fasterxml.jackson.datatype" % "jackson-datatype-jdk8",
     "com.fasterxml.jackson.datatype" % "jackson-datatype-jsr310",
-    "com.fasterxml.jackson.core" % "jackson-databind"
-  ).map(_ % Versions.jackson)
-
-  private val akkaSerializationJacksonOverrides = Seq(
```
tjsilver, line 22: Was this rolled into the main jackson overrides for any particular reason? (Other than pekko replacing akka here)

project/Dependencies.scala
```diff
@@ -66,38 +62,36 @@ object Dependencies {
...
       filters,
       ws,
-      // We can't update this to 4.0.0 due to an incompatibility with Play 2.8.x, attempt to update along with Play
-      "com.typesafe.akka" %% "akka-testkit" % "2.6.21" % Test,
-      "com.amazonaws" % "aws-java-sdk-rds" % Versions.awsRds
+      "org.apache.pekko" %% "pekko-testkit" % "1.0.0" % Test,
+      "com.amazonaws" % "aws-java-sdk-rds" % Versions.awsRds,
+      "org.scala-stm" %% "scala-stm" % "0.11.1",
+      // Play 3.0 currently uses logback-classic 1.4.11 which is vulnerable to CVE-2023-45960
+      "ch.qos.logback" % "logback-classic" % "1.4.14"
```
tjsilver, line 94: 👍

tjsilver (approved):
Looks good! 👍

project/Dependencies.scala
AshCorr: Pekko still uses Jackson, so I guess I could have just renamed it to `pekkoSerializationJacksonOverrides` but to be honest I didn't see too much of a reason to keep them separate anymore. Other dependencies like the AWS SDK also import Jackson and we don't explicitly have a section for that, so why have one specifically for Pekko.

PR #1321, opened 2024-03-05
feat: Add structured logs on successful parsing of `build.json`

akash1810:
## What does this change?
We want to enforce a relation between repository and project (at least). As a pre-requisite, this change adds some structured logging on each successful parsing of the `build.json` files. We can these use these logs to understand which projects, if any, might be impacted by the aforementioned restriction.

## How to test
Deploy to CODE, and [observe the logs](https://logs.gutools.co.uk/s/devx/goto/d2f3a060-db12-11ee-9631-efa6837e8b86).

### DIFF
--- a/riff-raff/app/ci/S3Build.scala
+++ b/riff-raff/app/ci/S3Build.scala
@@ -9,6 +9,11 @@ import play.api.libs.json._
 import scala.util.Try
 import scala.util.control.NonFatal
 import cats.syntax.either._
+import play.api.MarkerContext
+
+import scala.jdk.CollectionConverters._
+import net.logstash.logback.marker.Markers.appendEntries
+import utils.VCSInfo
 
 import scala.collection.Seq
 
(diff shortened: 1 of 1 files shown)


riff-raff/app/ci/S3Build.scala
```diff
@@ -38,6 +43,18 @@ class S3BuildOps(config: Config) extends Logging {
...
   } get
 
+  private def buildMarker(build: S3Build): MarkerContext = {
+    val params: Map[String, String] = Map(
+      "projectName" -> build.jobName,
+      "projectBuildNumber" -> build.number,
+      "projectBuildTool" -> build.buildTool.getOrElse("unknown"),
+      "projectVcsUrl" -> VCSInfo.normalise(build.vcsURL).getOrElse("unknown"),
+      "projectVcsBranch" -> build.branchName,
+      "projectVcsRevision" -> build.revision
```
akash1810, line 53: Prefixing these with `project` to match the [GrafanaAnnotationLogger](https://github.com/guardian/riff-raff/blob/16b4cd79c7dda073cb57fe987b7dee470f33b4fb/riff-raff/app/notification/GrafanaAnnotationLogger.scala#L35-L44).

riff-raff/app/ci/S3Build.scala
```diff
@@ -47,6 +64,18 @@ class S3BuildOps(config: Config) extends Logging {
         if (s3BuildEither.isLeft) {
           log.warn(s"Error parsing JSON from $location: $s3BuildEither")
         }
+
+        s3BuildEither match {
+          case Left(_) => {
+            log.warn(s"Error parsing JSON from $location: $s3BuildEither")
```
adamnfish, line 70: Does this duplicate L65, just above? Should we remove that if statement as well?

I also wonder whether we should log the value in the Left, rather than the Left itself? This is consistent with how it worked previously (and still does on L65), but it might make for clearer logs if we name the `_` on line 69 and log that?

akash1810: 🤦🏽 was meant to remove lines 64-66! Good catch!

akash1810: Done.

adamnfish (approved):

Issue #1342, opened 2024-05-23
Apps can not auto-scale until an `autoscaling` deploy has successfully completed


rtyley:
Riff Raff [`autoscaling`](https://riffraff.gutools.co.uk/docs/magenta-lib/types#autoscaling) deploys have always disabled ASG scaling alarms at the start of a deploy (`SuspendAlarmNotifications `), and only re-enabled them at the end of the deploy, once deployment has successfully completed:

https://github.com/guardian/riff-raff/blob/60eb09f08db8806a42e1df2e2d666fc1004a513d/magenta-lib/src/main/scala/magenta/deployment_type/AutoScaling.scala#L170-L205

There are good reasons for this, but it leads to two problems:

- Even during a [successful](https://riffraff.gutools.co.uk/deployment/view/86cfcc20-9b8f-45e4-95ef-9f03f3c86889) deploy, there's **a window of ~3 minutes when the app cannot scale**
- If a deploy fails, the app will be left with *ASG scaling alarms disabled*, requiring developers to manually re-enable them (or run another deploy) - the response time to this can be much longer, even hours.

For apps where sudden unpredictable bursts of traffic can occur, where many deploys can take place every day, this adds up to significant windows of time where the odds are eventually that a deploy will coincide with a spike in traffic that they are unable to respond to.

# Examples

jacobwinch:
Thanks for raising this issue.

> In this case, had `ResumeAlarmNotifications` been enabled immediately before `WaitForStabilization`

I think trying to run `ResumeAlarmNotifications` earlier in the process[^1] is a good idea. I agree that it would've helped to mitigate the impact of this particular incident. It would also shorten the window where apps cannot scale during successful deployments[^2], which might help with performance more generally.

In order for this to work I think it might be prudent to replace the final `WaitForStabilization` task with a new task (`WaitForOldInstancesToTerminate`, or similar). 

The current task is checking for an [expected _number_ of instances](https://github.com/guardian/riff-raff/blob/c7433640ab0f6239f4b84ba1b5ff56863cb498b1/magenta-lib/src/main/scala/magenta/tasks/AWS.scala#L346-L363). This is currently deterministic because Riff-Raff is essentially holding a lock on the desired capacity setting by blocking scaling operations. 

[...]

PR #1345, opened 2024-05-23
`autoscaling` deploy: re-enable ASG scaling before final stabilisation check

rtyley:
This aims to address, to an extent, issue https://github.com/guardian/riff-raff/issues/1342 - that in an  [`autoscaling`](https://riffraff.gutools.co.uk/docs/magenta-lib/types#autoscaling) deploy, **apps _can not_ auto-scale** until the original number of instances in the ASG is capable of successfully handling the level of traffic, as checked by `WaitForStabilization`. On 22nd May 2024, this inability to auto-scale led to a severe outage in the Ophan Tracker, when a large increase in pageview traffic during a deploy meant the original number of instances could never have handled the load.

image

## Disabling auto-scaling

Ever since https://github.com/guardian/riff-raff/pull/83 in April 2013, Riff Raff has disabled ASG scaling alarms at the start of a deploy (`SuspendAlarmNotifications`), and only re-enabled them at the end of the deploy (`ResumeAlarmNotifications`) once deployment has successfully completed.

In December 2016, with https://github.com/guardian/riff-raff/pull/403, an additional `WaitForStabilization` was added as a penultimate deploy step, which served two purposes:

1. Ensure that **the cull of old instances has completed** _before_ the deploy ends
2. Prove that the new boxes can handle their load _without_ the old boxes.

However, the `WaitForStabilization` step was placed _before_ `ResumeAlarmNotifications` - so the original number of EC2 instances in the ASG _must_ be capable of supporting the latest level of traffic, otherwise `WaitForStabilization` and the entire deploy will fail, leaving auto-scaling _indefinitely_ disabled - potentially waiting hours for human intervention to fix the problem.

## The old-instance cull must _complete_

By putting `ResumeAlarmNotifications` _before_ `WaitForStabilization`, the Ophan outage would have been shortened from 1 hour to ~2 minutes, but as @jacobwinch [points out](https://github.com/guardian/riff-raff/issues/1342#issuecomment-2127241818), making _only_ that change could lead to a race condition:

1. Once `ResumeAlarmNotifications` executes, a 'low CPU' alarm could immediately decide to reduce the desired size of the ASG, randomly removing _new_ instances
3. `WaitForStabilization` could then see that the ASG is at its desired size, even if it happens that the termination of the old servers has not yet completed
4. The deploy then reports as 'complete', even though old EC2 instances are potentially still active

Consequently, with this change we are also introducing a new task, `WaitForCullToComplete`, that checks _specifically_ that EC2 instances _tagged_ for termination have been removed from the ASG - more specific than the old assumption with `WaitForStabilization` that the ASG being at desired capacity definitely means all the old instances are gone.

The final sequence of tasks in an `autoscaling` deploy now looks like this:

https://github.com/guardian/riff-raff/blob/0fb840e64ac924ff4b477ccf6d93a80cf0bc56cd/magenta-lib/src/main/scala/magenta/deployment_type/AutoScaling.scala#L199-L215

Common logic for _finding_ EC2 instances _tagged_ for termination, used by both `CullInstancesWithTerminationTag` & `WaitForCullToComplete`, has been factored out into the new `CullSummary` class.

# Testing

I've [deployed](https://riffraff.code.dev-gutools.co.uk/deployment/view/2d96ddb0-ed33-4198-b23e-7ee8bf41b83f) this branch to [Riff Raff CODE](https://riffraff.code.dev-gutools.co.uk/) as build 3385, and then used that to [deploy](https://riffraff.code.dev-gutools.co.uk/deployment/view/bb2d6042-44fe-4d8f-b166-b15a421c1743?verbose=1) the Ophan Dashboard to CODE, successfully! You can see that the new `WaitForCullToComplete` step took place after `ResumeAlarmNotifications`, and the `WaitForStabilization` step completed immediately after that:

image

cc @guardian/ophan

### DIFF
--- a/magenta-lib/src/main/scala/magenta/tasks/tasks.scala
+++ b/magenta-lib/src/main/scala/magenta/tasks/tasks.scala
@@ -29,6 +29,7 @@ import java.time.Duration.{between, ofMillis, ofSeconds}
 import java.time.{Duration, Instant}
 import scala.collection.parallel.CollectionConverters._
 import scala.concurrent.{Await, Future}
+import scala.util.{Failure, Try}
 import scala.util.control.NonFatal
 
 case class S3Upload(
@@ -274,22 +275,30 @@ trait PollingCheck {
   ): Unit = {
     val expiry = Instant.now().plus(duration)
 
-    def checkAttempt(currentAttempt: Int): Unit = if (!theCheck) {
-      if (stopFlag)
-        reporter.info("Abandoning remaining checks as stop flag has been set")
-      else {
-        val remainingTime = between(Instant.now(), expiry)
-        if (remainingTime.isNegative)
-          reporter.fail(
-            s"Check failed to pass within $duration milliseconds (tried $currentAttempt times) - aborting"
-          )
+    def checkAttempt(currentAttempt: Int): Unit = {
+      val checkResult =
+        Try(theCheck).recover { case NonFatal(e) =>
+          reporter.info(e.getMessage)
+          false
+        }.get // '.get' allows a fatal exception, like an OutOfMemoryError, to kill the process, which is what we want!
+
+      if (!checkResult) {
+        if (stopFlag)
+          reporter.info("Abandoning remaining checks as stop flag has been set")
         else {
-          val sleepyTime = calculateSleepTime(currentAttempt)
-          reporter.verbose(
-            f"Check failed on attempt #$currentAttempt (Will wait for a further ${remainingTime.toSeconds} seconds, retrying again after ${sleepyTime.toSeconds}s)"
-          )
-          Thread.sleep(sleepyTime.toMillis)
-          checkAttempt(currentAttempt + 1)
+          val remainingTime = between(Instant.now(), expiry)
+          if (remainingTime.isNegative)
+            reporter.fail(
+              s"Check failed to pass within $duration milliseconds (tried $currentAttempt times) - aborting"
+            )
+          else {
+            val sleepyTime = calculateSleepTime(currentAttempt)
+            reporter.verbose(
+              f"Check failed on attempt #$currentAttempt (Will wait for a further ${remainingTime.toSeconds} seconds, retrying again after ${sleepyTime.toSeconds}s)"
+            )
+            Thread.sleep(sleepyTime.toMillis)
+            checkAttempt(currentAttempt + 1)
+          }
         }
       }
     }
--- a/magenta-lib/src/main/scala/magenta/deployment_type/AutoScaling.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/AutoScaling.scala
@@ -202,12 +202,17 @@ object AutoScaling extends DeploymentType with BucketParameters {
           target.region,
           terminationGrace(pkg, target, reporter)
         ),
-        WaitForStabilization(
+        ResumeAlarmNotifications(autoScalingGroup, target.region),
+        WaitForCullToComplete(
           autoScalingGroup,
           secondsToWait(pkg, target, reporter),
           target.region
         ),
-        ResumeAlarmNotifications(autoScalingGroup, target.region)
+        WaitForStabilization(
+          autoScalingGroup,
+          secondsToWait(pkg, target, reporter),
+          target.region
+        )
       )
     }
     val groupsToUpdate: List[AutoScalingGroupInfo] =
--- a/magenta-lib/src/main/scala/magenta/tasks/ASGTasks.scala
+++ b/magenta-lib/src/main/scala/magenta/tasks/ASGTasks.scala
@@ -1,14 +1,18 @@
 package magenta.tasks
 
 import magenta.deployment_type.AutoScalingGroupInfo
+import magenta.tasks.EC2.withEc2Client
+import magenta.tasks.autoscaling.CullSummary
 import magenta.{KeyRing, _}
 import software.amazon.awssdk.awscore.exception.AwsServiceException
 import software.amazon.awssdk.services.autoscaling.AutoScalingClient
 import software.amazon.awssdk.services.autoscaling.model.{
   AutoScalingGroup,
+  Instance,
   LifecycleState,
   SetInstanceProtectionRequest
 }
+import software.amazon.awssdk.services.ec2.Ec2Client
 
 import java.time.Duration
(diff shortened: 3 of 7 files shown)


rtyley:
### `WaitForStabilization` for culled instances?

@jacobwinch [comments](https://github.com/guardian/riff-raff/issues/1342#issuecomment-2127241818):

> In order for this to work I think it would be desirable to replace the final `WaitForStabilization` task with a new task (`WaitForOldInstancesToTerminate`, or similar). 
> 
> The current task is checking for an [expected _number_ of instances](https://github.com/guardian/riff-raff/blob/c7433640ab0f6239f4b84ba1b5ff56863cb498b1/magenta-lib/src/main/scala/magenta/tasks/AWS.scala#L346-L363). This is currently an acceptable way to check because Riff-Raff is essentially holding a lock on the desired capacity setting by blocking scaling operations (i.e. we know that the number won't change).
> 
> Once we re-enable scaling the desired number of instances becomes a moving target, so I think it would be better for Riff-Raff to check that there are 0 instances with the [`Magenta=Terminate` tag](https://github.com/guardian/riff-raff/blob/c7433640ab0f6239f4b84ba1b5ff56863cb498b1/magenta-lib/src/main/scala/magenta/tasks/ASGTasks.scala#L246-L247) still running in the ASG. This would allow us to be sure that all instances are now running the build currently being deployed, which means it is safe/correct to mark the deployment as successful.

@jacobwinch, rather than introduce `WaitForOldInstancesToTerminate`, would it reasonable to update the implementation of `CullInstancesWithTerminationTag` so that it just blocks until all the instances it's been asked to cull actually have terminated?

jacobwinch:
> would it reasonable to update the implementation of `CullInstancesWithTerminationTag` so that it just blocks until all the instances it's been asked to cull actually have terminated?

I think it's probably safe to allow scaling actions to run again as soon as we've _asked_ AWS to terminate the instances running the old build. At this point there is no risk of AWS trying to scale down the 'wrong' instances (i.e. the ones running the new build) and we know it is safe to launch more instances that will run with the newest build (as instances running this build have already passed the health check).

With that in mind, the only downside of the suggested approach is that while we're blocking and waiting for the instances to terminate, we could probably be allowing the app to start scaling again if we were to split the 'waiting' work into a separate task. We're probably talking about a pretty small time window here though so if it's considerably easier to implement the version you've suggested then I think that would still be a big improvement!

rtyley:
You've convinced me @jacobwinch, so I've added some WIP to this PR to implement a `WaitForCullToComplete` task! Not quite tidied up yet, but you probably get the idea. Some code from `CullInstancesWithTerminationTag` is factored out into a `CullSummary` class so that `WaitForCullToComplete` can also use it.

magenta-lib/src/main/scala/magenta/tasks/tasks.scala
```diff
@@ -274,22 +275,30 @@ trait PollingCheck {
...
-        if (remainingTime.isNegative)
-          reporter.fail(
-            s"Check failed to pass within $duration milliseconds (tried $currentAttempt times) - aborting"
-          )
+    def checkAttempt(currentAttempt: Int): Unit = {
+      val checkResult =
+        Try(theCheck).recover { case NonFatal(e) =>
+          reporter.info(e.getMessage)
+          false
+        }.get // '.get' allows a fatal exception, like an OutOfMemoryError, to kill the process, which is what we want!
```
rtyley, line 283: @jacobwinch & I decided that as polling checks inevitably involve network calls, it makes sense to put an exception-catching guard around the invocation of `theCheck` - if a non-fatal exception is received, we probably _don't_ want to kill the deploy, just attempt the check again after the next `sleepyTime`.

jacobwinch (approved):
This looks good to me 👍 

I ran one additional test to confirm that automatically scaling up while Riff-Raff is running `WaitForCullToComplete` works successfully.

## Test Setup

1. [Redeployed this branch to Riff-Raff `CODE`](https://riffraff.code.dev-gutools.co.uk/deployment/view/76b56d60-b554-4465-a489-a0f4926aef4f)
2. Started sending some artificial traffic to a [test service](https://github.com/guardian/cdk-playground), with the aim of forcing the scale up action to coincide with the Riff-Raff deployment
3. Started a [deployment of the test service via Riff-Raff `CODE`](https://riffraff.code.dev-gutools.co.uk/deployment/view/f7bc23cb-070e-4e44-9395-d8d313a9ae84?verbose=1)

## Observations

**12:16:56 - Deployment starts**: `[12:16:56] deploy for devx::cdk-playground (build 801) to stage PROD` 

**12:17:05 - Scaling actions are disabled by Riff-Raff**: `[12:17:05] task SuspendAlarmNotifications Suspending Alarm Notifications - playground-PROD-cdk-playground-AutoScalingGroupCdkplaygroundASGD6E49F0F-9KOUOLFGBKZB will no longer scale on any configured alarms`

**12:18:41[^1] - Alarm used for automatically scaling up enters alarm state**: 

image

_Automatic scale up does not happen yet because it is still blocked by Riff-Raff._

**12:19:39 - Scaling actions are re-enabled by Riff-Raff:** `[12:19:39] task ResumeAlarmNotifications Resuming Alarm Notifications - playground-PROD-cdk-playground-AutoScalingGroupCdkplaygroundASGD6E49F0F-9KOUOLFGBKZB will scale on any configured alarms`

**12:19:40 -  New `WaitForCullToComplete` task starts running:** `[12:19:40] task WaitForCullToComplete Check that all instances tagged for termination in playground-PROD-cdk-playground-AutoScalingGroupCdkplaygroundASGD6E49F0F-9KOUOLFGBKZB have been terminated`

**12:19:45 - The ASG scales up automatically:**

image

**12:21:41 - New `WaitForCullToComplete` task (and overall deployment) completes:** `[12:21:41] task WaitForStabilization Check the desired number of hosts in both the ASG (playground-PROD-cdk-playground-AutoScalingGroupCdkplaygroundASGD6E49F0F-9KOUOLFGBKZB) and ELB are up and that the number of hosts match`

[^1]: The screenshot here shows a UTC timestamp, I've converted to BST to make it easier to follow the timeline

rtyley:
> I ran one additional test to confirm that automatically scaling up while Riff-Raff is running `WaitForCullToComplete` works successfully.

Magnificent! Great test, and gratifying to see it work correctly! I shall now merge 👍

PR #1348, opened 2024-05-30
Use strongly-typed `Duration`s for time

rtyley:
This PR improves time & duration handling in Riff Raff by switching representation from `Int` or `Long` types (types which don't incorporate a time-unit, meaning we have to work out from our _context_ whether we are using seconds, milliseconds, epoch-ms, etc) to using the great `java.time.*` classes `Instant` & `Duration`.

This fixes a lot of things like:

* slightly obscure integer constants like `5 * 60 * 1000` (denoting 5 minutes)
* the inability of the compiler to check we're performing the right kind of operation to change time-units - the compiler can't know if what we're doing with an `Int` makes sense!
* variable names like `durationMillis`, where 'millis' had to be included because the declared type of `Long` isn't _telling_ us what the type is
* frequently multiplying numbers by 1000 to transform a value that we are quite sure is measured in seconds, into milliseconds!
* the risk that integer arithmetic might overflow when making time calculations, silently corrupting the result
 
If you can use a strongly-typed time class instead - like `Duration` or `Instant` - the compiler and library author can do all of that stuff for us, it's much better! You also stand a better chance of avoiding the hallowed [_falsehoods programmers believe about time_](https://gist.github.com/timvisee/fcda9bbdff88d45cc9061606b4b923ca).

### From `Param[Int]` to `Param[Duration]` - durations are still specified in seconds ⏱️ 

  Riff Raff has several parameters which specify durations (eg `healthcheckGrace`, `warmupGrace`, etc). Historically these have all expected integer values and been documented to be duration in **seconds**:

```yaml
    parameters:
      healthcheckGrace: 420
      warmupGrace: 300
```

We don't want to make everyone change their Riff Raff configuration, and so those parameters need to continue to accept the same numeric values and still interpret them as seconds.

The `Param.waitingSecondsFor()` helper method lets us create parameters that are a `Param[Duration]` (which could have been problematic because `java.time.Duration` has no _specified_ time unit) that still _explicitly_ parse their input as seconds, and are tested to do so.

### Migrating from Joda-Time to java.time
We should be using `java.time.*` classes in our codebases, in preference to the wonderful but old Joda-Time library:

https://blog.joda.org/2014/11/converting-from-joda-time-to-javatime.html

### Testing

I've [deployed](https://riffraff.code.dev-gutools.co.uk/deployment/view/98e02d80-825d-4075-865e-bf85dc34df3f) this branch to CODE Riff Raff (https://riffraff.code.dev-gutools.co.uk/) and then used CODE Riff Raff to deploy the Ophan Dashboard to PROD, and DotCom Frontend to CODE, successfully.

Frontend [uses](https://github.com/guardian/frontend/blob/f2c54a6df0344e80468208dd94deb8adf42f45ee/riff-raff.yaml#L14) a custom warmupGrace value of 60 seconds, different to the default of 1 second we normally use:

https://github.com/guardian/riff-raff/blob/feaf4f0661fe50d6b3c9ea0f246c246691622403/magenta-lib/src/main/scala/magenta/deployment_type/AutoScaling.scala#L100-L105

We can see in this [successful deploy](https://riffraff.code.dev-gutools.co.uk/deployment/view/fbb157ef-e58b-4fea-8205-eec5a9d45bfa) of Frontend CODE by this branch of Riff Raff, that the warmupGrace value of 60 has still been correctly parsed, and displayed as 60000 milliseconds:

image

See also:

* https://github.com/guardian/ophan/pull/3988 - another PR achieving a similar thing for Ophan

### DIFF
--- a/magenta-lib/src/main/scala/magenta/deployment_type/AutoScaling.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/AutoScaling.scala
@@ -16,6 +16,9 @@ import software.amazon.awssdk.services.autoscaling.model.AutoScalingGroup
 import magenta.tasks.{S3 => S3Tasks}
 import software.amazon.awssdk.services.ssm.SsmClient
 
+import java.time.Duration
+import java.time.Duration.{ofMinutes, ofSeconds}
+
 sealed trait MigrationTagRequirements
 case object NoMigration extends MigrationTagRequirements
 case object MustBePresent extends MigrationTagRequirements
@@ -88,22 +91,24 @@ object AutoScaling extends DeploymentType with BucketParameters {
       | - scale up, wait for the new instances to become healthy and then scale back down
     """.stripMargin
 
-  val secondsToWait = Param(
-    "secondsToWait",
-    "Number of seconds to wait for instances to enter service"
-  ).default(15 * 60)
-  val healthcheckGrace = Param(
-    "healthcheckGrace",
-    "Number of seconds to wait for the AWS api to stabilise"
-  ).default(20)
-  val warmupGrace = Param(
-    "warmupGrace",
-    "Number of seconds to wait for the instances in the load balancer to warm up"
-  ).default(1)
-  val terminationGrace = Param(
-    "terminationGrace",
-    "Number of seconds to wait for the AWS api to stabilise after instance termination"
-  ).default(10)
+  val secondsToWait: Param[Duration] = Param
+    .waitingSecondsFor("secondsToWait", "instances to enter service")
+    .default(ofMinutes(15))
+  val healthcheckGrace: Param[Duration] = Param
+    .waitingSecondsFor("healthcheckGrace", "the AWS api to stabilise")
+    .default(ofSeconds(20))
+  val warmupGrace: Param[Duration] = Param
+    .waitingSecondsFor(
+      "warmupGrace",
+      "the instances in the load balancer to warm up"
+    )
+    .default(ofSeconds(1))
+  val terminationGrace: Param[Duration] = Param
+    .waitingSecondsFor(
+      "terminationGrace",
+      "the AWS api to stabilise after instance termination"
+    )
+    .default(ofSeconds(10))
 
   val prefixStage = Param[Boolean](
     "prefixStage",
@@ -165,7 +170,7 @@ object AutoScaling extends DeploymentType with BucketParameters {
         autoScalingGroup: AutoScalingGroupInfo
     ): List[ASGTask] = {
       List(
-        WaitForStabilization(autoScalingGroup, 5 * 60 * 1000, target.region),
+        WaitForStabilization(autoScalingGroup, ofMinutes(5), target.region),
         CheckGroupSize(autoScalingGroup, target.region),
         SuspendAlarmNotifications(autoScalingGroup, target.region),
         TagCurrentInstancesWithTerminationTag(autoScalingGroup, target.region),
@@ -174,32 +179,32 @@ object AutoScaling extends DeploymentType with BucketParameters {
         HealthcheckGrace(
           autoScalingGroup,
           target.region,
-          healthcheckGrace(pkg, target, reporter) * 1000
+          healthcheckGrace(pkg, target, reporter)
         ),
         WaitForStabilization(
           autoScalingGroup,
-          secondsToWait(pkg, target, reporter) * 1000,
+          secondsToWait(pkg, target, reporter),
           target.region
         ),
         WarmupGrace(
           autoScalingGroup,
           target.region,
-          warmupGrace(pkg, target, reporter) * 1000
+          warmupGrace(pkg, target, reporter)
         ),
         WaitForStabilization(
           autoScalingGroup,
-          secondsToWait(pkg, target, reporter) * 1000,
+          secondsToWait(pkg, target, reporter),
           target.region
         ),
         CullInstancesWithTerminationTag(autoScalingGroup, target.region),
         TerminationGrace(
           autoScalingGroup,
           target.region,
-          terminationGrace(pkg, target, reporter) * 1000
+          terminationGrace(pkg, target, reporter)
         ),
         WaitForStabilization(
           autoScalingGroup,
-          secondsToWait(pkg, target, reporter) * 1000,
+          secondsToWait(pkg, target, reporter),
           target.region
         ),
         ResumeAlarmNotifications(autoScalingGroup, target.region)
--- a/magenta-lib/src/main/scala/magenta/deployment_type/GCS.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/GCS.scala
@@ -9,13 +9,22 @@ object GCS extends DeploymentType {
   val documentation = "For uploading files into a GCS bucket."
 
   val prefixStage =
-    Param("prefixStage", "Prefix the GCS bucket key with the target stage")
+    Param[Boolean](
+      "prefixStage",
+      "Prefix the GCS bucket key with the target stage"
+    )
       .default(true)
   val prefixPackage =
-    Param("prefixPackage", "Prefix the GCS bucket key with the package name")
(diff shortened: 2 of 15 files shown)


magenta-lib/src/main/scala/magenta/deployment_type/GCS.scala
```diff
@@ -9,13 +9,22 @@ object GCS extends DeploymentType {
   val documentation = "For uploading files into a GCS bucket."
 
   val prefixStage =
-    Param("prefixStage", "Prefix the GCS bucket key with the target stage")
+    Param[Boolean](
```
rtyley, line 12: The change in signature for the `Param` constructor does mean we need to add an explicit type here, otherwise we get this compilation error:

```
[error] /Users/Roberto_Tyley/code/riff-raff/magenta-lib/src/main/scala/magenta/deployment_type/GCS.scala:18:10: diverging implicit expansion for type play.api.libs.json.Reads[T1]
[error] starting with method Tuple22R in trait GeneratedReads
[error]     Param(
[error]          ^
```

magenta-lib/src/main/scala/magenta/deployment_type/Param.scala
```diff
@@ -38,21 +40,24 @@ case class Param[T](
...
+      .flatMap(jsValue => parse(jsValue))
+
+  def parse(jsValue: JsValue): Option[T] = Json.fromJson[T](jsValue).asOpt
+
   def apply(
       pkg: DeploymentPackage,
       target: DeployTarget,
       reporter: DeployReporter
-  )(implicit reads: Reads[T], manifest: Manifest[T]): T = {
+  ): T = {
```
rtyley, line 60: This method does need a `Reads[T]` instance, to know how to parse a JSON value into an instance of a `T`, but asking for it to be supplied here, where the method is called, means that every single caller of this code must have the _right_ `Reads[T]` instance available, and they must all agree, otherwise a parameter may get parsed differently depending on which code calls it! This makes testing that a parameter is being correctly parsed harder than it needs to be.

Better to add the requirement for the `Reads[T]` to the _constructor_ of the `Param[T]` class, so that once the `Param[T]` is constructed, it always behave in the same way. Consequently, we can write a test using a param - like the `healthcheckGrace` param - and be sure that it will behave the same way when the implementation code calls it.

magenta-lib/src/test/scala/magenta/input/resolver/DeploymentTypeResolverTest.scala
```diff
@@ -95,7 +95,7 @@ class DeploymentTypeResolverTest
     val deploymentTypesWithParams = List(
       stubDeploymentType(
         Seq("upload", "deploy"),
-        register => List(Param[String]("param1")(register))
+        implicit register => List(Param[String]("param1"))
```
rtyley, line 98: `register` is getting shifted around here because the constructor of `Param` now takes 3 implicit parameters, and it's easier to make `register` implicit rather than pass in 3 parameters.

magenta-lib/src/main/scala/magenta/deployment_type/AutoScaling.scala
```diff
@@ -174,32 +179,32 @@ object AutoScaling extends DeploymentType with BucketParameters {
         HealthcheckGrace(
           autoScalingGroup,
           target.region,
-          healthcheckGrace(pkg, target, reporter) * 1000
```
rtyley, line 177: We no longer have to randomly multiply things by 1000! 🎉

magenta-lib/src/main/scala/magenta/tasks/UpdateCloudFormationTask.scala
```diff
@@ -31,21 +30,23 @@ import software.amazon.awssdk.services.s3.S3Client
...
 
 /** A simple trait to aid with attempting an update multiple times in the case
   * that an update is already running.
   */
 trait RetryCloudFormationUpdate {
-  def duration: Long = 15 * 60 * 1000 // wait fifteen minutes
-  def calculateSleepTime(currentAttempt: Int): Long =
-    30 * 1000 // sleep 30 seconds
+  def duration: Duration = ofMinutes(15)
+  def calculateSleepTime(currentAttempt: Int): Duration = ofSeconds(30)
```
rtyley, line 44: Note that we no longer need inline comments telling us what the numbers mean 😃

magenta-lib/src/test/scala/magenta/deployment_type/ParamTest.scala
```diff
@@ -1,20 +1,31 @@
...
-  val paramsList = mutable.Map.empty[String, Param[_]]
-  def add(param: Param[_]) = paramsList += param.name -> param
+class ParamRegistrationTest extends AnyFlatSpec with Matchers {
+  class TestRegister extends ParamRegister {
+    val paramsList = mutable.Map.empty[String, Param[_]]
+    def add(param: Param[_]) = paramsList += param.name -> param
+  }
+
+  "Param registration" should "occur when a param is created" in {
+    implicit val register: TestRegister = new TestRegister
```
rtyley, line 22: Separating this `ParamRegistrationTest` class out lets us give a specific `TestRegister` to the one test that cares about it, and give a single shared stub `ParamRegister` to all the other tests that don't care about it all - this removes this repetitive line of noise from each test case:

https://github.com/guardian/riff-raff/blob/192e63ff5ba27382d3b2620185c7c817e6fae471/magenta-lib/src/test/scala/magenta/deployment_type/ParamTest.scala#L45

magenta-lib/src/test/scala/magenta/deployment_type/ParamTest.scala
```diff
@@ -31,165 +42,100 @@ class ParamTest extends AnyFlatSpec with Matchers with MockitoSugar {
...
-    val param = Param[String]("test")
-
-    register.paramsList.size shouldBe 1
-    register.paramsList shouldBe Map("test" -> param)
-  }
-
-  it should "extract a value from a package using get" in {
-    implicit val register = new TestRegister
-    val pkg = DeploymentPackage(
+  def stubDeploymentPackage(pkgSpecificData: Map[String, JsValue]) =
```
rtyley, line 45: Introducing this `stubDeploymentPackage()` helper method has made each test case _much_ more concise - to the point where you can see what each test case is actually testing! 🎉

magenta-lib/src/main/scala/magenta/deployment_type/Param.scala
```diff
@@ -38,21 +40,24 @@ case class Param[T](
       (DeploymentPackage, DeployTarget) => Either[String, T]
     ] = None,
     deprecatedDefault: Boolean = false
-)(implicit register: ParamRegister) {
+)(implicit register: ParamRegister, reads: Reads[T], manifest: Manifest[T]) {
```
rtyley, line 43: As explained below, it's better to add the requirement for the `Reads[T]` to the _constructor_ of the `Param[T]` class, so that once the `Param[T]` is constructed, it always parses things in the same way, rather than depending on how it's called.

jacobwinch (approved):
Thanks for improving this 👍

rtyley:
An unintended consequence of this PR is that unfortunately Riff Raff's [autogenerated documentation](https://riffraff.gutools.co.uk/docs/magenta-lib/types#autoscaling) is displaying default values which now look like [ISO 8601 Durations](https://en.wikipedia.org/wiki/ISO_8601#Durations) (`PT20S`, `PT15M`, etc)

image

These are nicer to look at, but the real `riff-raff.yaml` fields actually currently _do not_ accept ISO 8601 Duration strings, and only accept the number-of-seconds as they always did - so I've managed to make the documentation more confusing.

Looking at the `docs.DeployTypeDocs.defaultFromParam()` method, it looks difficult to cleanly fix this to render a default `Duration` value as seconds. It might be easier to make the field accept ISO 8601 Duration strings, as well as the numeric seconds fields that currently exist in the wild and which we will need to continue supporting.

PR #1351, opened 2024-06-05
Upgrade to Scanamo v1.1.1

rtyley:
Here we're moving up from Scanamo [1.0.0-M11](https://github.com/scanamo/scanamo/releases/tag/v1.0.0-M11) to [1.1.1](https://github.com/scanamo/scanamo/releases/tag/v1.1.1), which involves several changes to requirements:

* **AWS SDK v1 → v2** - Scanamo [1.0-M13](https://github.com/scanamo/scanamo/releases/tag/v1.0-M13) and above require AWS SDK v2. AWS SDK v2 is actually used quite widely already in Riff Raff, so this wasn't hard - there does still seem to be _some_ other code in Riff Raff using AWS SDK v1 at the moment.
* **Scanamo API changes**
  - Auto-generated `DynamoFormat` handlers are now created with `org.scanamo.generic.auto._` rather than `org.scanamo.auto._` (https://github.com/scanamo/scanamo/pull/538 - [1.0-M12](https://github.com/scanamo/scanamo/releases/tag/v1.0.0-M12))
  - `DynamoFormat.coercedXmap()` now has a single set of two arguments, rather than curried arguments (https://github.com/scanamo/scanamo/pull/588 - [1.0-M13](https://github.com/scanamo/scanamo/releases/tag/v1.0-M13))
  - `Table.put()` now returns `Unit` rather than the old value- you can use `Table.putAndReturn()` if you want a value returned, which allows you to specify whether it's the prior value or the new one you want (https://github.com/scanamo/scanamo/pull/486 - [1.0-M13](https://github.com/scanamo/scanamo/releases/tag/v1.0-M13))

See also:

* https://github.com/guardian/maintaining-scala-projects/issues/6

### DIFF
--- a/riff-raff/app/persistence/ContinuousDeploymentConfigRepository.scala
+++ b/riff-raff/app/persistence/ContinuousDeploymentConfigRepository.scala
@@ -3,9 +3,8 @@ package persistence
 import java.util.UUID
 import ci.{ContinuousDeploymentConfig, Trigger}
 import org.scanamo.syntax._
-import org.scanamo.auto._
+import org.scanamo.generic.auto._
 import org.scanamo.{DynamoFormat, Table}
-import cats.syntax.either._
 import conf.Config
 
 class ContinuousDeploymentConfigRepository(config: Config)
@@ -13,8 +12,9 @@ class ContinuousDeploymentConfigRepository(config: Config)
 
   implicit val triggerModeFormat =
     DynamoFormat.coercedXmap[Trigger.Mode, String, NoSuchElementException](
-      Trigger.withName
-    )(_.toString)
+      Trigger.withName,
+      _.toString
+    )
 
   override val tablePrefix = "continuous-deployment-config"
 
--- a/riff-raff/app/persistence/TargetDynamoRepository.scala
+++ b/riff-raff/app/persistence/TargetDynamoRepository.scala
@@ -2,11 +2,9 @@ package persistence
 
 import ci.Target
 import org.scanamo.Table
-import org.scanamo.error.DynamoReadError
 import conf.Config
 import org.joda.time.DateTime
-import org.scanamo.auto._
-import org.scanamo.syntax._
+import org.scanamo.generic.auto._
 
 case class TargetId(
     targetKey: String,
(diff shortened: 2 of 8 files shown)


riff-raff/app/persistence/ContinuousDeploymentConfigRepository.scala
```diff
@@ -3,18 +3,18 @@ package persistence
...
 class ContinuousDeploymentConfigRepository(config: Config)
     extends DynamoRepository(config) {
 
   implicit val triggerModeFormat =
     DynamoFormat.coercedXmap[Trigger.Mode, String, NoSuchElementException](
-      Trigger.withName
-    )(_.toString)
+      Trigger.withName,
+      _.toString
+    )
```
rtyley, line 17: `DynamoFormat.coercedXmap()` now has a single set of two arguments, rather than curried arguments (due to https://github.com/scanamo/scanamo/pull/588 - released with [1.0-M13](https://github.com/scanamo/scanamo/releases/tag/v1.0-M13))

riff-raff/app/persistence/TargetDynamoRepository.scala
```diff
@@ -42,7 +40,7 @@ class TargetDynamoRepository(config: Config) extends DynamoRepository(config) {
       target: Target,
       projectName: String,
       lastSeen: DateTime
-  ): Option[Either[DynamoReadError, TargetId]] =
+  ): Unit =
     exec(table.put(TargetId(target, projectName, lastSeen)))
```
rtyley, line 44: `Table.put()` now returns `Unit` rather than the old value- you can use `Table.putAndReturn()` if you want a value returned, which allows you to specify whether it's the prior value or the new one you want (due to https://github.com/scanamo/scanamo/pull/486 - release with [1.0-M13](https://github.com/scanamo/scanamo/releases/tag/v1.0-M13))

jacobwinch: I presume this will throw an exception if the update fails, but this seems safe given that our only usage of this function is here:

https://github.com/guardian/riff-raff/blob/295ec4b91f9ee1769bc5daf3f46da59f1471bf4f/riff-raff/app/ci/TargetResolver.scala#L57-L59

rtyley: Yeah, I think it would always have thrown an exception if the update fails, the difference is just that if the _update_ succeeds, there's now no value being returned, and no chance of a `DynamoReadError` representing a deserialisation error. And you're right, `TargetResolver` is catching any erxception!

rtyley:
This was [deployed](https://riffraff.gutools.co.uk/deployment/view/ecbb60bc-1ab9-476f-b30f-b5d35ee77ae8) to PROD as build 3392, and subsequent deploys look good:

image

PR #1374, opened 2024-09-06
Add riff raff project tag

jorgeazevedo:
## What does this change?

Riff raff tags every resource it deploys with a set of common tags, and this patch adds a new one called `gu:riff-raff-project` with the name of the riff raff project.

This is useful to quickly find the riff raff project, and hence the deployment history, of any resource that's been deployed by riff raff.

## How to test

This can easily tested by deploying to riff raff CODE, and using it to deploy an application. We've deployed cdk playground and amiable and saw that the new tag is being set correctly.

Screenshot 2024-09-20 at 11 54 42

Merging this PR means that every project that gets deployed with riff raff after this change will have to perform a tag update across all the project resources. It's a one time cost that shouldn't cause any issues.


riff-raff/app/AppComponents.scala
```diff
@@ -108,7 +108,8 @@ class AppComponents(
 
       Map(
         "gu:build-tool" -> buildTool.getOrElse(default),
-        "gu:repo" -> repoUrl.getOrElse(default)
+        "gu:repo" -> repoUrl.getOrElse(default),
+        "gu:riff-raff-project" -> projectName
```
akash1810, line 112: minor: Should we namespace this as `gu:riff-raff:project`? This'll be similar behaviour to the [GuCDK tags](https://github.com/guardian/cdk/blob/main/src/constants/metadata-keys.ts). Saying that, the `gu:build-tool` and `gu:build-identifier` are not namespaced 🤷🏽‍♂️.

akash1810 (approved):

akash1810:
This will have a huge positive impact to so many scenarios, e.g. auditing, incident response, etc. 💎

PR #1378, opened 2024-09-25
Revert "Add riff raff project tag"

jorgeazevedo:
Reverts guardian/riff-raff#1374

It's emerged that this change might break deployments for projects that use the ami update functionality. If the stack has enough components that all need to be tagged, the list of events in the cloudformation stack becomes so long that it breaks. 

See
https://github.com/guardian/riff-raff/blob/main/magenta-lib/src/main/scala/magenta/deployment_type/AmiCloudFormationParameter.scala#L34

We think that removing this tag should be safe as riff raff should not remove the tags that have already been added. It will just stop adding tags to new deployments, and the ~40 projects that rely on this ami update deployment type should continue to work without issue.

### DIFF
--- a/riff-raff/app/AppComponents.scala
+++ b/riff-raff/app/AppComponents.scala
@@ -108,8 +108,7 @@ class AppComponents(
 
       Map(
         "gu:build-tool" -> buildTool.getOrElse(default),
-        "gu:repo" -> repoUrl.getOrElse(default),
-        "gu:riff-raff:project" -> projectName
+        "gu:repo" -> repoUrl.getOrElse(default)
       )
     }
   }


jorgeazevedo:
@akash1810 Question: we should be able to test if tags do not get removed by deploying this branch to CODE deploying cdk playground, right?

akash1810:
> @akash1810 Question: we should be able to test if tags do not get removed by deploying this branch to CODE deploying cdk playground, right?

Yes. Indeed we patched this behaviour in #1379.

jorgeazevedo:
@akash1810 tested with the latest changes from #1379 and yeah, it no longer removes the tags and behaves how we originally expected. I've enabled auto-merge if you're up for giving it a thumbs up

akash1810 (approved):

PR #1380, opened 2024-09-26
fix(ami-cloudformation-parameter): Perform update within a CFN change set

akash1810:
> [!NOTE]
> Might be easier to review [commit by commit](https://github.com/guardian/riff-raff/pull/1380/commits).

## What does this change?
Updates the `ami-cloudformation-parameter` deployment type to execute within a change set for increased reliability. This brings it in line with the [`cloud-formation` deployment type](https://github.com/guardian/riff-raff/blob/26b241875887e69f900d8cafc400addbbfbd4932/magenta-lib/src/main/scala/magenta/deployment_type/CloudFormation.scala#L228-L249).

When creating a change set, AWS will calculate what, if any, diff there is. When there is no diff, we will now succeed fast.
This differs from current behaviour where we poll the CFN event log, even if the CFN update was a no-op.

More specifically, we poll the first page of the CFN event log, and use a particular status message as an anchor point. If this anchor point is on the 2+ page, the deployment fails. This can happen if a CFN change affected lots of resources,
for example a new tag was added to the stack, or an ASG rolling update was performed.

## How to test
- [Deploy this branch to Riff-Raff CODE](https://riffraff.code.dev-gutools.co.uk/deployment/view/7ecde65a-e51a-4598-bd28-eda8c118900c)
- Using RIff-Raff CODE, deploy [a stack that uses the `ami-cloudformation-parameter` deployment type](https://github.com/search?q=org%3Aguardian+ami-cloudformation-parameter+language%3AYAML++NOT+is%3Aarchived&type=code&l=YAML).
- Observe the deploy log

### AMIable
Build 1560 of AMIable has been setup for to use `ami-cloudformation-parameter`. See https://github.com/guardian/amiable/pull/721. This draft PR uses `ami-cloudformation-parameter` twice; once with `amiParametersToTags` configuration, and once with `amiTags`. This is to demonstrate the functionality is the same in each case.

https://riffraff.code.dev-gutools.co.uk/deployment/view/2588f8c9-30f5-4008-9b9c-980689817b78 is a deploy of AMIable CODE:

image

Note there are multiple lines for `AWS::AutoScaling::AutoScalingGroup` being `UPDATE_IN_PROGRESS` due to the AMI update being performed via [AutoScalingRollingUpdate](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-updatepolicy.html#cfn-attributes-updatepolicy-rollingupdate). See https://github.com/guardian/amiable/pull/719.

### S3 Uploader
I've also successfully [deployed S3 Uploader CODE](https://riffraff.code.dev-gutools.co.uk/deployment/view/8802ef78-7351-49f1-84fd-c48427245f4c), as an [example](https://github.com/guardian/s3-upload/blob/5d7412d140d57bc2d9bd05c35adb9d88103b1e2b/conf/riff-raff.yaml#L10) of a CFN stack that is in a different account, demonstrating the IAM Role assumption logic still works.


magenta-lib/src/main/scala/magenta/deployment_type/AmiCloudFormationParameter.scala
```diff
@@ -31,19 +37,51 @@ object AmiCloudFormationParameter
...
+      val amiLookupFn = getLatestAmi(pkg, target, reporter, resources.lookup)
+
+      val unresolvedParameters = new CloudFormationParameters(
+        target = target,
+        stackTags = stackTags,
+        amiParameterMap = amiParameterMap,
+        latestImage = amiLookupFn,
+
+        // Not expecting any user parameters in this deployment type
+        userParameters = Map.empty
```
akash1810, line 64: The `amiParametersToTags` parameter of the [`ami-cloudformation-parameter` deployment type](https://riffraff.gutools.co.uk/docs/magenta-lib/types#amicloudformationparameter) gets mapped to `amiParameterMap`. That is, we can assert that there are no other CFN parameters being adjusted.

akash1810:
TODO:
- [ ] Deploying project using `ami-cloudformation-parameter` with `amiParametersToTags`
- [ ] Deploying project using `ami-cloudformation-parameter` with `amiTags`

The behaviour should be the same. Should the deployment logs be the same too?

PR #1383, opened 2024-10-11
feat: Set `MinInstancesInService` via CFN parameters

akash1810:
## Context
In https://github.com/guardian/cdk/pull/2417 support was added to `@guardian/cdk` (GuCDK) to deploy application updates to an autoscaling group (ASG) via CloudFormation (CFN) with the [`AutoScalingRollingUpdate` update policy](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-updatepolicy.html#cfn-attributes-updatepolicy-rollingupdate).

During the [testing](https://github.com/guardian/testing-asg-rolling-update), we [observed](https://github.com/guardian/testing-asg-rolling-update/tree/main?tab=readme-ov-file#mininstancesinservice) that for a service that horizontally scales, the `MinInstanceInService` property should be set relative to the current state of the ASG. Consequently when an autoscaling group has scaling policies, GuCDK adds a parameter to the CFN template in the form `MinInstancesInServiceFor<APP>`:

> To support this dynamic property, services that horizontally scale should expose a CloudFormation parameter for Riff-Raff to set during deployment.
> 
> – https://github.com/guardian/testing-asg-rolling-update/blob/main/README.md

## What does this change?
This change adds a new `riff-raff.yaml` property to the `ami-cloudformation-parameter` and `cloud-formation` deployment types called `minInstancesInServiceParameters`.

The CFN parameter added by GuCDK does not provide a direct connection the to ASG it relates to. The new `minInstancesInServiceParameters` property attempts to provide this connection via tag lookups. That is, it is a map of CFN parameter -> ASG tags.

In the below example, we're telling Riff-Raff to expect a CFN parameter named `MinInstancesInServiceForcdkplayground` in the `CdkPlayground.template.json` CFN template and to set it's value from the desired capacity of an ASG tagged `App=cdk-playground`, `Stage=PROD`, `Stack=playground`.

```yaml
allowedStages:
  - PROD
stacks:
  - playground
regions:
  - eu-west-1
deployments:
  cfn-deployment:
    type: cloud-formation
    parameters:
      templateStagePaths:
        PROD: CdkPlayground.template.json
      minInstancesInServiceParameters:
        MinInstancesInServiceForcdkplayground:
          App: cdk-playground
```

Once the ASG has been found, the `MinInstancesInService` property to whichever is lower between desired or 75% of max, meaning:
- When running normally, then we double capacity and half it again once healthy
- When partially or fully scaled, we rotate instances in batches, which is slightly slower

This contrasts with the recommendation from https://github.com/guardian/testing-asg-rolling-update:

> - If the service is running at "normal" capacity (e.g. desired = min), then `MinInstancesInService` can be set to match min.
> - If the service is partially scaled, then `MinInstancesInService` should be set to match the current desired capacity.
> - If the service is fully scaled, then `MinInstancesInService` should be set to (at least) max -1.
> 
> – https://github.com/guardian/testing-asg-rolling-update/blob/main/README.md

This is because being percentage based offers faster deployment.
## How to test
See added unit tests.

We can also perform some practical tests. Each of the following have been performed on https://github.com/guardian/cdk-playground.

### "Normal" capacity (i.e. `desired = min`)
https://github.com/guardian/cdk-playground/pull/548 tests this scenario. It demonstrates the CFN parameter was correctly set to `desired`.

<details><summary>Deployment log showing value of CFN parameter</summary>
<p>

image

</p>
</details> 

### Partial scaled (i.e. `min < desired < max`)
https://github.com/guardian/cdk-playground/pull/548 tests this scenario. It demonstrates the CFN parameter was correctly set to `desired`.

<details><summary>Deployment log showing value of CFN parameter</summary>
<p>

image

</p>
</details> 

### Fully scaled (i.e. `desired = max`)
https://github.com/guardian/cdk-playground/pull/548 tests this scenario. It demonstrates the CFN parameter was set to 7 (75% of 10, rounded down).

<details><summary>Deployment log showing value of CFN parameter</summary>
<p>

image

</p>
</details> 

### No scaling policy (i.e. our typical ASG)
This test amounts to omitting the `minInstancesInServiceParameters` property in the `riff-raff.yaml` file.

This has been done (https://github.com/guardian/cdk-playground/pull/547), and the [deployment](https://riffraff.code.dev-gutools.co.uk/deployment/view/cb8577e0-7717-4f34-a126-454f32de5e52?verbose=1) behaved as expected: CFN doubled the ASG capacity, waited for `SUCCESS` signals, then removed old instances by halving the ASG capacity.

This can be seen on [the dashboard](https://metrics.gutools.co.uk/d/adyqx5hf3w1s0e/cdk-playground?orgId=1&from=1728653082346&to=1728653334575) (note: this link might not work in future).

### `minInstancesInServiceParameters` describes a CFN parameter that does not exist
https://github.com/guardian/cdk-playground/pull/549 tests this scenario.

The deployment should fail when referencing a CFN parameter that does not exist in the CFN template.

<details><summary>Deployment log showing deployment error</summary>
<p>

image

</p>
</details> 

### `minInstancesInServiceParameters` describes an ASG that does not exist
https://github.com/guardian/cdk-playground/pull/550 tests this scenario.

The deployment should fail when referencing an ASG that does not exist.

<details><summary>Deployment log showing deployment error</summary>
<p>

image

</p>
</details>

### Initial deployment (i.e. CFN stack does not yet exist)
This code path is the same as the "`minInstancesInServiceParameters` describes an ASG that does not exist" test.

This change means we're able to use the `AutoScalingRollingUpdate` deployment mechanism on all ASG architectures: those that horizontally scale and those that do not horizontally scale.

### DIFF
--- a/magenta-lib/src/main/scala/magenta/deployment_type/CloudFormationDeploymentTypeParameters.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/CloudFormationDeploymentTypeParameters.scala
@@ -224,13 +224,13 @@ trait CloudFormationDeploymentTypeParameters {
           TagMatch("Stack", target.stack.name),
           TagMatch("Stage", target.parameters.stage.name)
         )
-        params.map({ case (cfnParam, tagRequirements) =>
+        params.map { case (cfnParam, tagRequirements) =>
           cfnParam -> {
-            tagRequirements
-              .map({ case (key, value) => TagMatch(key, value) })
-              .toList ++ stackStageTags
+            tagRequirements.map { case (key, value) =>
+              TagMatch(key, value)
+            }.toList ++ stackStageTags
           }
-        })
+        }
       case _ => Map.empty
     }
   }
--- a/magenta-lib/src/main/scala/magenta/deployment_type/CloudFormationDeploymentTypeParameters.scala
+++ b/magenta-lib/src/main/scala/magenta/deployment_type/CloudFormationDeploymentTypeParameters.scala
@@ -1,11 +1,15 @@
 package magenta.deployment_type
 
+import magenta.deployment_type.CloudFormationDeploymentTypeParameters.CfnParam
+import magenta.tasks.ASG
+import magenta.tasks.ASG.TagMatch
 import magenta.tasks.UpdateCloudFormationTask.{
   CloudFormationStackLookupStrategy,
   LookupByName,
   LookupByTags
 }
 import magenta.{DeployReporter, DeployTarget, DeploymentPackage, Lookup}
+import software.amazon.awssdk.services.autoscaling.AutoScalingClient
 
 import java.time.Duration
 import java.time.Duration.ofMinutes
@@ -104,6 +108,24 @@ trait CloudFormationDeploymentTypeParameters {
       """.stripMargin
   ).default(false)
 
+  val minInstancesInServiceParameters = Param[Map[CfnParam, TagCriteria]](
+    "minInstancesInServiceParameters",
+    optional = true,
+    documentation = """Mapping between a CloudFormation parameter controlling the MinInstancesInService property of an ASG UpdatePolicy and the ASG.
+        |
+        |For example:
+        |```
+        |  minInstancesInServiceParameters:
+        |    MinInstancesInServiceForApi:
+        |      App: my-api
+        |    MinInstancesInServiceForFrontend:
+        |      App: my-frontend
+        |```
+        |This instructs Riff-Raff that the CFN parameter `MinInstancesInServiceForApi` relates to an ASG tagged `App=my-api`.
+        |Additional requirements of `Stack=<STACK BEING DEPLOYED>`, `Stage=<STAGE BEING DEPLOYED>` and `aws:cloudformation:stack-name=<CFN STACK BEING DEPLOYED>` are automatically added.
+      """.stripMargin
+  )
+
   val secondsToWaitForChangeSetCreation: Param[Duration] = Param
     .waitingSecondsFor(
       "secondsToWaitForChangeSetCreation",
@@ -190,5 +212,26 @@ trait CloudFormationDeploymentTypeParameters {
           }
         }
       }
+  }
+
+  def getMinInServiceTagRequirements(
+      pkg: DeploymentPackage,
+      target: DeployTarget
+  ): Map[CfnParam, List[TagMatch]] = {
+    minInstancesInServiceParameters.get(pkg) match {
+      case Some(params) =>
+        val stackStageTags = List(
+          TagMatch("Stack", target.stack.name),
+          TagMatch("Stage", target.parameters.stage.name)
+        )
+        params.map({ case (cfnParam, tagRequirements) =>
+          cfnParam -> {
+            tagRequirements
+              .map({ case (key, value) => TagMatch(key, value) })
+              .toList ++ stackStageTags
+          }
+        })
+      case _ => Map.empty
+    }
   }
 }
--- a/magenta-lib/src/main/scala/magenta/tasks/AWS.scala
+++ b/magenta-lib/src/main/scala/magenta/tasks/AWS.scala
@@ -497,54 +497,50 @@ object ASG {
   ): Int = {
     groupWithTags(tagRequirements, client, reporter) match {
       case Some(asg) =>
-        (asg.desiredCapacity, asg.maxSize) match {
-          case (desired, max) =>
-            val seventyFivePercent: Int = (max * 0.75).toInt
-            val minInstancesInService = Math.min(
-              seventyFivePercent,
-              desired
-            )
+        val max = asg.maxSize
+        val desired = asg.desiredCapacity
+        val seventyFivePercent: Int = (max * 0.75).toInt
+        val minInstancesInService = Math.min(
+          seventyFivePercent,
+          desired
+        )
 
-            // Happy path. We have enough headroom to double the instances, then half once healthy.
-            if (2 * desired <= max) {
-              reporter.info(
-                s"""
-                   |Deploying new instances all at once.
-                   |
-                   |Max=$max. Desired=$desired.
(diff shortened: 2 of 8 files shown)


magenta-lib/src/main/scala/magenta/tasks/AWS.scala
```diff
@@ -490,6 +490,60 @@ object ASG {
...
+                   |""".stripMargin
+              )
+            } else if (minInstancesInService < desired) {
+              reporter.warning(
+                s"""
+                   |Deploying new instances slower than usual.
+                   |
+                   |Max=$max. Desired=$desired.
+                   |Setting MinInstancesInService=$minInstancesInService.
+                   |This is 75% of max capacity, and less than desired capacity, meaning availability will be impacted.
```
akash1810, line 524: TODO: Make this wording more intention revealing.

akash1810: Done.

magenta-lib/src/main/scala/magenta/deployment_type/CloudFormationDeploymentTypeParameters.scala
```diff
@@ -104,6 +108,24 @@ trait CloudFormationDeploymentTypeParameters {
...
+    "minInstancesInServiceParameters",
+    optional = true,
+    documentation = """Mapping between a CloudFormation parameter controlling the MinInstancesInService property of an ASG UpdatePolicy and the ASG.
+        |
+        |For example:
+        |```
+        |  minInstancesInServiceParameters:
+        |    MinInstancesInServiceForApi:
+        |      App: my-api
+        |    MinInstancesInServiceForFrontend:
```
adamnfish, line 121: I think this is a bit optimistic! `MinInstancesInServiceForFrontend` looks nice, but I think it is actually going to be `MinInstancesInServiceForfrontend` (note the lower case `f`). This is even more confusing when we don't have nice short names like "Frontend" and "Api". I was wondering whether we should add a delimiter to the name?

`MinInstancesInServiceFor-frontend` is probably easier to parse at a glance? Probably need to think about this alongside the CDK change as well, so not something to consider in isolation here!

adamnfish (commented):
Lots of comments, sorry! Some tiny Scala style things, but a few other observations in there as well.

We have nice tests for the fiddly min / 75% logic, which is great. The other tricky bit is the way tags get added from multiple places now, it'd be great if we could make sure this is covered?

magenta-lib/src/main/scala/magenta/deployment_type/CloudFormationDeploymentTypeParameters.scala
akash1810: GuCDK's logic is here - https://github.com/guardian/cdk/pull/2476/files#diff-60795fa553ec02295a0718711b79382e5aa10022970df8240822e4a214fe114dR130-R132.

It's worth noting that these map to CFN parameters in the template; `-` are not allowed in CFN parameters.

adamnfish: I checked the tags restrictions but not the parameter name ones, sorry! If those are just alphanumerics then this is a bit harder 😅 

No dramas!

magenta-lib/src/test/scala/magenta/tasks/ASGTest.scala
```diff
@@ -398,6 +398,150 @@ class ASGTest extends AnyFlatSpec with Matchers with MockitoSugar {
     ) shouldBe Right(())
   }
 
+  it should "calculate MinInstancesInService as desired when there is capacity to double" in {
```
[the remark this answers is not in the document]
akash1810: I'm unsure if the, albeit slight, indirection reduces accessibility/maintainability. Additionally, given this is a new deployment mechanism, having explicit capacity values in the tests makes it easier to explain teams as they migrate.

Maybe we can implement these changes in a separate PR?

magenta-lib/src/test/scala/magenta/deployment_type/CloudFormationTest.scala
```diff
@@ -602,7 +604,8 @@ class CloudFormationTest
       Map.empty,
       Map.empty,
       (_: CfnParam) =>
-        (_: String) => (_: String) => (_: Map[String, String]) => None
+        (_: String) => (_: String) => (_: Map[String, String]) => None,
```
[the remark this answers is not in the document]
adamnfish: Yep and as I said no dramas if you want to leave it anyway.

akash1810:
> The other tricky bit is the way tags get added from multiple places now, it'd be great if we could make sure this is covered?

I can't think of the best way to do this:
- `getMinInstancesInService` take explicit tags, rather than a general list of tags, but that's quite a complex interface
- Another option could be to cross reference the discovered ASG with a CFN list resources call and cross reference the two. However, this felt overly complex.
- Tags prefixed `aws:` can only be created by AWS. Maybe `getMinInstancesInService` could assert that the list of tag requirements has at least one `aws:cloudformation:*` tag? This doesn't directly address your concern though.

Tricky!

magenta-lib/src/test/scala/magenta/deployment_type/CloudFormationTest.scala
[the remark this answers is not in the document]
akash1810: 👍🏽 

I'll leave this comment "unresolved" to make it easier to locate/reference in any future PR(s).

magenta-lib/src/test/scala/magenta/tasks/ASGTest.scala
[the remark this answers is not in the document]
akash1810: 👍🏽 

I'll leave this comment "unresolved" to make it easier to locate/reference in any future PR(s).

adamnfish (approved):
We've chatted over all the feedback, and I think it looks fine! The main thing is to do a bunch of testing so let's get it merged and checked 👍

PR #1413, opened 2025-01-20
Update the error message if fastly deployment fails to reflect the fact that it needs a currently active version

fredex42:
Update the error message if fastly deployment fails to reflect the fact that it needs a currently active version

## Accessibility


akash1810 (approved):
Would be good to have Riff-Raff perform that initial deployment too, but this is a great improvement by itself. Thanks!

jorgeazevedo:
@fredex42 I think this would be good to merge if you can

PR #1417, opened 2025-01-30
Delete .github/workflows/snyk.yml

akash1810:
## What does this change?
We're moving away from Snyk and onto Dependabot. Dependabot doesn't have native SBT support, but since #1397 we're submitting to the dependency graph.


kelvin-chappell (approved):
👍

PR #1429, opened 2025-05-20
ci: Make Node version deterministic

akash1810:
## What does this change?
We currently author client side assets with CoffeeScript, which is then compile to JavaScript via an SBT task. This step requires Node.

This change updates CI to make the version of Node used deterministic. Previously, we were using whatever version of Node was on the GitHub runner.

At the time of writing, 22.15.1 is the latest LTS version.

## How to test
Observe CI. It should pass.


.nvmrc
```diff
@@ -0,0 +1 @@
+v22.15.1
```
adamnfish, line 1: If you're using mise, would it be better to say v22 instead of specifiying today's patch version?

The trouble with this approach is that this is arguably worse than what was there before. Previously, GitHub were keeping this up to date for us and if there was a problem we'd (likely) find out at build time.

`22.15.1` is a good choice today but we have no tooling to keep this up to date. Unless we remember to regularly bump this version number, it'll start accumulating vulnerabilities.

PR #1431, opened 2025-05-20
docs: Hint towards using the new deployment mechanism

akash1810:
## What does this change?
Update the "fixing a failed deploy" page with a nudge towards the new deployment mechanism.

### DIFF
--- a/riff-raff/public/docs/howto/fix-a-failed-deploy.md
+++ b/riff-raff/public/docs/howto/fix-a-failed-deploy.md
@@ -1,5 +1,9 @@
 # Fixing a failed deploy
 
+> [!NOTE]
+> This guide is only necessary when using the legacy `autoscaling` deployment type.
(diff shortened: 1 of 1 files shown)


riff-raff/public/docs/howto/fix-a-failed-deploy.md
```diff
@@ -1,5 +1,9 @@
 # Fixing a failed deploy
 
+> [!NOTE]
+> This guide is only necessary when using the legacy `autoscaling` deployment type.
```
jacobwinch, line 4: It might be worth linking out to more details here, just in case people are unsure what they're currently using?

```suggestion
> This guide is only necessary when using the legacy [`autoscaling` deployment type](https://riffraff.gutools.co.uk/docs/magenta-lib/types#autoscaling).
```

jacobwinch (approved):

PR #1443, opened 2025-06-23
add cache generator for cfn types list

johnduffell:
We have trouble in support-service-lambdas where deploys frequently time out with a Rate Exceeded error when doing 40+ simultanous lambda deploys.

This is narrowed down to the fact that

- We use the safe/non destructive riffraff method of deploying cloudformation, which applies a stack policy before starting the deploy
- the stack policy needs to only use types that exist in the given region
- we fetch the valid types on every deploy
- there are 1300+ types in eu-west-1
- our rate limit for ListTypes is 2 calls per minute
- the max page size is 100
- we do 40+ deploys on every merge to support-service-lambdas
- AWS java sdk only does 3 minutes 51 worth of retries

Although there's an argument that 40+ deploys at the same time is unnecessary, it's probably not needed to refetch the types on every deploy, and that would speed everyone's deploys up by a few seconds.

This PR adds
- a new file with a hard coded set of eu-west-1 types as of today.
- a small SBT task to update it/add other regions
- small changes to make it check the hard coded set before calling out to AWS

Tested in CODE and seems to work fine.
This is a deploy of a lambda in PROD, you can see it takes ~50 seconds: https://riffraff.gutools.co.uk/deployment/view/cb8b9917-8e84-4c24-8980-24a683a3e6c7?verbose=1
```
 [09:36:03] Found stack support-CODE-alarms-handler (arn:aws:cloudformation:eu-west-1:865473395570:stack/support-CODE-alarms-handler/c0e60cb0-eabb-11ee-8e13-06c0c5e0f1d3)
  [09:36:51] Setting update policy for stack support-CODE-alarms-handler to DenyReplaceDeletePolicy
(this time varies depending on how many other deploys are happening around that time)
```
and this with this branch on CODE it takes 4 seconds
https://riffraff.code.dev-gutools.co.uk/deployment/view/60b6bee6-c56e-4570-9c7e-e82c4c14e6fd?verbose=1
```
[10:58:15] Found stack support-CODE-alarms-handler (arn:aws:cloudformation:eu-west-1:865473395570:stack/support-CODE-alarms-handler/c0e60cb0-eabb-11ee-8e13-06c0c5e0f1d3)
  [10:58:19] Setting update policy for stack support-CODE-alarms-handler to DenyReplaceDeletePolicy
```

TODO
- improve this description with some links


project/GenerateCfnTypes.scala
```diff
@@ -0,0 +1,59 @@
...
+      |
+      |object $objectName {
+      |  val regionTypes: Map[String, Set[String]] = Map(
+      |    ${getTypes(regions, profileName).map { case (region, types) => s""""$region" ->
+      |      Set(
+      |        ${types.map(t => s""""$t"""").mkString(",\n        ")}
+      |      )""".stripMargin}.mkString(",\n      ")}
+      |  )
+      |}
+      |""".stripMargin
```
johnduffell, line 32: it does feel a bit low level using string interpolation to make a scala file, but I think using something like https://scalameta.org/docs/trees/guide.html is probably overkill for something like this.

johnduffell: being honest I did start down exactly that route with a shell script and I was using sed to strip out all the irrelevant stuff  (as it's a massive amount of data for each one) and the file would just be one line per record:
`aws cloudformation list-types --profile membership --visibility PUBLIC --type RESOURCE --filters Category=AWS_TYPES --output yaml |sed -n 's/^ *TypeName: //p'` 

However ultimately if it was in a file we'd have to handle if the file is in the wrong place or can't be parsed for some reason etc.  It might behave differently in runtime as locally.

Making it scala code means it's all checked and type checked at compile time locally and in the build, and I'm a big fan of checking things as early as possible in the develop/test cycle.  It also meant that I could reuse the existing code that's used at run time.  It would cost a little bit of compile time though.

The other thing I was slightly unsure about was including the AWS SDK in the build definition, as it will be loaded on every build.  A more efficient way would be to have a proper subproject that does the generation, and only compile and run it when we're generating the code, rather than having it in the build.

Happy to revisit accordingly though, or pair on it.

PR #1448, opened 2025-07-02
Enhance README with comprehensive deployment guide

andrade0:
## What does this change?

This PR significantly enhances the `README.md` documentation for Riff-Raff, while preserving all existing content.  
**Why:** The previous documentation was concise but lacked comprehensive guidance for new and existing developers.  
**How:**  
- Adds an in-depth architecture overview
- Introduces visual Mermaid diagrams for runtime and CI/CD flows
- Documents all supported deployment types and configuration best practices
- Provides step-by-step GitHub Actions integration guidance
- Includes troubleshooting, advanced features, and resource links
- Uses rich formatting, emojis, and anchor links for improved readability

## How to test

- Open the updated `README.md` in your local editor or on GitHub.
- Verify that all original content remains unchanged at the top.
- Check that the new sections appear after the original content.
- Confirm that Mermaid diagrams render correctly (use a Markdown viewer with Mermaid support, or GitHub's web UI).
- Review for clarity, completeness, and technical accuracy.

## How can we measure success?

- Developers (internal and external) can quickly understand how to use and extend Riff-Raff.
- Fewer questions about deployment, configuration, and CI/CD integration.
- Improved onboarding experience for new team members.
- Positive feedback from reviewers and users of the documentation.

## Have we considered potential risks?

- **Risk:** The README becomes too long or overwhelming.
  - **Mitigation:** Retained all original content and structured new content with clear sections and a table of contents.
- **Risk:** Mermaid diagrams may not render in all Markdown viewers.
  - **Mitigation:** Diagrams are compatible with GitHub’s Markdown renderer and most modern tools.
- **Risk:** Outdated information if Riff-Raff evolves.
  - **Mitigation:** Sections are modular and easy to update in future PRs.

No sensitive information is introduced. No changes to application code or infrastructure.

## Images

**Before:**  
- README was concise, with basic architecture and usage notes.

**After:**  
- README includes detailed diagrams, configuration guides, and comprehensive documentation sections.

*(See the rendered README in this branch for a full preview.)*

## Accessibility

-   [ ] [Tested with screen reader](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#screen-reader)
-   [ ] [Navigable with keyboard](https://github.com/guardian/accessibility/blob/main/people-and-technology/02-physical.md#keyboard)
-   [ ] [Colour contrast passed](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#contrast)
[...]

PR #1451, opened 2025-07-11
Improve Riff-Raff's documentation

jacobwinch:
This PR attempts to address some of the opportunities for improvement identified in: https://github.com/guardian/riff-raff/pull/1448#issuecomment-3044076672.

The main changes here are:

* Removing references to the legacy tools for uploading artifacts to Riff-Raff and mention https://github.com/guardian/actions-riff-raff instead
* Telling people about the `riff-raff.yaml` auto-generation so that time isn't spent writing these files by hand for new projects
* Updating the menu to make it easier to find the validation feature and docs about configuring new projects
* Removing some very old documentation which is now out of date

I've also made some other minor edits along the way (e.g. slightly shortening and re-ordering the main `readme`).


TODO.txt
```diff
@@ -1,18 +0,0 @@
...
- * Blog on guardian developer blog
- * Yak on about it at conferences
-
-RiffRaff as development hub
-===========================
-
-It may be desirable to allow RiffRaff to do things like track pull requests so that it can act as a hub for tracking project activity.
-
-Upgrade Scala/Play etc.
-======================
```
jacobwinch, line 18: I don't think there are currently plans to work on any of this.

readme.md
```diff
@@ -5,18 +5,18 @@
 The Guardian's scala-based deployment system is designed to automate deploys by providing a web application that 
 performs and records deploys, as well as providing various integration points for automating deployment pipelines.
 
+## Documentation
+Riff-Raff's documentation is available in the application (under the Documentation menu) but can also be viewed under
+[riff-raff/public/docs](riff-raff/public/docs) in GitHub.
+
 ## Requirements
-Riff-Raff and Magenta have been built with the tools we use at the Guardian
```
jacobwinch, line 9: I think mentioning these as two distinct things here will be confusing for most people.

readme.md
```diff
@@ -5,18 +5,18 @@
...
+Riff-Raff has been built with the tools we use at the Guardian
 and you will find it easiest if you use a similar set of tools. Riff-Raff:
 
  - relies on artifacts and `riff-raff.yaml` files describing builds being in S3 buckets with the artifacts having paths of 
   the form `project-name/build-number`
  - uses the AWS SDK and [Prism](https://github.com/guardian/prism) to do resource discovery
- - stores configuration, history and logs in a PostgreSQL database and a handful of DynamoDB tables (the eventual aim is to ditch DynamoDB altogether)
```
jacobwinch, line 15: I don't think we're currently planning to remove DynamoDB and this feels like too much detail given the brevity of the readme overall

riff-raff/app/controllers/Application.scala
```diff
@@ -63,6 +63,10 @@ class Menu(config: Config) {
     SingleMenuItem("History", routes.DeployController.history()),
     SingleMenuItem("Deploy", routes.DeployController.deploy),
     SingleMenuItem("Deployment Resources", routes.Application.deployInfoData),
+    SingleMenuItem(
```
jacobwinch, line 66: This isn't really documentation, so I've moved it out into it's own menu item

riff-raff/app/controllers/Application.scala
```diff
@@ -85,12 +89,12 @@ class Menu(config: Config) {
       "Documentation",
       Seq(
         SingleMenuItem(
-          "Deployment Types",
-          routes.Application.documentation("magenta-lib/types")
+          "Configuring a project",
```
jacobwinch, line 92: This is a pretty common use-case, so I think it's worth moving it out of `Everything else`

riff-raff/public/docs/riffraff/administration/move.md
```diff
@@ -1,25 +0,0 @@
-<!--- prev:auth -->
```
jacobwinch, line 1: We now rotate the instance automatically every week, so this should no longer be needed

riff-raff/public/docs/riffraff/s3buckets.md
```diff
@@ -1,67 +0,0 @@
-<!--- prev:externalRequest -->
-Riff-Raff, AWS credentials and S3 uploads
```
jacobwinch, line 2: I think this is now covered by https://github.com/guardian/deploy-tools-platform/blob/main/cloudformation/riffraff/riff-raff-roles.md

riff-raff/public/docs/riffraff/s3buckets.md
```diff
@@ -1,67 +0,0 @@
...
-      }
-    }
-
-In this particular case we are interested in the `credentials:aws` section. The dicts in this section are processed in
-order and the first match for the app being deployed and the stage being deployed to is selected.  The app and stage
-values are regular expressions.  Note that order is important - in the above example frontend-gallery being deployed to
-CODE will use key BBB, rather than SHADOWED because it arrives at that entry first even though the latter entry
-is more specific.
-
-Allowing riff-raff access to your bucket
```
jacobwinch, line 40: And this is no longer necessary...

riff-raff/public/docs/howto/configure-a-project.md
```diff
@@ -1,45 +1,25 @@
 Configuring a project
 =====================
 
-To make a project deployable with riff-raff it needs:
+To make a project deployable with Riff-Raff it needs:
  
  - A [`riff-raff.yaml`](../reference/riff-raff.yaml.md) file that describes the deployment process
  - A [`build.json`](../reference/build.json.md) file that details the CI build
- - The two files above along with any assets uploaded into riff-raff's S3 artifact buckets
+ - The two files above along with any assets uploaded into Riff-Raff's S3 artifact buckets
```
akash1810, line 8: Worth changing the order here, with the artiact files as the first item as it's the set of files that one controls?

riff-raff/public/docs/index.md
```diff
@@ -22,10 +22,5 @@ Using Riff-Raff
  - [Continuous Integration and Deployment](riffraff/hooksAndCD.md) - guide to the Riff-Raff hooks available to make continuous
  deployment easy
  - [External deploy requests](riffraff/externalRequest.md) - how to help a user start a deploy
- - [AWS S3 bucket configuration for uploads](riffraff/s3buckets.md)
  - [Administration](riffraff/administration/) - details on how to configure Riff-Raff
- - [Restrictions](riffraff/restrictions.md) - how to restrict deployments
-
-Riff-Raff not picking up builds?
```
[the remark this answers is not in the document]
akash1810: I don't remember any code changes we've made to explicitly improve here. However, I also do not remember the last report/question about this. Maybe a combination of scheduled instance rotation and using an updated AWS SDK have helped.

riff-raff/public/docs/reference/build.json.md
```diff
@@ -4,8 +4,7 @@ The build.json file
 The type ahead of the project name and build number and also the continuous deployment feature is powered by the
 `build.json` files that are created for each project.
 
-This file is normally created by the [sbt-riffraff-artifact plugin](https://github.com/guardian/sbt-riffraff-artifact)
-or [node-riffraff-artifact plugin](https://github.com/guardian/node-riffraff-artifact) plugin.
+This file is normally created by [actions-riff-raff](https://github.com/guardian/actions-riff-raff).
```
akash1810, line 7: Wonder if we should advise against manually creating this file as it makes it harder to change the shape of it? For example, we may want to add the [name of a repository's default branch](https://docs.google.com/document/d/1KkKNLCTZ_5Y1sCy0DetpDH-8aV_SwVKqEFaN3P6A3eU/edit?tab=t.0).

riff-raff/public/docs/reference/riff-raff.yaml.md
```diff
@@ -144,21 +149,35 @@ Examples
...
+    type: cloud-formation
+    app: amigo
+    parameters:
+      amiTags:
+        Recipe: arm64-jammy-java21-deploy-infrastructure
+        AmigoStage: PROD
+      amiEncrypted: true
+      templateStagePaths:
+        CODE: AMIgo-CODE.template.json
+        PROD: AMIgo-PROD.template.json
```
akash1810, line 176: I might be mistaken, but usage of `templateStagePaths` is limited to CDK defined templates as we produce a distinct template per stage. CFN YAML templates use CFN conditions on a `Stage` parameter to achieve stage awareness. If this section is providing an example of a hand-written `riff-raff.yaml`, should we tweak the example to be CFN YAML focused?

riff-raff/public/docs/reference/riff-raff.yaml.md
```diff
@@ -144,21 +149,35 @@ Examples
...
+      amiParameter: AMIAmigo
 ```
 
-This is a more complex example that uses templates and overrides at various levels. This also uses the JSON notation
+This is an example of how templates and overrides can be used. This example also uses the JSON notation
```
akash1810, line 180: Does  `JSON notation for the arrays` relate to the YAML spec? Is this too much information here?

riff-raff/public/docs/riffraff/administration/move.md
akash1810: TIL that we can block queueing for new deploys! We don't so this in the instance rotation process. In https://github.com/guardian/riff-raff/pull/691 we mentioned:

> ## Have we considered potential risks?
> [Continuous deployments](https://github.com/guardian/recommendations/blob/main/continuous-deployment.md) that are triggered during the (roughly 2 minute) downtime will not run, meaning users will have to manually deploy their change ([PRout](https://github.com/guardian/prout) can help surface this!). This risk can be somewhat mitigated by scheduling the cronjob to run just outside of typical working hours.
> 
> A similar concern applies to scheduled deployments. Any scheduled deployment that is due to start when Riff-Raff is unavailable will not be automatically started when Riff-Raff becomes available again. As this is the current status quo, would suggest this be resolved in future work/PRs.

I wonder if blocking queuing can help here, as we should be able to run multiple boxes in the ASG vs the current process of terminating a box and waiting for the ASG to recover.

I don't think its a priority as the current behaviour has been in place for while w/out issue.

akash1810 (commented):
Thanks for these changes! 💎 Added a couple of comments/suggestions. None are blocking though.

riff-raff/public/docs/howto/configure-a-project.md
jacobwinch: https://github.com/guardian/riff-raff/pull/1451/commits/81dcda8f2bf5171ef16d93babbd3a59e4bb1c3c7

riff-raff/public/docs/reference/build.json.md
jacobwinch: I didn't go into any detail here but have advised users not to create these by hand: https://github.com/guardian/riff-raff/pull/1451/commits/df0d28f1fd953a2bffd86d75b1f3e6a374af13c0

riff-raff/public/docs/reference/riff-raff.yaml.md
jacobwinch: Yeah, this was already there but I agree that we probably don't need this detail. I've removed it in https://github.com/guardian/riff-raff/pull/1451/commits/5cbcd97914932e05ef09014016c75601af01077b.

akash1810 (approved):

Issue #1459, opened 2025-08-06
`Warnings encountered` striped banner is intermittently showing on builds without explanation


rtyley:
Several recent deploys of Frontend have unpredictably shown the `Warnings encountered` striped banner (originally introduced with https://github.com/guardian/riff-raff/pull/435 & https://github.com/guardian/riff-raff/pull/437):

https://riffraff.gutools.co.uk/deployment/history?projectName=dotcom:frontend-all&stage=CODE

<img width="1642" height="566" alt="Image" src="https://github.com/user-attachments/assets/8299d4a2-d3bd-4955-8fea-acf14b0eec4c" />

However, **no further details of what the warnings are is given**, and the deploy pages themselves seem clean:

* https://riffraff.gutools.co.uk/deployment/view/1b37d1cd-17b9-4cbc-b908-5160f174bd94
* https://riffraff.gutools.co.uk/deployment/view/d260f4dd-13a6-4b3d-b410-ae58640fb61f
* https://riffraff.gutools.co.uk/deployment/view/92bfeb70-5b37-4404-81dd-96ed38b138cc
* https://riffraff.gutools.co.uk/deployment/view/7c46bffe-7efb-4d53-900e-194c06407f78

<img width="779" height="908" alt="Image" src="https://github.com/user-attachments/assets/69541345-1956-4881-891b-818a8c4bf718" />

adamnfish-gu:
I've opened a PR that adds an "expand all" button to this page. This would mean someone can skip most of the above manual steps, and would be able to quickly scroll through the logs looking for the highlighted warning line(s).

Not an ideal solution, but much easier than busting out the dev tools etc.
https://github.com/guardian/riff-raff/pull/1460

PR #1460, opened 2025-08-06
Add an expand-all button to the deploy log

adamnfish-gu:
## What does this change?

Adds an expand-all button to Riff Raff's deploy log UI, so that colleagues can then easily scan the logs to visually identify any warnings.

## Background

Please refer to https://github.com/guardian/riff-raff/issues/1459, which provides much more context and detail.

Riff Raff often emits warnings that are very challenging to find in the deploy log. We add an "expand all" button that ensures the UI is showing all log entries, which makes it much easier for a human to quickly scan the output to find warnings (which are highlighted in the UI).

This isn't a perfect solution to the problem, because it depends on someone being able to visually parse the error warning. It's an easy fix though, and very significantly improves these situations.

## How to test

This is a simple markup change, which I have tested by editing the page in PROD. We can confirm this works in-situ after release.

People will find it much easier to locate deployment warnings.

I've already addressed a layout bug where the new button runs over other content.

The approach has two weaknesses:

1. it still depends on the user being able to visually scan the expanded log lines
2. this really is an "expand all" button, it can't also be used to contract elements

The first is ok becavuse it's a big improvement for a simple change. We aren't trying to do something perfect here.

The second point is ok because someone can simply reload the page to restore the contracted view. The issue is that the bootstrap functionality we're using here would **toggle** all sections, which causes very confusing behaviour for the user if some items had been manually contracted / expanded prior to hitting the expand all button. It's much clearer to have a simple "expand all" button.

## Images
expand-all

### DIFF
--- a/riff-raff/app/views/deploy/logContent.scala.html
+++ b/riff-raff/app/views/deploy/logContent.scala.html
@@ -7,9 +7,24 @@
 @if(!record.isSummarised) {
     @record.report match {
         case report : DeployReportTree => {
-            <ul class="magenta-reporttree">
+            <div class="clearfix">
+                <div class="pull-right">
+                    <button
+                    type="button"
+                    aria-expanded="false"
+                    class="btn btn-default"
+                    data-toggle="collapse"
+                    role="button"
+                    data-target=".collapsing-node.magenta-reporttree:not(.in)"
+                    >
+                        🔽 expand all
+                    </button>
+                </div>
+
+                <ul class="magenta-reporttree">
                 @snippets.reportTree(report, 0)
-            </ul>
+                </ul>
+            </div>
             @logSummary(record)
         }
         case EmptyDeployReport => {


JuliaBrigitte (approved):

rtyley (approved):
Looks great! Has it been deployed to CODE RIff-Raff for a try out? 

Thanks for the big improvement!

adamnfish-gu:
I've deployed it to code to test and it is worth observing that expanded sections are contracted evry time the log refreshes.

This is super annoying.

We might consider using virtualdom to apply the changes on refresh instead of simply overwriting the content every few seconds, but that would be a job for another PR.

I don't love that this raises the prominence of a super annoying UX papercut, but I think this is still improvement enough to be worth including. If people want to only include the expand all when the deploy is complete then we could do that as well, but it is (theoretically) nice to be able to expand the log without having to wait for a deploy to finish.

PR #1465, opened 2025-08-15
Fix Deploy Again and Preview features

jacobwinch:
We accidentally broke the Deploy Again and Preview features when we merged https://github.com/guardian/riff-raff/pull/1458. This PR fixes both features.[^1]

I've tested:

- Deploying a build again (all tasks)
- Deploying a build again (subset of tasks)
- Previewing a deployment before actually deploying it
- Previewing a deployment, choosing a subset of tasks and then deploying the subset of tasks

All of these behaviours are working normally again.

There is still one small issue with the new warning functionality. Riff-Raff takes a few minutes to load all of the builds/branch information when it first starts up. This means that you can only deploy during these first few minutes by manually copying and pasting your build number into the Riff-Raff UI (this behaviour existed prior to #1458). If you do this to deploy to `PROD` it will now make you use the new warning checkbox, even if you are deploying a build number that was generated by the `main` branch. I had a very quick look at fixing this but couldn't get it working. Personally I think it's fine to go ahead despite this bug as it's a bit of an edge case.

[^1]: Users are not currently affected by this anyway as I've deployed a previous build of `main` to `PROD` for now.


riff-raff/app/controllers/DeployController.scala
```diff
@@ -348,14 +348,14 @@ class DeployController(
     }
 
     val params = DeployParameterForm(
-      record.buildName,
-      record.buildId,
-      "",
```
jacobwinch, line 353: The stage and branch fields were being passed in the wrong order in both of these controllers. This led to [strange behaviour/errors](https://riffraff.gutools.co.uk/deployment/view/7513d323-9fda-4189-b8c4-ffc4b307f677). This PR should fix the problem and also uses named parameters to make it clearer what's going on in these cases. 

Unfortunately it's easy to make this type of mistake when we are just passing around lots of unnamed `String`s!

akash1810 (approved):
Have tested on CODE. Looks good.

PR #1464, opened 2025-08-13
Check Google Groups per request

jacobwinch:
## Accessibility

-   [ ] [Tested with screen reader](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#screen-reader)
-   [ ] [Navigable with keyboard](https://github.com/guardian/accessibility/blob/main/people-and-technology/02-physical.md#keyboard)
-   [ ] [Colour contrast passed](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#contrast)
-   [ ] [The change doesn't use only colour to convey meaning](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#use-of-colour)

PR #1471, opened 2025-09-22
Removed unused auth code

jacobwinch:
## What does this change?

This is a follow-up to https://github.com/guardian/riff-raff/pull/1464. Now that we are using Google Group checks for authorisation, we can remove a lot of code that was used to display and manage the old auth list.

We should also remove the unused private config and DB table after merging this change.

## How to test

I've deployed this to `CODE` and confirmed that I can still login and deploy a test project.

We'll have less code to maintain. The migration that we started in https://github.com/guardian/riff-raff/pull/1464 will be complete.

Yes, let's give it a few days before merging this PR in case we receive any negative feedback about the new authorisation approach that we moved to in https://github.com/guardian/riff-raff/pull/1464.


riff-raff/app/controllers/authentication.scala
```diff
@@ -1,76 +1,28 @@
...
-    Json.parse(res.string(1)).as[AuthorisationRecord]
-
-  override val tableName = "auth"
-}
-
-trait AuthorisationValidator {
-  def emailDomainAllowList: List[String]
-  def emailAllowListEnabled: Boolean
-  def emailAllowListContains(email: String): Boolean
-  def isAuthorised(id: UserIdentity) = authorisationError(id).isEmpty
```
jacobwinch, line 49: There's more background and why we no longer need this functionality [here](https://github.com/guardian/riff-raff/pull/1464#discussion_r2301054390).

riff-raff/app/persistence/postgresdb.scala
```diff
@@ -43,47 +43,9 @@ class PostgresDatastore(config: Config) extends DataStore(config) with Logging {
...
-      DB readOnly { implicit session =>
-        sql"SELECT content FROM auth".map(AuthorisationRecord(_)).list().apply()
-      }
-    }
-
-  def setAuthorisation(auth: AuthorisationRecord): Either[Throwable, Unit] =
-    logExceptions(Some(s"Creating auth object $auth")) {
-      DB localTx { implicit session =>
-        val json = Json.toJson(auth).toString()
-        sql"INSERT INTO auth (email, content) VALUES (${auth.email}, $json::jsonb) ON CONFLICT (email) DO UPDATE SET content = $json::jsonb"
```
akash1810, line 74: Should we drop this table?

jacobwinch: Yes definitely, let's have a dedicated PR for that as a follow-up to this one.

PR #1472, opened 2025-09-22
Update contributing docs

jacobwinch:
The config requirements for running locally have changed since https://github.com/guardian/riff-raff/pull/1464 was merged.

I think downloading an example file is easier than replacing figuring out appropriate values for all of the placeholders, so I've also updated the docs to include this suggestion.


akash1810 (approved):
LGTM. Added a non-blocking comment.

PR #1273, opened 2023-10-27
Create a new `service_catalogue` user and schema with views to riff-r…

AshCorr:
## What does this change?

Creates a new `service_catalogue` DB user with access to the `service_catalogue` schema.

We'll be using this new user and views to source riffraff data into our Service Catalogue CloudQuery tables.

## How to test

1. Run riff-raff locally
2. Connect to local DB as `riffraff`
3. Run `ALTER USER service_catalogue PASSWORD 'password';
4. Connect to local DB as `service_catalogue` and the newly set `password` password.
5. Run:

`SELECT * FROM service_catalogue.riffraff_deploys` should run sucessfully
`SELECT * FROM public.deploy` should fail with a permissions error.

image

image

## Before Deployment / Merge

1. Rotate Master credentials on Riff Raff DB and make a note of them
2. SSH onto a riff-raff node to setup a tunnel to the DB `eval $(ssm ssh --profile deployTools -t riff-raff,deploy,(STAGE) --raw --newest) -L (riff raff DB port):(riff raff DB host):(riff raff DB port)`
3. Connect to riff-raff DB and run `ALTER USER (riff raff app username) CREATEROLE;`
    4.

PR #1475, opened 2025-09-30
Remove auth table and service catalogue view of it

jacobwinch:
## What does this change?

This PR removes the `auth` table from the database using [Play's evolutions feature](https://www.playframework.com/documentation/3.0.x/Evolutions). This table has been unused since https://github.com/guardian/riff-raff/pull/1471 was merged.

The PR also removes a view that was made available to Service Catalogue (in https://github.com/guardian/riff-raff/pull/1273). For more context on that change, see https://github.com/guardian/service-catalogue/pull/1645.

Note that to get this to work we first need to manually run: `ALTER TABLE auth OWNER TO riffraff_app;` against the DB.

This is because these tables were created by the `riffraff` user, rather than the `riffraff_app` user that we [moved to when we started using IAM auth for RDS](https://github.com/guardian/deploy-tools-platform/pull/166). In Postgres [the user that creates a table is the owner](https://www.postgresql.org/docs/13/ddl-priv.html).

## How to test

I've tested this in `CODE` and confirmed that the `auth` table and view are dropped as expected.

We've finished cleaning up after https://github.com/guardian/riff-raff/pull/1464 and https://github.com/guardian/riff-raff/pull/1471.

The main risk here is the fact that we'll need to connect to / modify the `PROD` DB manually as part of the rollout. We should pair on this to minimise risk.

### DIFF
--- a/riff-raff/conf/evolutions/default/5.sql
+++ b/riff-raff/conf/evolutions/default/5.sql
@@ -1,0 +1,9 @@
+-- !Ups
+
+DROP TABLE auth;
+DROP VIEW service_catalogue.riffraff_authorized_users;
+
+-- !Downs
+
+CREATE TABLE auth (email VARCHAR(100) PRIMARY KEY, content jsonb NOT NULL);
+CREATE VIEW service_catalogue.riffraff_authorized_users AS SELECT * FROM public.auth;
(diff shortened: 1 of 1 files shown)


riff-raff/conf/evolutions/default/5.sql
```diff
@@ -0,0 +1,9 @@
+-- !Ups
+
+DROP VIEW service_catalogue.riffraff_authorized_users;
+DROP TABLE auth;
+
+-- !Downs
```
jacobwinch, line 6: > The Downs part that describes how to revert them.

PR #1597, opened 2026-05-08
Upgrade to Java 21

jorgeazevedo:
## What does this change?

This upgrades riff-raff to Java 21. There are a number of changes needed to make this work, which I'll call out below.

### java.net.URL deprecation

The usage of (`new URL(...)`) [was deprecated in Java 20](https://www.oracle.com/java/technologies/javase/20-relnote-issues.html#JDK-8294241). The recommended alternative is to replace it with `URI.create(url).toURL`, [as per the documentation](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/net/URI.html#toURL())

> `URI.create(url).toURL()` works as if invoking it were equivalent to evaluating the expression `new URL(this.toString())` after first checking that this URI is absolute.

This is equivalent, but does throw more exceptions that the original `URL` methods, so some of the code was refactored to accommodate this.

For reference, a similar patch had to be done in [prism](https://github.com/guardian/prism/pull/1101)

### Markdown

Riff-raff uses [pegdown](https://github.com/sirthias/pegdown) to render Markdown as html. This library does not work with Java 21, because it relies on another library by the same author called [parboiled-java](https://github.com/sirthias/parboiled) which broke in modern versions of Java.

```
org.pegdown:pegdown:1.6.0
└── org.parboiled:parboiled-java:1.1.7   ← generates parsers via reflection
    ├── org.parboiled:parboiled-core:1.1.7
```

As per [this github issue](https://github.com/sirthias/parboiled/issues/175), parboiled stopped working some years ago due to the release of [JEP-403](https://openjdk.org/jeps/403) which enforces strong encapsulation of JDK internals. There is however a workaround in the github issue, which is to add several `--add-opens` flags, which I've done in this PR to maintain compatibility.

This leads me to a second problem. The more natural solution would be for `pegdown` to be updated to work with a never version of `parboiled` that was refactored to no longer require this workaround. However, pegdown [was deprecated 10 years ago](https://github.com/sirthias/pegdown).

> Although still one of the most popular Markdown parsing libraries for the JVM, pegdown has reached its end of life.
> The project is essentially unmaintained with [tickets](https://github.com/sirthias/pegdown/issues) piling up and crucial bugs not being fixed.

I have a follow up PR to replace `pegdown` with a different library that doesn't require what is now considered "illegal" access in the JVM.

## How has this change been tested?

Deployed to CODE and tested locally.


riff-raff/app/ci/S3Build.scala
```diff
@@ -69,7 +69,7 @@ class S3BuildOps(config: Config) extends Logging {
             )
           }
           case Right(s3Build) => {
-            log.info(s"Successfully parsed JSON from $location")(
+            log.debug(s"Successfully parsed JSON from $location")(
```
jorgeazevedo, line 72: I've changed these to debug so riffraff doesn't flood stdou with thousands of lines of output on every start! It's quite disruptive when running locally

riff-raff/app/controllers/HooksController.scala
```diff
@@ -67,9 +69,7 @@ class HooksController(
       "postBody" -> optional(text)
     )(HookForm.apply)(HookForm.unapply).verifying(
       "URL is invalid",
-      form =>
-        try { new URL(form.url); true }
-        catch { case e: MalformedURLException => false }
```
jorgeazevedo, line 72: The new URI package also throws `IllegalArgumentException`, so just catching all non-fatals and returning true/false seems like a good fit to me.

jacobwinch (approved):
Great work!

PR #1650, opened 2026-08-14
Fix: expanded deploy log steps collapse again on every auto-refresh

davidfurey:
NB - AI PR, reviewed and tested manually

### Problem
When viewing the progress of a deploy, the log content is replaced every second by an ajax refresh to keep it up to date. Since the report tree HTML is fully re-rendered from the server on each refresh, any step the user had manually expanded via the collapse toggle would revert to its default state (collapsed, unless it's the currently-running step) a second later.

### Fix
collapsing.coffee now remembers a node's expand/collapse state only when the user actually triggers it — via the toggle click or the "expand all" button — by hooking into Bootstrap's `show.bs.collapse`/`hide.bs.collapse` events. After each ajax refresh (using the existing `postRefresh` hook), those recorded overrides are reapplied to the matching nodes (keyed by their stable message id), and the toggle listeners are rebound to the new DOM.

Nodes the user hasn't interacted with are left alone and keep following the server-rendered default (open while running, collapsed once done) — so, e.g., the previously-running step still collapses naturally once the next step starts.

### Testing
- Manually verified: expanding a completed step's detail persists across refresh cycles until the user collapses it again, while untouched steps still collapse automatically once complete. ✅

### DIFF
--- a/riff-raff/app/assets/javascripts/auto-refresh.coffee
+++ b/riff-raff/app/assets/javascripts/auto-refresh.coffee
@@ -1,7 +1,6 @@
 intervalId = null
 
 callbackList = $.Callbacks()
-preRefreshCallbackList = $.Callbacks()
 
 visibleHeight = ->
   window.innerHeight || document.documentElement.clientHeight || document.body.clientHeight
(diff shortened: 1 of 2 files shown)


akash1810 (approved):
LGTM.

minor: Would it be worth adding some before/after gifs to further illustrate the problem/fix?

emdash-ie:
Ooooh thank you so much! I looked at more complicated ways of fixing this, this is much simpler 🎉

